PR #6, opened 2014-06-26
Pass docker hostconfig in StartContainer task

edannenberg:
This PR enables running of containers with docker hostconfig params. Also adds volumes param to CreateContainer, to enable mounting of host directories.

For mounting host directories into a container the docker api requires the mount config as volumes on creation, as well as binds on container start. Else you end up with a docker volume or nothing depending on which param you ommit.

``` groovy
ext.mountConfig = [ "/host/path:/var/www/localhost/htdocs" ]

task createLocalDev(type: DockerCreateContainer) {
    imageId = 'foo'
    volumes = mountConfig
}

task startLocalDev(type: DockerStartContainer) {
    dependsOn createLocalDev
    targetContainerId { createLocalDev.getContainerId() }
    binds = mountConfig
    portBindings = [ [scheme: "tcp", port: "80", hostIp: "127.0.0.1", hostPort: "80" ],
                     [scheme: "tcp", port: "443", hostIp: "127.0.0.1", hostPort: "443" ] 
                   ]
    links = [ "some-other-container:db" ]
}
```

### DIFF
--- a/src/main/groovy/org/gradle/api/plugins/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/org/gradle/api/plugins/docker/tasks/container/DockerCreateContainer.groovy
@@ -85,6 +85,10 @@ class DockerCreateContainer extends AbstractDockerTask {
 
     @Input
     @Optional
+    List<String> volumes
+
+    @Input
+    @Optional
     String volumesFrom
 
     @Input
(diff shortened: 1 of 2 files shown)


tclift:
I was looking for the `PortBindings` config that was a part of this PR. Was it never added?

I want to map a container port onto a host port, and this seems to be required. Perhaps I'm missing something?

bmuschko:
@tclift At the moment we only have this for `DockerCreateContainer` but not for `DockerStartContainer`. Would you be interested in working on a pull request to support it? Part of this pull request would be to expose all configuration options for starting a container and not just the `PortBindings`.

Issue #25, opened 2015-02-04
Consider adding project group information to the default tag value


holgerstolzenberg:
If a `group` property is defined for the project, is should be included in the default `tag` value.

```groovy
docker {
  javaApplication {
    tag = "$group/$project.name:$version"
  }
}
```

This will provide more sophisticated image names.

bmuschko:
I am not sure if I want this to be the default. What's the reasoning? It will pretty much break all consumers of this plugin that currently rely on the existing default tag.

It's actually very simple to set this tagging strategy for your project as shown in your code snippet above.

holgerstolzenberg:
You are right, I haven´t thought of backwards compatibility so far. I think that having the group within the default tag is closer to the Docker standard naming the community is cultivating right now. I find it much more pleasant to issue a `docker images` command and grep for a specific group. This gets really interesting within a companies private repo when trying to find a specific app from that company.

bmuschko:
Fair enough. That makes sense. Can you touch up your pull request based on my comments?

PR #26, opened 2015-02-04
Add project group to javaApplication tag if present.

holgerstolzenberg:
If the user specifies a property 'group', as it is done with most projects that are published somewhere, This property will be used as prefix of the docker image tag.

This will result in an image tag like 'org.github/any-image:$version'.

Fixes gh-25.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-1.12-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-1.12-bin.zip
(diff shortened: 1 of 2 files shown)


holgerstolzenberg:
I do have only very small groovy knowledge. The `static` stuff was an suggestion from IntelliJ, why not?

I have pushed your suggested changes.

What do you mean by additional external dependencies?

holgerstolzenberg:
Removed the `static` modifiers and pushed again.

bmuschko:
> I do have only very small groovy knowledge. The static stuff was an suggestion from IntelliJ, why not?

I don't know why IDEA suggested that. You would not be able to invoke the method via the class as the methods are `private`. Therefore, adding `static` doesn't make sense to me.

> What do you mean by additional external dependencies?

By external dependencies I mean Google Guava for example which you are referencing here. It might be included with the Gradle API but I don't want this plugin to break if the used classes changes in any form in future versions of Gradle.

holgerstolzenberg:
Okay understood. I will remove the `Strings` class.

As with the static stuff, I think the IDE suggests it because these methods do not use any instance references. Having a JDK lower than 1.8, this will have an impact on the HEAP space, where `static`stuff will be put into permgen space and will therefore not undergo the standard garbage collection.

bmuschko:
Thanks for your pull request. It has been merged.

Issue #28, opened 2015-02-10
Case typo prevents javaApplication configuration to applied


holgerstolzenberg:
There is a typo in `DockerJavaApplicationPlugin.configureExtension`. In line 64 `DockerExtension` is typed with a upper case 'D'. In my multi module project this causes the `javaApplication` configuration to not be applied. Changing that to a lower case 'd' fixes that.

A pull request is on the way.

PR #29, opened 2015-02-10
Fix case typo that causes javaApplication configuration to not be applied

holgerstolzenberg:
This pull request fixes a typo within `DockerJavaApplication.configureExtension()` that causes the given configuration to not be applied within a multi module build with multiple docker submodules.

Fixes gh-28.


holgerstolzenberg:
SNAP CI build fails because there is no docker server available. I do not think that this is related to my change.

bmuschko:
Good catch and thanks for the pull request!

Issue #30, opened 2015-02-12
Missing method to specify port bindings for DockerCreateContainer


jtgasper3:
Hopefully I'm just very dense but after struggling for a while and looking through code, I don't the proper way to map local port 1443 to exposed port 443. The API [1] references:

> PortBindings - A map of exposed container ports and the host port they should map to. It should be specified in the form `{ <port>/<protocol>: [{ "HostPort": "<port>" }] } `Take note that port is specified as a string and not an integer value.

The DockerCreateContainer class[2] doesn't reference this property. *Is portSpecs support to fill this role?*

Thanks for the consideration, and thanks for sharing your plugin.

[1] http://docs.docker.com/reference/api/docker_remote_api_v1.17/
[2] https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy

bmuschko:
You are referencing two different tasks: `DockerCreateContainer` and `DockerStartContainer`. Which one are you referring to? 

At least for `DockerCreateContainer` you can use the property [portSpecs](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy#L31). A good way to find out what properties are exposed is to check the GroovyDocs for each custom task class from the [README.asciidoc](https://github.com/bmuschko/gradle-docker-plugin/blob/master/README.asciidoc) file. 

`DockerStartContainer` doesn't have this property yet. If it is missing at the moment, then we'd need to add it. The fastest way to get it started would be a pull request.

jtgasper3:
Sorry, bad cutting and pasting on my part with too many browser tabs open. I believe it should be on the create. I'll clean up the request to properly reflect the issue. I'll try to get a PR put together.

jtgasper3:
How is portSpecs supposed to work? There is no mention of it in the Docker Remote API, at least not latter versions? I also found old references to it in the docker-java library, but they seem to be very old and not in later versions.

I've attempted to use it "1443:443" and it does not give the desired mapping when running `docker ps`.

bmuschko:
To be honest, I am not sure. I am basically just calling off to the Docker Java library. I think it would be best to ask the question on their project page.

jtgasper3:
[...]

PR #38, opened 2015-03-19
Issue #30: Specify port bindings when starting a container.

todd1251:
I tested a local build of this plugin and was able to specify port bindings when starting a container. Also added an example of the syntax to the "Executing functional tests against a running container" section of README (and `startContainer` example in `DockerWorkflowIntegrationTest`).

The integration tests pass because I'm running boot2docker on OS X and the docker API is not available (most are skipped). I notice the same thing happens in the snap-ci build anyhow, perhaps still TODO?

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -210,6 +210,7 @@ task createMyAppContainer(type: DockerCreateContainer) {
 task startMyAppContainer(type: DockerStartContainer) {
     dependsOn createMyAppContainer
     targetContainerId { createMyAppContainer.getContainerId() }
+    portBindings = ['8080:8080']
 }
 
 task stopMyAppContainer(type: DockerStopContainer) {
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy
@@ -114,6 +114,7 @@ task createContainer(type: DockerCreateContainer) {
 task startContainer(type: DockerStartContainer) {
     dependsOn createContainer
     targetContainerId { createContainer.getContainerId() }
+    portBindings = ['8080:8080']
 }
 """
         expect:
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainerIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainerIntegrationTest.groovy
@@ -17,6 +17,9 @@ package com.bmuschko.gradle.docker.tasks.container
 
 import org.gradle.api.Task
 import com.bmuschko.gradle.docker.tasks.DockerTaskIntegrationTest
+import spock.lang.IgnoreIf
+
+import static com.bmuschko.gradle.docker.AbstractIntegrationTest.isDockerServerInfoUrlReachable
 
 class DockerStartContainerIntegrationTest extends DockerTaskIntegrationTest {
     @Override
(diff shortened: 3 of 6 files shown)


coco1979ka:
We have also implemented the feature - nearly the same solution like yours. Unfortunately 13 hours too late :)  
Our test is a little bit more specific:

     def "port bindings should allow external service access"() {
        given: "container with netcat webserver"
        def createContainerTask = project.task('createContainer', type: DockerCreateContainer) {
            imageId = 'busybox'
            exposedPorts = ["tcp":80]
            cmd = ["/bin/sh","-c","echo -e \"HTTP/1.1 200 OK\\n\\nHallo\" | nc -l -p 80" ]
        }
        createContainerTask.execute()

        when: "container is started with port binding"
        def startContainerTask = project.task('startContainer', type: DockerStartContainer) {
            containerId = createContainerTask.containerId
            portBindings = ["8088:80"]
        }
        startContainerTask.execute()

        then:
        HttpClient client = new DefaultHttpClient()
        HttpGet httpGet = new HttpGet("http://localhost:8088")
        HttpResponse httpResponse = client.execute(httpGet)
        httpResponse.statusLine.statusCode == HttpStatus.SC_OK
		httpResponse.entity.content.getBytes() == "Hallo\n".getBytes()


    }

I don't know if it is enough to just check if the container could be started by checking the containerId. 

I think the next step should be to externalize the docker configuration, to specify docker host, port and certs for the integration tests. I had the same problem and I plan to implement more features like container linking and waiting for service inside the container to be started, it would be very helpful. 

Second thing: My tests need a very long time to execute because gradle pulls the docker-java api from mavenCentral for every single test. I could only solve this by using mavenLocal and put the dependencies there.

bmuschko:
@todd1251 Thanks for the pull request. It was merged. I added some more tests and made minor tweaks.

> The integration tests pass because I'm running boot2docker on OS X and the docker API is not available (most are skipped). I notice the same thing happens in the snap-ci build anyhow, perhaps still TODO?

There's not a good Open Source offering that allows us to run the test against. The tests have a precondition that check whether you have a local Docker instance running. If you don't have Docker running with the expected setup the test will be skipped.

@coco1979ka It would be great to have more tests. Feel free to send another pull request.

> I think the next step should be to externalize the docker configuration, to specify docker host, port and certs for the integration tests.

It's pretty simple to pass certain system properties from the project's `build.gradle` file. We'll just have to work on it. Right now it's assumed that TLS is not enabled and we run them against a local instance.

> Second thing: My tests need a very long time to execute because gradle pulls the docker-java api from mavenCentral for every single test. I could only solve this by using mavenLocal and put the dependencies there.

That shouldn't be the case. Once downloaded the artifacts are reused from the cache.

todd1251:
> ...
> I don't know if it is enough to just check if the container could be started by checking the containerId.

@coco1979ka: Apologies for the duplication of effort! Hope you do merge in the test you've written as it actually proves that the feature works. ;-)

@bmuschko: Thanks! Hope Snap CI adds docker support as the pipelines are awesome. Is CircleCi a possibility (now free for one container and supports docker)?

> > Second thing: My tests need a very long time to execute because gradle pulls the docker-java api from mavenCentral for every single test. I could only solve this by using mavenLocal and put the dependencies there.

> That shouldn't be the case. Once downloaded the artifacts are reused from the cache.

I had the same problem. The artifacts are cached but it looks like each POM is being checked (you can see it with `gradle --info`; strangely `gradle --offline` doesn't seem to make a difference). On my slow internet connection, the integration tests take 30+ minutes to essentially do nothing when docker is not available. Perhaps this is why they also take 8-9 minutes to do the same in Snap CI?

coco1979ka:
@bmuschko for every used dependency I get a message like this:

10:39:19.902 [main] INFO  o.g.i.r.t.DefaultCacheAwareExternalResourceAccessor - Found locally available resource with matching checksum: [https://repo1.maven.org/maven2/org/sonatype/oss/oss-parent/9/oss-parent-9.pom ...]

This is done with every single test and slows down the test. I could not find any information on how to avoid this.

stellingsimon:
The documentation on the homepage wasn't updated. Therefore, if you follow the examples, you will get the following error:

    * What went wrong:
    A problem occurred evaluating project ':myproject'.
    > No such property: portBindings for class: com.bmuschko.gradle.docker.tasks.container.DockerStartContainer_Decorated

Of course, it works fine once you move the `portMappings` definition to the create task.

bmuschko:
@stellingsimon Thanks for letting me know. Fixed.

Issue #41, opened 2015-03-31
Can't push tagged image created with docker-java-application to private repo


david-resnick:
This configuration 
```
docker {
	javaApplication {
		tag = 'private-registry.company.com:5000/namespace/project:latest'
	}
}
```
fails with this outcome:
```
...
:dockerPushImage
Pushing image with name 'private-registry.company.com:5000/namespace/project:latest'.
:dockerPushImage FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dockerPushImage'.
> Invalid repository name (facts-repo:latest), only [a-z0-9-_.] are allowed
```

This can be fixed by overriding the dockerPushImage convention defaults:

```
dockerPushImage.imageName = 'private-registry.company.com:5000/namespace/project'
dockerPushImage.tag = 'latest'
```

The real problem here is the confusing way docker chose to have images named, tagged and pushed. Sometimes a tag is just a tag, while other times it includes the registry, namespace and "repository".

bmuschko:
I agree, it's really a Docker issue. Would you care for providing an integration test for this use case?

david-resnick:
A (currently) failing integration test?

Is there a setup list that describes how I can successfully run the integration tests you already have? I tried running them yesterday and had failures. I also aborted them before they finished because they were taking a very long time.

bmuschko:
> A (currently) failing integration test?

No, it would a a new integration test.

Is there a setup list that describes how I can successfully run the integration tests you already have? 
> Not really. Currently, they assume TLS to be disabled. However, this can be made configurable. Didn't have the time to do it yet.

bmuschko:
This might just be a Docker Java issue (see [issue](https://github.com/docker-java/docker-java/issues/190)).

johnrengelman:
This is working fine for me:
```
project.docker.javaApplication {
  tag = "registry.mydomain.com/myapp:latest"
}

project.tasks.dockerPushImage {
  imageName = project.tasks.dockerBuildImage.getTag().split(':')[0]
  tag = project.tasks.dockerBuildImage.getTag().split(':')[1]
  registry = new DockerRegistry(url: 'http://registry.mydomain.com/v1', username: 'user', password: 'pass', email: 'user@mydomain.com')
}
```

holgerstolzenberg:
Same problem here, but was able to workaround. This is what works for me:

```groovy
docker {
  url = property 'dockerHostUrl'

[...]

PR #43, opened 2015-04-02
Integration test to catch issue #41

david-resnick:
Test:
Can create image with tag for Java application and push to private registry

I haven't been able to run this because of the prerequisites for running the test.


bmuschko:
I added [instructions](https://github.com/bmuschko/gradle-docker-plugin#executing-the-plugins-test-suite-with-custom-configuration) on how to execute the integration tests with custom configuration (server URL, cert path, private registry URL). Let me know if something shouldn't work there.

Wouldn't your test fail either way as it uses the version as part of the assigned tag? Wasn't that the issue you reported? My idea was to add a test that demonstrates the use of your "workaround".

david-resnick:
Thanks, I've tried running the integration tests and opened a separate issue for the failure there. Something strange going on there...

About the test in this pull request: thanks for clarifying. I'll change it once I am able run the tests.

Issue #42, opened 2015-04-01
Creating a container is not offering options for linking container


MartinAhrer:
Are there any plans to include container linking along with container creation? The Java API seems to provide that capability: https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/model/HostConfig.java

bmuschko:
Generally speaking, yes. Do you know with which version of Docker Java this was added? The plugin is still on version 0.10.5. I am waiting for 1.0.1 which is still not release. 1.0.0 had an issue.

Would you be interested in working on a pull request for this?

MartinAhrer:
To me it looks like this feature is already in 0.10.5: https://github.com/docker-java/docker-java/blob/docker-java-0.10.5/src/main/java/com/github/dockerjava/api/model/HostConfig.java. HostConfig is encapsulating much of the more advanced settings.

Ok, yes I can try working on a pull request. It shouldn't be too hard to pass through a collection of links to the Java API.

bmuschko:
As a side note: You'll need to create `HostConfig` via reflection. See [DockerThreadContextClassLoader.groovy](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy) for more info.

MartinAhrer:
I'm already working on it and have already fleshed out the factory for the HostConfig, Links and Link types of the Docker Java API. However I seem to be running into a issue introduced with DockerThreadContextClassLoader. The [Links](https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/model/Links.java) type unlike most other container types does not offer a constructor accepting a List but only a Link[]. 

```groovy
    def createLinks(List<Object> links) {
        def linkClass = loadClass('com.github.dockerjava.api.model.Link')
        def linksClass = loadClass('com.github.dockerjava.api.model.Links')
        def linksArray=Array.newInstance(linkClass, 0)
        Constructor constructor = linksClass.getConstructor(linksArray.class)
        constructor.newInstance((Object)links.toArray(linksArray))
    }
```
Doing the same with classes loaded with ```Class.fromString``` in some extra test project works fine. The easiest thing to do will be asking the Docker Java API maintainer to add the extra constructor to Links. But this will probably take much more time - so I would appreciate your input.

bmuschko:
[...]

PR #44, opened 2015-04-02
Added container name to create container command.

MartinAhrer:
Next, I will start adding container linking as suggested in issue #42

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy
@@ -96,10 +96,14 @@ task inspectImage(type: DockerInspectImage) {
         File imageDir = createDir(new File(projectDir, 'images/minimal'))
         createDockerfile(imageDir)
 
+        def uniqueContainerName = createUniqueContainerName()
+
         buildFile << """
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
 import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
 import com.bmuschko.gradle.docker.tasks.container.DockerStartContainer
+import com.bmuschko.gradle.docker.tasks.container.DockerInspectContainer
+
 
 task buildImage(type: DockerBuildImage) {
     inputDir = file('images/minimal')
@@ -109,6 +113,7 @@ task buildImage(type: DockerBuildImage) {
 task createContainer(type: DockerCreateContainer) {
     dependsOn buildImage
     targetImageId { buildImage.getImageId() }
+    containerName = "$uniqueContainerName"
 }
 
 task startContainer(type: DockerStartContainer) {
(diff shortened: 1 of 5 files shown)


bmuschko:
Thanks for your pull request. Would you mind adding an integration test as well? Please have a look at [DockerWorkflowIntegrationTest.groovy](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy).

MartinAhrer:
Sure, didn't spot that test before. Thanks for the pointer! Will update my pull request.

MartinAhrer:
I have added your requested integration test. Actually I enhanced your ```"Can build an image, create and start a container"``` test. It is now adding a container name and then inspecting the container. However, I still can't execute the tests they are skipped. So the pull-request is not yet ready for merge !

src/integTest/groovy/com/bmuschko/gradle/docker/ToolingApiIntegrationTest.groovy
```diff
@@ -60,4 +60,8 @@ repositories {
     protected String createUniqueImageId() {
         "gradle/${UUID.randomUUID().toString().replaceAll('-', '')}"
     }
+
+    protected String createUniqueContainerName() {
```
bmuschko: Instead of adding a new method I think we should simply rename the existing method `createUniqueImageId ` to something else. Probably just `createUniqueId`.

src/integTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowIntegrationTest.groovy
```diff
@@ -116,9 +119,15 @@ task startContainer(type: DockerStartContainer) {
...
+
+task inspectImage(type: DockerInspectImage) {
+    dependsOn buildImage
+    targetImageId { buildImage.getImageId() }
+}
 """
         expect:
-        runTasks('startContainer')
+        GradleInvocationResult result = runTasks('startContainer', 'inspectImage')
+        result.output.contains("""Name": "/$uniqueContainerName""")
```
bmuschko: This doesn't need to be a multi-line String.

bmuschko:
You can now adapt to your Docker setup (see FAQ for more information). To be able to execute the test, you'll have to make sure that the Docker server URL is resolvable.

src/integTest/groovy/com/bmuschko/gradle/docker/ToolingApiIntegrationTest.groovy
MartinAhrer: Sure, I will refactor that. Didn't want to interfere too much with the other test code.

MartinAhrer: Finally I realized that a container name may not contain a '/' like the image id. So I have kept the 2 methods.

MartinAhrer:
I had to add a DockerInspectContainer task type in order to facilitate testing. However I was wondering of what use those Inspect tasks are as they are just logging some image/container attributes. How about exposing the full inspect response to the task client? Generally speaking I think a inspect task will be useful for querying any image/container attribute.
Certainly it is a difficult decision what format the response of the inspect task would be? JSON or some typed data structure. The latter would mean that it is a lot of work to keep it in sync with the docker-java implementation.

bmuschko:

MartinAhrer:
Thanks for supporting, expect a new pull-request for container linking soon.

PR #46, opened 2015-04-09
Add a dockerFile option to the DockerBuildImage task.

llamahunter:
Separate the build context from the Dockerfile location.


llamahunter:
So, I added an integration test, but I was unable to run the integration tests on my machine.  Also, it's really unclear to me how the dependency on docker-java is resolved.

llamahunter:
something is wrong with this...  perhaps version of docker-java?

PR #47, opened 2015-04-10
Add optional dockerFile option to the DockerBuildImage task.

llamahunter:
Separate the build context from the Dockerfile location.
Update version of docker-java to 1.1.0

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageSeparatedIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageSeparatedIntegrationTest.groovy
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.bmuschko.gradle.docker.tasks.image
+
+import com.bmuschko.gradle.docker.tasks.DockerTaskIntegrationTest
+import org.apache.commons.io.FileUtils
+import org.gradle.api.Task
+
+class DockerBuildImageSeparatedIntegrationTest extends DockerTaskIntegrationTest {
+    File dockerInputDir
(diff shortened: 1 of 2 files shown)


llamahunter:
Ok... think this works now.  Still unable to run the integration test, tho.  But I did test that I was able to set the dockerFile in my own gradle scripts.  Unfortunately, 'docker' itself doesn't allow setting a Dockerfile outside of the build context, but at least now it is possible to set it to a file named anything anywhere within the context.

bmuschko:
What's the reason you cannot run the integration tests?

llamahunter:
Ah.. did not scroll all the way to the bottom of the readme, it seems.  I'll give that a try.  I'm running boot2docker, for which I haven't yet figured out how to make non TLS.

bmuschko:
I added this today so you couldn't have seen this earlier. With that you should be able to use TLS as well.

BTW: In case you want to disable TLS, [the documentation](https://github.com/bmuschko/gradle-docker-plugin#working-with-a-docker-instance-without-tls) shows how to do it.

bmuschko:
I just released a new version of the plugin that uses the latest Docker Java version. Please rebase your pull request.

llamahunter:
Horray... integration tests passed.

llamahunter:
rebased (I think).  integration tests still seem to be passing.

bmuschko:
Thanks for the pull request. I think I want to solve this slightly differently by just having a single property `File dockerFile`. I am going to make the relevant changes and will let you know when this goes out.

llamahunter:
No problem.  I have a local build in my own maven repo.  Look forward to your solution!  To make it *really* work the way I think it should, however (arbitrary location of Dockerfile completely separate and outside the build context) will likely require a change in the way docker-java creates its staged InputStream in Dockerfile.ScannedResult.buildDockerFolderTar().

llamahunter:
Hmm... but if you get rid of inputDir, aren't you just going to continue with conflating the root of the build context with the location of the Dockerfile?

bmuschko:
I had actually thought about a different problem you were trying to solve - allowing the use of other files than the standard convention as Dockerfiles. This makes more sense now. I am going to have a look at this again soon.

llamahunter:
Well, that too.  Actually, looking at the current code in docker-Java, I think people could pass the path to the "otherly named" Dockerfile file as inputDir (instead of a directory), and it would do the right thing.

bmuschko:
That was how I wanted to implement it before I noticed that you had a different requirements. Generally speaking, allowing a parameter to be different things is a bad practice IMHO. But that's something the Docker Java library has to deal with and this plugin can hide.

llamahunter:
Wasn't a fan of the docker-java approach either. Thus why I added a new parameter.

llamahunter:
Rebased to HEAD. Any chance this will get pulled someday?

bmuschko:
Sorry for the delay. I'll have to review this again soon.

bmuschko:
Thanks for the pull request. It has been merged. I need to reshuffle the tests a bit. For now I remove it.

Issue #48, opened 2015-04-13
Can't run integration tests


david-resnick:
I'm trying to run the integration tests against docker running with boot2docker and a private docker registry that is running "insecurely" at port 5000.

./gradlew build -DdockerServerUrl=https://192.168.59.103:2376 -DdockerCertPath=/Users/dresnick/.boot2docker/certs/boot2docker-vm -DdockerPrivateRegistryUrl=private-registry.company.com:5000

Some tests fail with this:
```
com.bmuschko.gradle.docker.AbstractDockerRemoteApiTaskIntegrationTest > Can create and execute custom remote API task with default extension values STANDARD_OUT
    21:56:30.333 [Test worker] DEBUG org.gradle.tooling.GradleConnector - Connecting from tooling API consumer version 2.3
    21:56:30.401 [Connection worker] DEBUG o.g.t.i.c.l.DefaultToolingImplementationLoader - Using tooling provider from Gradle distribution 'https://services.gradle.org/distributions/gradle-2.3-bin.zip'
[...]

PR #57, opened 2015-04-21
New task for copying files from a container to a host.  Fixes #56

### LINKED ISSUE #56: Add support for DockerCPContainer to copy files out of a container
I have a need to copy files that are created within the container as part of a gradle build process.  Supporting the docker cp command would be useful via this API.

petehayes:
I wasn't successful in running the integration tests on my machine so I am not sure if the one I added will work.  Otherwise, this will support copying files out of a container and storing them into a tar file on the host with the option of gzip compression.

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -101,15 +101,16 @@ The plugin provides the following custom task types for managing containers:
 
 [options="header"]
 |=======
-|Type                                                                                                                                                              |Description
-|link:http://bmuschko.github.io/gradle-docker-plugin/docs/groovydoc/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.html[DockerCreateContainer]   |Creates a container.
-|link:http://bmuschko.github.io/gradle-docker-plugin/docs/groovydoc/com/bmuschko/gradle/docker/tasks/container/DockerInspectContainer.html[DockerInspectContainer] |Returns low-level information on the container.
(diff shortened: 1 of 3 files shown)


bmuschko:
Thanks for the pull request. I'd like to ask you for some minor cleanup before I merge it:

* Would you mind converting tabs into spaces?
* Remove semicolons

petehayes:
I pushed those changes.  Thanks.

bmuschko:
Thanks. I merged the pull request.

Two comments:

- The provided integration test didn't work. It threw a `com.github.dockerjava.api.NotFoundException`. We should add a test to `DockerWorkflowIntegrationTest` that creates a new container first and tries to copy the file.
- The response of the task should be handled by an instance of `com.bmuschko.gradle.docker.response.ResponseHandler`. The appropriate code should be moved there.

Would you be interested in working on any of these items in a separate pull request? At the moment I am not even sure if the task works properly.

petehayes:
I tried to get the integration tests running but I couldn't figure out what was needed to do so.  I did test the task manually though.  If I could get help on how to run the tests properly, then I would make the necessary changes.

bmuschko:
I still have to address some issues with running the integration tests with against a TLS-enabled Docker instance (see issue #48). Feel free to help out with that as well. Generally, you can run the tests without TLS.

petehayes:
OK. I'll try running them again.

PR #79, opened 2015-06-24
Add add-host parameter option

jonathantait:
Added a call to the 'ExtraHosts' option within docker remote api to enable containers to be run that will edit the hosts file.

This is the equivalent to using the docker run command with the 'add-host' parameter.


bmuschko:
Thanks for the pull request. It has been merged. I will actually turn the `String[]` into a `List<String>` as it is easier to work with.

jonathantait:
No problem, thanks for the great plugin!

PR #82, opened 2015-06-30
apply the docker-remote-api plugin in the examples

fpape:
in the examples be more explicit to apply the docker-remote-api plugin


bmuschko:
Thanks for the pull request. It has been merged. Generally speaking the documentation says that you should apply one of the plugins earlier. However, I've seen people miss this more often than once. I guess it makes it more clear then.

PR #86, opened 2015-07-21
Adds Dockerfile support for the LABEL instruction

steve-perkins:
Adds support for key-value instruction types in general, with a specific implementation for the LABEL Dockerfile instruction.

See additional comments on specific lines within the pull request.


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -373,6 +373,30 @@ class Dockerfile extends DefaultTask {
         }
     }
 
+    static abstract class MapInstruction implements Instruction {
```
steve-perkins: Since the LABEL instruction accepts a series of key-value pairs, I decided to use a Map data type for representing this (I would have thought twice with plain Java, but the Groovy syntax for working with Maps is very clean and easy).

I wondered whether it makes sense to create a base class for this instruction type at all, since LABEL is (for now) the only Dockerfile instruction that works with key-value pairs.  However, I decided to aim for consistency with the other instruction types here... and who knows what additional instruction will be added to the Dockerfile spec in the future as it matures.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -373,6 +373,30 @@ class Dockerfile extends DefaultTask {
...
+        final Object command
+
+        MapInstruction(Map<String, String> command) {
+            this.command = command
+        }
+
+        MapInstruction(Closure command) {
```
steve-perkins: Again, I just went for consistency here.  I'm not using the closure-constructor form in my `LabelInstruction` subclass, but I see that the abstract classes are generally built to support this even if many of the implementation subclasses make no use of it.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -373,6 +373,30 @@ class Dockerfile extends DefaultTask {
...
+            this.command = command
+        }
+
+        @Override
+        String build() {
+            if (command instanceof Map<String, String>) {
+                String args = ""
+                ((Map<String, String>) command).each { key, value -> args += " \"$key\"=\"$value\"" }
+                keyword + args
```
steve-perkins: This was mildly concerning to me.  Originally I was inserting a space between the keyword (e.g. "LABEL") and the arguments.  However, the unit tests were failing due to an extra space character... because something else is apparently already adding a space on the end of the keyword.  I couldn't quite figure that out (Groovy isn't my native tongue), so I'm relying on it here and assuming that the space will be present.

src/test/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileTest.groovy
```diff
@@ -11,29 +11,31 @@ class DockerfileTest extends Specification {
...
+        new CopyFileInstruction({ 'config.xml' }, { '/test' })                      | 'COPY'       | 'COPY config.xml /test'
+        new EntryPointInstruction('java', '-jar', '/opt/jenkins.war')               | 'ENTRYPOINT' | 'ENTRYPOINT ["java", "-jar", "/opt/jenkins.war"]'
+        new EntryPointInstruction({ ['java', '-jar', '/opt/jenkins.war'] })         | 'ENTRYPOINT' | 'ENTRYPOINT ["java", "-jar", "/opt/jenkins.war"]'
+        new VolumeInstruction('/jenkins')                                           | 'VOLUME'     | 'VOLUME ["/jenkins"]'
+        new UserInstruction('ENV JAVA_HOME /usr/java')                              | 'USER'       | 'USER ENV JAVA_HOME /usr/java'
+        new WorkDirInstruction('/some/dir')                                         | 'WORKDIR'    | 'WORKDIR /some/dir'
+        new OnBuildInstruction('ENV JAVA_HOME /usr/java')                           | 'ONBUILD'    | 'ONBUILD ENV JAVA_HOME /usr/java'
+        new LabelInstruction([ 'description':'Single label' ])                      | 'LABEL'      | 'LABEL "description"="Single label"'
```
steve-perkins: Only these bottom two lines have changed.  The lines above are different only in that I adjusted the column width.

bmuschko:
Thanks for the pull request. It has been merged. Made some minor changes.

Issue #99, opened 2015-08-19
Only treat pem files in DOCKER_CERT_PATH as an input


ajmath:
When switching to docker-machine, I noticed that it uses a single flat directory for each machine's configuration.  For virtualbox, this includes things like the vm disk image.  With the certPath defined as a input directory, this means changes to any file in that directory will need to be scanned by gradle.  For large files, this can take several seconds.

cdancy:
Closing as per #100. If we need to re-open we can.

PR #100, opened 2015-08-19
Only include pem files in cert_path as input

ajmath:
Addresses issue #99.

Removes `@InputDir` annotation from certPath and creates a setter that sets task inputs to only contain pem files in that directory.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -36,7 +36,6 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     /**
      * Path to the <a href="https://docs.docker.com/articles/https/">Docker certificate and key</a>.
      */
-    @InputDirectory
     @Optional
     File certPath
 
@@ -49,6 +48,13 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         }
     }
 
+    public void setCertPath(File certPath) {
+      File[] pemFiles = certPath.listFiles().findAll( { it.name.endsWith("pem") } )
+      getInputs().files(pemFiles)
+
+      this.certPath = certPath
+    }
+
     private DockerClientConfiguration createDockerClientConfig() {
         DockerClientConfiguration dockerClientConfig = new DockerClientConfiguration()
         dockerClientConfig.url = getUrl()
@@ -58,4 +64,3 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
 
     abstract void runRemoteCommand(dockerClient)
 }
-
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -48,11 +48,12 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         }
     }
 
-    public void setCertPath(File certPath) {
-      File[] pemFiles = certPath.listFiles().findAll( { it.name.endsWith("pem") } )
-      getInputs().files(pemFiles)
-
-      this.certPath = certPath
+    @InputFiles
+    public File[] getCertPathKeys() {
+      if (certPath) {
+        return certPath.listFiles().findAll( { it.name.endsWith("pem") } )
+      }
+      return []
     }
 
     private DockerClientConfiguration createDockerClientConfig() {


src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -36,7 +36,6 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     /**
      * Path to the <a href="https://docs.docker.com/articles/https/">Docker certificate and key</a>.
      */
-    @InputDirectory
```
bmuschko: This annotation should go on to a `getCertPath()` method.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -49,6 +48,13 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         }
     }
 
+    public void setCertPath(File certPath) {
+      File[] pemFiles = certPath.listFiles().findAll( { it.name.endsWith("pem") } )
+      getInputs().files(pemFiles)
```
bmuschko: I prefer the annotation.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -49,6 +48,13 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         }
     }
 
+    public void setCertPath(File certPath) {
+      File[] pemFiles = certPath.listFiles().findAll( { it.name.endsWith("pem") } )
```
bmuschko: You can assign this as default value for the field `certPath`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -49,6 +48,13 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         }
     }
 
+    public void setCertPath(File certPath) {
```
bmuschko: Setter method won't be needed as Groovy generates it.

bmuschko:
Thanks for the pull request. I added some comments. Would you mind fixing them and updating the PR?

ajmath:
Thanks.  I've updated the PR

bmuschko:
Thanks for the pull request. Before I'd merge this one, I'd like to have it covered by functional tests. Do you think you could provide the following cases? The implementation would basically extend `AbstractFunctionalTest`.

* Create task of type `AbstractDockerRemoteApiTask`, set the cert path
   * execute it once, the action is executed
   * execute it a second time, the action is up-to-date
   * the cert path is passed on properly
*  Create task of type `AbstractDockerRemoteApiTask`, do not set the cert path
   * execute it once, the action is executed
   * execute it a second time, the action is up-to-date

cdancy:
@ajmath any interest in addressing the above noted issues? I know its been a while but I'm here helping out at and we could try to move this along if you're still up for it.

@bmuschko I don't think this PR should have any bearing on your recent thoughts surrounding cert paths but if they do please don't hesitate to jump in.

cdancy:
Looking over things it appears that docker is standardizing on "~/.docker" for storing your certs. Maybe we can default on that here but then only use it should we have an https url?

ajmath:
@cdancy The issue here is that large files like VM disks in the `certPath` are causing extra long times when gradle does it scan of input files to see if they've changed.  It seems the common practice is to use the environment variable `DOCKER_CERT_PATH` for this.  The issue though is that docker-machine currently outputs the `DOCKER_CERT_PATH` environment variable as the same one that contains the vm image `~/.docker/machine/machines/default`.  

However, in the current version of docker-machine there appears to be a different certPath that only has the certs `~/.docker/machine/certs`.  Setting `certPath` on one of my gradle docker builds still worked as expected.  It's unfortunate that we can't rely on `DOCKER_CERT_PATH` here as I can't rely on devs having installed docker-machine, and thus can't assume `~/.docker/machine/cert` will always work.

TLDR: I think we can close this PR as there appears to be a workaround that doesn't require goofy file extension mapping.

cdancy:
@ajmath gotcha and thanks for the insight. You could probably get away, assuming you could enforce, devs to create "~/.docker-java.properties" and within here hardcode whichever location you look and/or want to standardize on. We do something similar in that we provide a container with environment defaults, property files, and the aforementioned '~/.docker-java.properties' to get everyone on the same page. That may/may-not suffice for your scenario but it is an option.

PR #102, opened 2015-08-27
Correct the withVolumesFrom call on DockerCreateContainer task which needs to get a VolumesFrom[] array as the parameter

uglyog:
DockerCreateContainer volumesFrom field is a string array. Trying to use this field results in the following error:

```
No signature of method: com.github.dockerjava.core.command.CreateContainerCmdImpl.withVolumesFrom() is applicable for argument types: ([Ljava.lang.String;) values: [[transaction-service-mysql-data]]
  Possible solutions: withVolumesFrom([Lcom.github.dockerjava.api.model.VolumesFrom;), getVolumesFrom(), withVolumes([Lcom.github.dockerjava.api.model.Volume;)
```

This is due to the `CreateContainerCmdImpl.withVolumesFrom()` method requiring a VolumesFrom[] array instead.

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
@@ -259,6 +259,28 @@ class DockerThreadContextClassLoaderIntegrationTest extends AbstractIntegrationT
         instance.configs.size() == 2
     }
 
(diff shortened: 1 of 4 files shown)


bmuschko:
Thanks for the pull request. Would you mind adding test cases to `DockerThreadContextClassLoaderIntegrationTest` and extend the existing `DockerCreateContainerIntegrationTest`?

uglyog:
Test added. I also noticed the constructor was not handling the access modes, so I fix that.

bmuschko:

Do you think you could also add a functional test for this use case to `DockerWorkflowFunctionalTest`?

PR #109, opened 2015-09-18
Update to latest docker-java 2.1.1

llamahunter:
Addressing issue https://github.com/docker-java/docker-java/issues/322

PR #110, opened 2015-09-22
updated the docker java version from 2.1.0 . It has a bug where the Rep…

jrhoward:
…onse item is missing fields. THis causes (at least) a pull request to fail with unknown field exceptions

see https://github.com/docker-java/docker-java/commit/1d38e935efe47c2b316b083358af78f141c68d27


bmuschko:
Duplicate of #109.

Issue #81, opened 2015-06-29
On support for setting log-driver for DockerCreateContainer task


cdancy:
This is mainly WRT this known issue:

https://github.com/docker/docker/issues/9139

Because we are using this plugin in a CI environment we are often running into OOM problems due to the way docker is handling, or not handling, logs. To get around this one can set '--log-driver=none' on the command line so docker is not saving the logs. We are already recording the logs anyway so it's not necessary for docker to keep these logs around and run out of memory. The docker remote api exposes this at HostConfig.LogConfig which can be found here:

https://docs.docker.com/reference/api/docker_remote_api_v1.19/#create-a-container

Is it possible we could expose this through the DockerCreateContainer task? Maybe something as simple as a variable 'disableLogging' which is set to false by default but if true will pass in the necessary json.

Thoughts?

bmuschko:
I guess it could be added through the [HostConfig parameter](https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/command/CreateContainerCmd.java#L58) in Docker Java. If you send me a pull request for it I will merge.

cdancy:
@bmuschko update: change has been made and committed to docker-java and now we are just waiting on a new release. Once that happens we will have to wait for you to re-base off of the updated docker-java and then I can make the change here.

https://github.com/docker-java/docker-java/issues/218

cdancy:
@bmuschko looks like you just released 5 days ago but the new docker-java was released 3 days ago! Does the plugin pull in the latest docker-java library?

https://github.com/docker-java/docker-java/releases/tag/docker-java-2.0.0

bmuschko:
I won't be able to get to it till next week. You might want to send a pull request to speed up the process.

bmuschko:
I integrated the latest Docker Java version. You can go ahead and work on a pull request for this issue.

cdancy:
@bmuschko I seen that and thank you! I've got the code teed up from a couple months ago. Let me pull the latest changes and look things over and send in a PR.

cdancy:
@bmuschko thoughts?

PR #118, opened 2015-10-22
ISSUE-81: expose support for LogDriver

cdancy:
PR address #81 . @bmuschko take a look at the implementation and tell me what you think. I went the closure route, and subsequently needing that closure to return 2 values (1 for logtype and 1 for optional-parameters), as I didn't see any other sensible way of exposing this functionality whilst still keeping the client/caller code sane.

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
@@ -221,6 +221,24 @@ class DockerThreadContextClassLoaderIntegrationTest extends AbstractIntegrationT
         instance[1].volume.path == '/other/volume'
     }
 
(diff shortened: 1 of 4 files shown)


cdancy:
Hung here for me every other time or so I ran tests. Not sure what that's about but they did succeed when they did go through.

cdancy:
@bmuschko any thoughts here?

bmuschko:
We should represent `LogConfig` as a class here instead of a `Closure`. That will make it more clear to consumers on what to pass in here. The instance should be annotated with `@Nested` in `DockerCreateContainer`.

    class LogConfig {
        @Input String type
        @Input Map<String, String> config = [:]
    }

Could you please touch up the pull request?

cdancy:
@bmuschko just got back to this now. It hadn't even occured to me to use another class for LogConfig as I've never seen class instantiation on a task parameter. I took your example above and put it as an inner class to DockerCreateContainer as that's the only place it can be used anyway. Fixed another line or 2 to get things working. All tests passed as expected.

bmuschko:

PR #135, opened 2015-12-31
FIX for ISSUE-134

cdancy:
Refactor is intended to make DockerCopyFileFromContainer behave exactly like "docker cp". Because the API call to docker cp returns a tar stream I think it best, in keeping with the semantics of the docker remote API and what users would expect, to also write out ONLY the tar file to disk, and not gzip it, assuming the user requested so.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
@@ -35,8 +35,7 @@ class DockerCopyFileFromContainer extends DockerExistingContainer {
      * This is consistent with how 'docker cp' behaves.
      */
     @Input
-    @Optional
-    String hostPath
+    String hostPath = project.projectDir.path
 
     /**
      * Whether to leave file in its compressed state or not.
@@ -52,9 +51,6 @@ class DockerCopyFileFromContainer extends DockerExistingContainer {
 
     @Override
     void runRemoteCommand(dockerClient) {
-
-        if (!hostPath)
-            hostPath = "${project.projectDir}"
 
         def containerCommand = dockerClient.copyFileFromContainerCmd(getContainerId(), getRemotePath())
         logger.quiet "Copying '${getRemotePath()}' from container with ID '${getContainerId()}' to '${getHostPath()}'."
@@ -97,15 +93,19 @@ class DockerCopyFileFromContainer extends DockerExistingContainer {
         // the regular file itself exists, delete to avoid clobbering.
         if (hostDestination.exists()) {
             if (!hostDestination.isDirectory())
-                hostDestination.delete()
+                if(!hostDestination.delete())
+                    throw new GradleException("Failed deleting previously existing file at ${hostDestination.path}")
+
         } else {
-            hostDestination.parentFile.mkdirs()
+            if (!hostDestination.parentFile.mkdirs())
+                throw new GradleException("Failed creating parent directory for ${hostDestination.path}")
         }
 
-        new File(compressedFileLocation,
-                compressedFileName).
-                withOutputStream { it << tarStream }
-
+        def tarFile = new File(compressedFileLocation, compressedFileName)
+        if(!tarFile.createNewFile())
+            throw new GradleException("Failed creating file at ${tarFile.path}")
+        else
+            tarFile.withOutputStream { it << tarStream }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cdancy:
@bmuschko build and tests run fine here locally. Running twice now I see the same error when built through snap-ci. Looking at the tests its obvious there are issues creating the various docker-java classes but looking at some of those classes they haven't changed in months (some years). I have to imagine then this has more to do with how we are loading docker-java at runtime and with a separate classloader.

cdancy:
Hmmm ... seems my latest refactor has caused the issues to go away? They didn't appear to be related...

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
```diff
@@ -15,55 +15,178 @@
...
+     * for a regular file or directory. Thus, we can give the caller
+     * back the tar file as is or explode it to some location like
+     * 'docker cp' does.
+     */
     @Input
     @Optional
     Boolean compressed = Boolean.FALSE
 
     @Override
     void runRemoteCommand(dockerClient) {
```
bmuschko: Can we simplify the logic a bit breaking it up into methods? It looks fairly complex.

cdancy: Broke things up into 2 methods and added more documentation on what exactly is being done and why. It is complex but written so to avoid copying around potentially large files on the host (using cheap renames instead) as we try to create the directory structure the user requested and would expect if doing the same operation via 'docker cp'. Have a look and let me know what you think.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
```diff
@@ -15,55 +15,209 @@
...
-        logger.quiet "Copying '${getResource()}' from container with ID '${getContainerId()}' to '${getDestFile()}'."
 
-        getDestFile().withOutputStream  { OutputStream out ->
-            def input
+        if (!hostPath)
+            hostPath = "${project.projectDir}"
```
bmuschko: Should this be `project.projectDir.path` instead? We could simply assign this as default value to the property `hostPath` and make it non-optional.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
```diff
@@ -15,55 +15,209 @@
...
+        new File(compressedFileLocation,
+                compressedFileName).
+                withOutputStream { it << tarStream }
+
+    }
+
+    /**
+     * Copy regular file or directory from container to host
+     */
+    private void copyFile(InputStream tarStream, File hostDestination) {
```
bmuschko: We'll need to break this up into methods. The method is more than 100 lines long and really hard to understand as one block of code. A good rule of thumb here is to break it up into methods where ever you added a comment.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
```diff
@@ -15,55 +15,209 @@
...
+            tempDir.eachFileRecurse(FileType.FILES) { fileCount++ }
+            if (fileCount == 0) {
+
+                logger.quiet "Nothing to copy."
+
+            } else if (fileCount == 1) {
+
+                // ensure regular file does not exist as we don't want clobbering
+                if (hostDestination.exists() && !hostDestination.isDirectory())
+                    hostDestination.delete()
```
bmuschko: We'll need to make sure the return value of `delete()` and `mkdirs()` is handled. Otherwise, we'll swallow an error and the task will fail. Need to check for result and throw `GradleException` if not true.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
cdancy: I agree. Changing now.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
cdancy: Addressing now.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
cdancy: @bmuschko take a look. Should be more digestible now.

bmuschko:
Thanks for touching up. It has been merged. Two more comments for future changes:

- I personally prefer typing in Groovy over the `def` keyword where ever the typ is clear. It just makes the code easier to read and the API easier to comprehend. I'd love to stick to it.
- I like to wrap one line `if` statements with curly braces. It also make is much easier to parse.

Issue #123, opened 2015-11-26
Support for docker step reports for DockerBuildImage


rfhayashi:
When using `DockerBuildImage` task, I miss docker step reports, example:

    Step 1 : FROM jenkins:1.625.1

       ---> 496f04f9620b

    Step 2 : COPY plugins.txt /usr/share/jenkins/plugins.txt

       ---> Using cache

       ---> ab0554ac035b

    Step 3 : RUN /usr/local/bin/plugins.sh /usr/share/jenkins/plugins.txt

       ---> Using cache

 	   ---> 228d801a2adc

	Successfully built 228d801a2adc

As I could find, it would be a case of subclassing `BuildImageResultCallback`, overwritting `onNext` and log `BuildResponseItem.stream`. The thing is that `BuildImageResultCallback` is instantiated by reflection as docker-java dependency is injected dinamically (btw, I didn't understand why, probably my gradle newbieness). I'm not sure what would be the best way of subclassing using reflection.

What do you think? I'd be willing to work on that, but I'd like to discuss the best approach first.

bmuschko:
The reason for having to use a custom classloader in the plugin is classloader isolation. If you don't do that you have Gradle core libraries bleeding into the classpath of your plugin and might cause version conflicts with the libraries provided by Docker Java. This something that needs to be improved in Gradle core.

I never had the need to create a subclass via reflection. I guess you'd have to look into that. Docker Java changes the API all the time. In fact in an earlier version of the plugin (that used an earlier version of Docker Java) the use case you describe was already solved. However, new features in Docker Java required an upgrade which now makes it more complex to create a custom callback.

Feel free to work on a pull request for this. I'd be happy to review and merge.

cdancy:
Closed due to #145

PR #141, opened 2016-01-25
Issue #123 - Support for docker step reports for DockerBuildImage

rfhayashi:
I've managed to implement using cglib. It feels a little awkward, but it works. I created a `printStream` option to turn printing the stream on and off. I'd vote to have it on by default, but I left it off to keep the current behaviour. I created a `createPrintStreamProxyCallback` on `ThreadContextClassLoader` that can be used to implement the same behaviour for other async tasks. If you like this implementation I can work on that afterwards.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -25,6 +25,7 @@ task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
         BuildResult result = build('buildImage')
 
         then:
+        !result.standardOutput.contains('Step 1 : FROM ubuntu:12.04')
         result.standardOutput.contains('Created image with ID')
     }
 
@@ -39,11 +40,11 @@ task dockerFile(type: Dockerfile) {
 
 task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
     inputDir = file("build/docker")
-    printStream = true
 }
 """
         when:
-        BuildResult result = build('buildImage')
+        BuildResult result = build('buildImage', '-i')
+        println result.standardOutput
 
         then:
         result.standardOutput.contains('Step 1 : FROM ubuntu:12.04')
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -23,7 +23,6 @@ task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
 """
         when:
         BuildResult result = build('buildImage')
-        println result.standardOutput
 
         then:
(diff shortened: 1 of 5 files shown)


cdancy:
I like it but would vote to make this the default. It mimics what one would expect when using OOTB docker. 

@bmuschko thoughts?

rfhayashi:
@cdancy, would we keep the `printStream` option?

cdancy:
@rfhayashi yes as it gives people the option to toggle the verbosity. I personally like, and at times need, to see the output but I can imagine for some contexts it may just be noise and not anything folks would like to litter their logs with.

rfhayashi:
:+1:, I'll wait @bmuschko comments to make the changes all at once.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+    from 'ubuntu:12.04'
+}
+
+task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
+    inputDir = file("build/docker")
+}
+"""
+        when:
+        BuildResult result = build('buildImage')
+        println result.standardOutput
```
bmuschko: Let's remove that.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+}
+
+task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
+    inputDir = file("build/docker")
+    printStream = true
+}
+"""
+        when:
+        BuildResult result = build('buildImage')
+        println result.standardOutput
```
bmuschko: Let's remove that.

bmuschko:
I'd personally rather route the output through Gradle's logging system than exposing a new property just for that purpose e.g. Run with `-i` then print the message, without don't print it. Then you won't need that flag on any task.

cdancy:
@bmuschko does that mean you are in favor of enabling this functionality by default and then replace the println's with 'logger.quiet'? 

Having the output OOTB is what docker users would expect and getting at that output, via gradle's "-i", is what gradle users would expect. I'm in favor if this is the case.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
rfhayashi: My bad, forgot to remove it.

rfhayashi:
I changed the code to use Gradle's logging system. I'm not totally familiar with it. Can you review if that was the idea? I'm also thinking of removing `createPrintStreamProxyCallback` from `ThreadContextClassLoader` and add a `Logger` parameter to `createBuildImageResultCallback`, so the callback would always be created with the print stream proxy. Do you agree with that?

Ps: Should I rebase or merge to resolve the conflicts with master?

cdancy:
@rfhayashi looks like you need to rebase

cdancy:
Closing due to #145

Issue #67, opened 2015-05-05
Gradle docker plugin latency


nucatus:
Any idea why commands issued to the docker host (over https) have a much higher latency than the commands issued using the docker client (over tcp)? I put together a small POC to see how the plugin is performisg ->> [docker-gradle-POC](https://github.com/nucatus/gradle-docker-poc) hoping to find out where that latency is coming from and this is as far I could get:
```
09:25:54.950 [DEBUG] [org.apache.http.impl.conn.PoolingHttpClientConnectionManager] Connection leased: [id: 0][route: {s}->https://192.168.59.103:2376][total kept alive: 0; route allocated: 1 of 2; total allocated: 1 of 20]
09:25:54.953 [DEBUG] [org.apache.http.impl.execchain.MainClientExec] Opening connection {s}->https://192.168.59.103:2376
09:25:54.955 [DEBUG] [org.apache.http.conn.HttpClientConnectionManager] Connecting to /192.168.59.103:2376
09:26:00.027 [DEBUG] [org.apache.http.impl.execchain.MainClientExec] Executing request POST /build?t=igenox/docker-poc&rm=true HTTP/1.1
09:26:00.027 [DEBUG] [org.apache.http.impl.execchain.MainClientExec] Target auth state: UNCHALLENGED
09:26:00.028 [DEBUG] [org.apache.http.impl.execchain.MainClientExec] Proxy auth state: UNCHALLENGED
09:26:00.029 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> POST /build?t=igenox/docker-poc&rm=true HTTP/1.1
09:26:00.029 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> Accept: text/plain
09:26:00.030 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> Content-Type: application/tar
09:26:00.030 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> User-Agent: Jersey/2.11 (Apache HttpClient 4.3.1)
09:26:00.030 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> Transfer-Encoding: chunked
09:26:00.030 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> Host: 192.168.59.103:2376
09:26:00.030 [DEBUG] [org.apache.http.headers] http-outgoing-0 >> Connection: Keep-Alive
[...]

PR #145, opened 2016-02-01
Rebase #PR-141 / Issue #123 - Support for docker step reports for DockerBuildImage

rfhayashi:
Rebased #141 against master.

I also changed `createBuildImageResultCallback` to receive a logger parameter, instead of having an extra method to create the proxy callback.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -1,0 +1,54 @@
+package com.bmuschko.gradle.docker.tasks.image
+
+import com.bmuschko.gradle.docker.AbstractFunctionalTest
+import com.bmuschko.gradle.docker.TestPrecondition
+import org.gradle.testkit.runner.BuildResult
+import spock.lang.Requires
+
+@Requires({ TestPrecondition.DOCKER_SERVER_INFO_URL_REACHABLE })
+class DockerBuildImageFunctionalTest extends AbstractFunctionalTest {
+
+    def "Can build image"() {
+        buildFile << """
+import com.bmuschko.gradle.docker.tasks.image.Dockerfile
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+
+task dockerFile(type: Dockerfile) {
+    from 'ubuntu:12.04'
+}
+
+task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
+    inputDir = file("build/docker")
+}
+"""
+        when:
+        BuildResult result = build('buildImage')
+
+        then:
+        !result.standardOutput.contains('Step 1 : FROM ubuntu:12.04')
+        result.standardOutput.contains('Created image with ID')
+    }
+
+    def "Can build image and print stream"() {
+        buildFile << """
+import com.bmuschko.gradle.docker.tasks.image.Dockerfile
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+
+task dockerFile(type: Dockerfile) {
+    from 'ubuntu:12.04'
+}
+
+task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
+    inputDir = file("build/docker")
+}
+"""
+        when:
+        BuildResult result = build('buildImage', '-i')
+        println result.standardOutput
+
+        then:
+        result.standardOutput.contains('Step 1 : FROM ubuntu:12.04')
+        result.standardOutput.contains('Created image with ID')
+    }
+
+}
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
@@ -54,6 +54,7 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
             config.defaultDependencies { dependencies ->
                 dependencies.add(project.dependencies.create("com.github.docker-java:docker-java:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION"))
                 dependencies.add(project.dependencies.create('org.slf4j:slf4j-simple:1.7.5'))
+                dependencies.add(project.dependencies.create('cglib:cglib:3.2.0'))
             }
 
             group = DEFAULT_TASK_GROUP
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -113,7 +113,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             buildImageCmd.withBuildAuthConfigs(authConfigurations)
         }
 
-        def response = buildImageCmd.exec(threadContextClassLoader.createBuildImageResultCallback())
+        def response = buildImageCmd.exec(threadContextClassLoader.createBuildImageResultCallback(logger))
         imageId = response.awaitImageId()
         logger.quiet "Created image with ID '$imageId'."
     }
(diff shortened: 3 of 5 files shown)


cdancy:
+1 lgtm. If no objections from @bmuschko I'll merge shortly.

bmuschko:
@cdancy I'd personally would prefer a more generalized solution like the `ResponseHandler`. Hopefully, I get around to implement the actual fix in Gradle core that doesn't require you to do all the classloader isolation/reflection stuff in plugins anymore. I guess for now we can go with what we have.

cdancy:
@rfhayashi thanks for the commit!

@bmuschko is right in that in the long run we really need a proper solution so that a hack, a very clever hack I might add, like this is not necessary. Until then this will suffice.

llamahunter:
Hmm.. unclear exactly how to use this change.  I was expecting there to be an attribute or something on the DockerBuildImage task that I could set to get it to output the build steps being taken.  The only way I could figure out how to get the build steps to print out was to set the gradle log level to 'info' via the '-i' flag, which caused a massive spewage of uninteresting stuff.

Is there a plan to add an attribute to DockerBuildImage to control the output of docker build steps? Ideally, it should better match the output of the command line tool.

llamahunter:
Hmm.. the original code looks more like what I was expecting.  Why was this approach abandoned?

https://github.com/rfhayashi/gradle-docker-plugin/commit/fba32421428a8fd83364263db43848f5ef87099b

llamahunter:
Opened issue #67 to track missed expectations.

cdancy:
@llamahunter I believe this was talked about elsewhere but am coming up empty-handed as to where the conversation took place. If I remember correctly we were going to keep functionality as it is now and let folks trigger the extra verbosity should they want it. Using '-i' is a standard gradle practice to get at that logging.

With that said, I am in agreement with you that this should be turned on by default as it is what folks would expect. @bmuschko any objections to using logger.quite in place of logger.info here to get at the docker step reports?

llamahunter:
I get a lot of pushback from developers on my team because we are using the gradle-docker-plugin rather than doing something like systeming out to the docker command line binary *primarily* because they get no feedback from gradle-docker-plugin about what is going on during the build.  They do *not* want a bunch of extraneous information about other stuff gradle might do before or after the docker image build that comes with the '-i' option. 

As building a docker image has many steps, and can take a long time, they are in the dark right now about which layer is slow to construct, or if the build is proceeding at all.

llamahunter:
@cdancy discussion about *not* using a task attribute was in pull #141

cdancy:
@llamahunter from that link it looks like I WAS in favor of turning it on by default ;) Perhaps it was missed during the rebase? IDK. Yes using '-i' does give a lot of output but from I can tell it's pretty easy to discern the docker output from anything else ... given that nothing else should really be going on anyway. I agree it's not ideal, and unfortunate that we missed this typo, but you can still get at the info should you need it.

llamahunter:
I'm going to add back the attribute on DockerBuildImage.  I'll leave it off by default for the gradle purists, but at least docker people will have something they can turn on to get the expected build output.

llamahunter:
There's already a 'quiet' flag on DockerBuildImage that, if set to true, is being passed to the docker remote API to suppress verbose build output.  However, as the DockerBuildImage task isn't outputting anything anyway, it's sort of moot.  

I suggest we honor the existing 'quiet' attribute, which currently defaults to 'false', i.e. verbose mode, which is what docker users expect.  I am going to use logger.quiet to output, so that if users pass '-q' to gradle, but not to DockerBuildImage, the output will still get suppressed.

So, really, this is going to end up being a 1 line change.

rfhayashi:
Just to confirm, I intentionally removed the task attribute because @bmuschko advised to use the -i switch, so the attribute was unnecessary. As in my tests I was only invoking DockerBuildImage, it didn't make much difference using one way or the other, so I didn't realize the extraneous information problem @llamahunter mentioned. Having said that, I also think that having the step reports printed by default is the ideal.

llamahunter:
Thanks very much @rfhayashi for contributing this change.  IMO, the advice you received to make it part of gradle 'info' logging was not correct.  Turns out, tho, that the 'quiet' flag in the docker remote API doesn't really make it all that quiet.  So, it seems a parameter will need to be added back to DockerBuildImage to output the build steps.  I'm thinking of 'verbose'.  I may also parse/emit the progressDetail messages, and add an extra flag for that, probably 'progress'.

cdancy:
@llamahunter 'verbose' is left over relic from the Ant days. In the gradle word '-i' is what is used. Again, the extraneous information coming out should be nothing for any developer (gradle or otherwise) to discern. Is it a lot, yes, but nothing any engineer worth his/her salt can't easily sift through. And just to re-iterate: if you're building an image with '-i' turned on the docker step reports should really be the only thing you see when this process is invoked. Perhaps you're doing something concurrently in the same gradle script to cause the output to become clobbered?

bmuschko:
> I get a lot of pushback from developers on my team because we are using the gradle-docker-plugin (rather than doing something like systeming out to the docker command line binary) primarily because they get no feedback from gradle-docker-plugin about what is going on during the build.

A benefit of using the daemon is that you don't have to have Docker installed on the same machine that runs your build. Shelling out to Docker directly would assume that you are OK with that. In many enterprise environments, people decouple CI from Docker.

> They do not want a bunch of extraneous information about other stuff gradle might do before or after the docker image build that comes with the '-i' option.

I see this this as deficiency of Gradle itself at the moment. Unfortunately, you cannot say "just render the logging information of this plugin but not from Gradle core and/or other plugin". I'd suggest you bring this up on the Gradle forum.

> As building a docker image has many steps, and can take a long time, they are in the dark right now about which layer is slow to construct, or if the build is proceeding at all.

It's definitely a valid use case.

> Turns out, tho, that the 'quiet' flag in the docker remote API doesn't really make it all that quiet.

[This flag is just passed to Docker Java](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy#L108) and independent of Gradle's logging. I am not quite sure what they do with it.

> So, it seems a parameter will need to be added back to DockerBuildImage to output the build steps. I'm thinking of 'verbose'. I may also parse/emit the progressDetail messages, and add an extra flag for that, probably 'progress'.

I see two options here:

1) It's inconsistent and probably also confusing to other users to only introduce logging parameters from a single task type. We could introduce logging on the _plugin_ and expose the options through the DSL similar to Test logging in Gradle core. The downside would be more complicated code in the plugin.
2) You handle the response like you intent as proposed in #153 or similar. Then you have full control over the response. The tradeoff here is that you'll have to write a little code. IMHO not a big though.

llamahunter:
> A benefit of using the daemon is that you don't have to have Docker installed on the same machine that runs your build. Shelling out to Docker directly would assume that you are OK with running Docker and CI on the same box. In many enterprise environments, people decouple CI from Docker for scaling reasons.

The daemon and the command line tool are two different things (yet implemented in the same executable). Shelling out to docker just runs the cli, which does not mean that the docker daemon is running on the same machine.

>>Turns out, tho, that the 'quiet' flag in the docker remote API doesn't really make it all that quiet.

>This flag is just passed to Docker Java and independent of Gradle's logging. I am not quite sure what they do with it.

There have been issues opened with docker to make 'quiet' actually quiet.  However, they are reluctant to make the change at this point due to fears that it may be a breaking change for scripts that are using docker.  See https://github.com/docker/docker/issues/9404

> 1) It's inconsistent and probably also confusing to other users to only introduce logging parameters from a single task type. We could introduce logging on the plugin level and expose the options through the DSL similar to Test logging in Gradle core. The downside would be more complicated code in the plugin.

Yeah, there are other things that the gradle-docker-plugin should be outputting, too, such as the progress of DockerPushImage and DockerPullIImage.

>2) You handle the response like you intent as proposed in #153 or similar. Then you have full control over the response. The tradeoff here is that you'll have to write a little code. IMHO not a big though.

I *do* already have a thin wrapper around gradle-docker-plugin that does a couple of auto-configuration things to meet our build environment.  I will look at #153 and see what I can do with that approach.  However, it seems like something that *most* docker users would want, and to require consumers of gradle-docker-plugin to implement nearly identical code on their own seems less-than-ideal.  Perhaps there could be a utility class in the gradle-docker-plugin that people could attach as a ResponseHandler to get nicely formatted verbose docker output.

PR #147, opened 2016-02-05
ADDED: test for exposing ports through DockerCreateContainer.

cdancy:
REFACTOR: DockerInspectContainer will now return 'ExposedPorts' information.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -343,6 +343,42 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
         result.standardOutput.contains("UP-TO-DATE")
     }
 
+    def "Can build an image, create a container and expose a port"() {
+
+        String uniqueContainerName = createUniqueContainerName()
+
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+            import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+            import com.bmuschko.gradle.docker.tasks.container.DockerInspectContainer
+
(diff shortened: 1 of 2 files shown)


cdancy:
@bmuschko the below gets us what we want. You were close with the above just not close enough ;) It's not ideal but removes the uglier code and gets us the test we need. Upon thinking about it I'm not sure if we want docker-java to fix the issue. My reasoning is that they are just following what docker API is returning. It's docker that is inconsistent. I'm not sure if we want them to return something that does not mirror the API. I actually think we want them to return the null as that's what users would expect. Thoughts?

```
        String exposedPorts = container.config?.@exposedPorts ? container.config.exposedPorts : '[]'
        logger.quiet "ExposedPorts : $exposedPorts"
```

bmuschko:
What I'd expect from Docker Java would be this code:

    public ExposedPort[] getExposedPorts() {
        return exposedPorts != null ? exposedPorts.getExposedPorts() : null;
    }

I'd don't think that's necessarily dependent on whether Docker returns null or not. I am not super-excited about accessing a private field as it may change at any time. On the other hand, Docker Java changes their public API with pretty every release anyway. ;-)

cdancy:
@bmuschko I agree. I can throw together a PR for those guys to see what they think. In the meantime shall we keep the above check on our end if only to get a test in? Should they accept the PR our code wont break anyway we'll just have to refactor at some later date. I think the immediate trade off of accessing a private field, knowing there is no other way and hoping we can get a change in, is acceptable to not having a test at all.

bmuschko:
A PR for Docker Java sounds useful. We should probably also provide null checks for other fields they expose the same behavior. For now we can go with the fix you proposed.

cdancy:
Created PR here: https://github.com/docker-java/docker-java/pull/448

cdancy:
Squashed and merged. Thanks for looking over things @bmuschko.

bmuschko:
@cdancy No problem. Thanks for pressing on with the PRs.

PR #153, opened 2016-02-10
Demo of execution handler using closures

cdancy:
A demo for using closures as an exposable "execution handler" for downstream users. The idea is that users, by far and large, will not be using these but instead using the OOTB version. Should they want to use their own closure they will be responsible for implementing the contract already in place by the given task. A couple things to note:
1. We do not set the executionHandler closure by default thereby allowing users to supply their own.
2. All applicable tasks will now have a corresponding 'configureExecutionHandler()' method that will check if the executionHandler has been set by the user and if not, set our own relative to what the given task requires. Follows the semantics already set by the plugin.

Downstream users could then do something like the snippet below. This will allow the DockerCreateContainer task to succeed should a previous container with that name already exist. Users could then implement something similar for the DockerRemoveContainer task so they don't fail should a container not exist but instead just output that the container is not there or just the exception message like I've done below.

@bmuschko thoughts? I've obviously only done this in a handful of places so as to just get the point across. 

```
task createContainer(type: com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer) {
    targetImageId { "hello-world:latest" }
    containerName = "cdancy-test"
    executionHandler { command ->
        try {
            def container = command.exec()
            logger.quiet "OVERRIDE: Created container with ID '$container.id'."
            container.id
        } catch (Exception e) {
            logger.quiet("Found an exception: " + e.message)
        }
    }
}
```

EDIT: changed name of responseHandler to executionHandler

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -41,11 +41,11 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     File certPath
 
     /**
-     * The default response handler
+     * The default execution handler
      */
     @Input
     @Optional
-    Closure responseHandler
+    Closure executionHandler
 
     ThreadContextClassLoader threadContextClassLoader
 
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -41,11 +41,11 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     File certPath
 
     /**
-     * The default execution handler
+     * Response handler is given 2 objects: response and exception
      */
     @Input
     @Optional
-    Closure executionHandler
+    Closure responseHandler
 
     ThreadContextClassLoader threadContextClassLoader
 
@@ -68,4 +68,20 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     }
 
     abstract void runRemoteCommand(dockerClient)
+
+    public void executeCommand(def command) {
+        def possibleResponse
+        Exception possibleException
+        try {
+            possibleResponse = command.exec()
+        } catch (Exception e) {
+            possibleException = e
+        }
+
+        if(getResponseHandler()) {
+            getResponseHandler()(possibleResponse, possibleException)
+        } else {
+            if (possibleException) throw possibleException
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


cdancy:
@bmuschko if it's not already evident this is a replace for the ResponseHandler route we were previously going. I think this is more flexible and allows users to both control the execution and response. While generally that is not a good idea I think only a small number of "advanced" users will be using this functionality in the first place. And if that's the case, and should they decide to take this route, they need to understand the consequences of doing so and as such must adhere to the contract set by the task itself or risk failures/exceptions.

We've had numerous requests, myself included, to be able to capture the potential exceptions thrown by invoking the relevant command (so as not fail tasks/etc) and also to get at the greater response object. This solves both of those things. The only other option, with respect to this context, would be to do a try/catch in our code and if the user registers a resopnseHandler then hand them back the response AND the exception otherwise re-throw exception on our side like we would normally do. This may be a better route as it removes the need to have the user invoke the exec him/herself.

cdancy:
Don't mean to spam but the more I think about this the more I like the latter proposal. It fits in nicely with the gradle model and allows users the control the need. Let me know as I'd like to push on this work as we have need for it here and could remove our requirement on having a downstream "NEXT" version of this plugin that we host.

bmuschko:
Some good ideas here. I think I'd want to go with option 1 as it gives the most flexibility. Instead of getters and setters for `executionHandler` I'd expose a method with the following signature:

    void executionHandler(Closure c)

Maybe you can get this started as PR and we can review from there. I'd have to play around with the API myself.

cdancy:
I'm actually going to push back on solution #1 as I don't think we want to force the users to make the exec() call on the command object and thus have to muddle gradle tasks with try/catch blocks in the passed closure. Users are really only interested in the response at large and whether or not the exec threw an exception. Solution #2, of which I've just sent in a refactored version, gives users exactly what they need and no more. I don't think we want to force them to follow the greater contract set by the tasks themselves but instead keep things stupid simple. Take a look at the last commit I just sent in to get an idea of what I'm thinking and let me know your thoughts.

llamahunter:
I think the gradle plugin should do the exec, but should provide hooks for getting results (such as for printing progress messages from DockerBuildImage, DockerPullImage, DockerPushImage).

orzeh:
@cdancy any progress on this? It seems to be neat feature.

cdancy:
@orzeh been meaning to pick this back up at some point just haven't had time. Implementing this would eliminate the need for us to build a high-level plugin which uses this as its base.

orzeh:
@cdancy ok I understand. I can help it implementing it if you like.

cdancy:
Closing in favor of #302

QIvan:
For the future people who look for an ability to ignore exception in a Gradle task.
My understanding there is no simple way to do so in Gradle itself, but thanks for #302 you can do something like this:
```groovy
task removeImage(type: DockerRemoveImage) {
    targetImageId('your-image-name-here:latest')
    onError { logger.warn(it.message) }
}
``` 
if there is no such image the task run will just print out a message and continue execution  
```
> Task removeImage
Removing image with ID 'your-image-name-here:latest'.
Status 404: {"message":"No such image: your-image-name-here:latest"}

```

Issue #156, opened 2016-02-14
DockerCreateContainer: expose method to make working with LogConfig easier


cdancy:
Much like we do with exposePorts create a method, aptly named logConfig, to easily create a LogConfig instance for DockerCreateContainer to use.

cdancy:
Closing due to #157

PR #157, opened 2016-02-14
ADDED: expose method 'logConfig'

cdancy:
Addresses issue #156

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
@@ -16,5 +16,6 @@ class InspectContainerResponseHandler implements ResponseHandler<String, Object>
         logger.quiet "VolumesFrom : $container.hostConfig.volumesFrom"
         String exposedPorts = container.config?.@exposedPorts ? container.config.exposedPorts : '[]'
         logger.quiet "ExposedPorts : $exposedPorts"
+        logger.quiet "LogConfig : $container.hostConfig.logConfig.type.type"
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -159,6 +159,10 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         imageId
     }
 
+    void logConfig(String type, Map<String, String> config) {
+        logConfig = new LogConfig(type: type, config: config)
+    }
+
     void exposePorts(String internetProtocol, List<Integer> ports) {
         exposedPorts << new ExposedPort(internetProtocol, ports)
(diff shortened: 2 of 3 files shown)


cdancy:
@bmuschko second look? I was getting a bit tired of specifying the fully qualified name for LogConfig. Not exactly sure why I did not do this in the first place. Perhaps I was still a bit wet behind the ears.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -159,6 +159,10 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         imageId
     }
 
+    void logConfig(String type, Map<String, String> config) {
+        logConfig = LogConfig.newInstance(type: type, config: config)
```
bmuschko: No need to use the `newInstance` method as there aren't any constructors defined. You can just use `new LogConfig(type: type, config: config)`.

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
```diff
@@ -16,5 +16,6 @@ class InspectContainerResponseHandler implements ResponseHandler<String, Object>
         logger.quiet "VolumesFrom : $container.hostConfig.volumesFrom"
         String exposedPorts = container.config?.@exposedPorts ? container.config.exposedPorts : '[]'
         logger.quiet "ExposedPorts : $exposedPorts"
+        logger.quiet "LogConfig : $container.hostConfig.logConfig.type.type"
```
bmuschko: Is `type.type` correct? Shouldn't it just be `$container.hostConfig.logConfig.type`?

bmuschko:
@cdancy See my comments.

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
cdancy: The first 'type' returns an Enum while the second 'type' returns the string representation of that given enum.

https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/model/LogConfig.java

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
cdancy: Fixing now.

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
bmuschko: Ah, never mind. I thought it was the `LogConfig` below. Thanks for clarifying.

bmuschko:
@cdancy LGTM

cdancy:
@bmuschko thanks. Merging now.

Issue #148, opened 2016-02-07
Create Vagrant box used for functional testing


bmuschko:
To be able to execute the whole suite of functional tests a running instance of Docker needs to up and running. It would be far more convenient to provide a Vagrant box which uses Linux as base image and already has Docker installed and pre-configured. The Docker installation should also pre-configure a private registry to test relevant use cases. Given this setup, the [Gradle Vagrant plugin](https://github.com/bmuschko/gradle-vagrant-plugin) could be used to start up/tear down the Vagrant box as part of the functional test fixture from the build. Instead of using Vagrant, [Otto](https://www.hashicorp.com/blog/otto.html) could be explored as an option here as well. An initial Vagrant box definition can be found in PR #70.

This setup would also allow other contributors to execute functional tests before submitting their PR.

bmuschko:
Alright, we got a Vagrant box with Docker installed and properly configured to run all functional tests. No more excuses for contributors. :-)

cdancy:
;)

cdancy:
@bmuschko both tests within [AbstractDockerRemoteApiTaskFunctionalTest.groovy](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractDockerRemoteApiTaskFunctionalTest.groovy) fail for me as it is trying to assert that dockerClientConfig.username, password are my own but with vagrant are set to 'dockeruser', 'ilovedocker'. Not sure how we can programmatically, at least in this context, check if we are executing with vagrant or not. Thoughts?

bmuschko:
@cdancy Are you sure these are tests that are failing for you? The tests don't even make a call to Docker. They just configure the Docker client but never call out.

cdancy:
@bmuschko I'm sure. They don't fail under normal circumstances but for whatever reason under the vagrant umbrella they do.

bmuschko:
@cdancy That's weird. I ran them against the Vagrant box. What's the stack trace?

cdancy:
Expected no exception to be thrown, but got 'org.gradle.testkit.runner.UnexpectedBuildFailure'
	at spock.lang.Specification.noExceptionThrown(Specification.java:118)
	at com.bmuschko.gradle.docker.AbstractDockerRemoteApiTaskFunctionalTest.Can create and execute custom remote API task with extension values(AbstractDockerRemoteApiTaskFunctionalTest.groovy:72)
[...]

PR #158, opened 2016-02-16
Upgrade docker-java to 2.2.0

cdancy:


bmuschko:
Make sure you run through the whole suite of tests before merging. SnapCI does not run the functional tests so you'll have to do that locally. Again shows the importance of #148.

cdancy:
@bmuschko was just going to ping you about that. Let me poke around a bit more to confirm.

bmuschko:
@cdancy You can run them locally but at the moment only if Docker is configured to use plain HTTP.

cdancy:
@bmuschko I must be doing something wrong. Using my local docker instance and registry ALL tests inside `DockerWorkflowFunctionalTest` fail though they fail whether or not I change the docker-java version. Everything else succeeds.

It should be noted that docker-java 2.2.0 is a bug fix release before they release 3.0.0 which I imagine will break things substantially for us.

bmuschko:
@cdancy I suggest that we get the testing environment in place before working on any more "features". It's super important that everyone (also contributors) can execute the tests independent on their local setup.

cdancy:
Getting #148 going would be nice though not sure that should be a requirement to run functional tests? 

I'll look deeper into the failures later today. In theory one should be able to provide a docker url, registry url, and be good to go.

bmuschko:
@cdancy A running Docker instance is really the fixture for being able run the functional tests. That should be automated and standardized (e.g. establish the Docker version for testing, set up private repos etc.). Then you as developer won't need anything. The build will simply bring up the Vagrant box with Docker and stops it after running the functional tests.

Not sure why it fails for you. Might be a different Docker version or configuration I have. That's why it's so important.

cdancy:
@bmuschko nonono I completely agree I just meant that it should not be a hard requirement. There may be reasons why one does not want to download virtualbox as part of running functional tests. Perhaps these tests themselves, much like we do here, are running inside a container themselves and have limited space/etc.

I do see this error over and over again in my test report. From what I can reason it appears that we are requiring '/home/docker/overlay/*' to be a directory we need to have present and writable?

```
Reason:
Could not build image: mkdir /home/docker/overlay/57f136bbeed3ff5b2f4c15fbe9c8c2875e24a815bb0d69ab125262ec91aec136-init/merged/dev/shm: invalid argument
-----
	at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:94)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:91)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner.run(DefaultGradleRunner.java:161)
	at org.gradle.testkit.runner.internal.DefaultGradleRunner.build(DefaultGradleRunner.java:91)
	at com.bmuschko.gradle.docker.AbstractFunctionalTest.build(AbstractFunctionalTest.groovy:99)
	at com.bmuschko.gradle.docker.DockerWorkflowFunctionalTest.Can build an image and push to private registry(DockerWorkflowFunctionalTest.groovy:259)
```

bmuschko:
@cdancy Never ran into that issue. Are you using the checked in Wrapper version? I will try sometime tonight. What version of Docker are you using?

cdancy:
@bmuschko indeed using the wrapper. Here is my local setup:

Host Operating System: OSX
Docker version: 1.8.2 (running on a remote ubuntu box here in the office)
JDK: tried versions 1.7 and 1.8

cdancy:
@bmuschko Ok I've got my head around things and fixed some regressions that popped. I've sent in a commit with changes made thus far. Working on one more now. Looks like we may need to refactor the way `DockerBuildImage.UP-TO-DATE` check works. Think I have a better more simpler solution and one that does not depend on the task name. Stay tuned...

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectContainerResponseHandler.groovy
```diff
@@ -14,7 +14,7 @@ class InspectContainerResponseHandler implements ResponseHandler<String, Object>
         logger.quiet "Links       : $container.hostConfig.links"
         logger.quiet "Volumes     : $container.volumes"
         logger.quiet "VolumesFrom : $container.hostConfig.volumesFrom"
-        String exposedPorts = container.config?.@exposedPorts ? container.config.exposedPorts : '[]'
+        String exposedPorts = container.config.exposedPorts ? container.config.exposedPorts.toString() : '[]'
```
cdancy: This fix we talked about a couple weeks ago was merged. So a win for simplicity here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -71,9 +71,12 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
     DockerBuildImage() {
         ext.getImageId = { imageId }
-
+        onlyIf {
+            boolean runTask = previouslyBuiltImageExists() ? false : true
+            if (!runTask) { logger.quiet("UP-TO-DATE") }
+            runTask
+        }
```
cdancy: So the 'outputs.upToDateWhen { ... } does not work as expected for custom tasks. Searching around on the internets reveals a few reasons why. Others on the gradle forums suggest the above to get around this limitation. It feels hackish to me but really I don't see any way around it. @bmuschko thoughts?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -124,13 +127,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
...
     void saveImageId() {
-        TaskStateHelper state = new TaskStateHelper(name, project.buildDir)
+        TaskStateHelper state = new TaskStateHelper(this.class.simpleName, project.buildDir)
         state.put("imageId", getImageId())
     }
 
     boolean previouslyBuiltImageExists() {
         boolean imageExists = false
-        String prevImageId = new TaskStateHelper(name, project.buildDir).get("imageId")
+        String prevImageId = new TaskStateHelper(this.class.simpleName, project.buildDir).get("imageId")
```
cdancy: TaskStateHelper had to be changed to work off of the class name itself, which doesn't change, rather than the potentially changing name of the user defined task (i.e. task helloWorld(type: DockerBuildImage), task helloWorld2(type: DockerBuildImage)).

cdancy:
@bmuschko take a look over things and tell me what you think. The major changes are:

1. Step reports now start at 0 instead of 1
2. TaskStateHelper now is based off of class simple name instead of user defined task name which may/may-not change.
3. A few tests in the `DockerWorkFlowFunctionalTest` had to change to use the project dir rather than rely on the junit `TemporaryFolder` as it was at times not available. Not exactly sure why.
4. `DockerBuildImage.UP-TO-DATE` does not work as expected and had to be replaced with `onlyIf`. It's not great but gets the job done.

A couple of these could have been caught had I known that snap-ci was not running functional tests. Again, another reason why having vagrant might be a good idea but at least I'll know for the future. In either event things pass all the way through now. Have a look with your setup and tell me what you think.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -314,8 +317,8 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
     }
 
     def "Can build an image only once"() {
-        File imageDir = temporaryFolder.newFolder('images', 'minimal')
-        createDockerfile(imageDir)
+        File imageDir = new File(projectDir, 'images/minimal')
```
bmuschko: Why couldn't we use `temporaryFolder` here anymore?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -25,7 +25,7 @@ task buildImage(type: DockerBuildImage, dependsOn: dockerFile) {
         BuildResult result = build('buildImage')
 
         then:
-        !result.standardOutput.contains('Step 1 : FROM ubuntu:12.04')
+        !result.standardOutput.contains('Step 0 : FROM ubuntu:12.04')
```
bmuschko: Is this due to change in the Docker version or Docker Java?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -71,9 +71,12 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
     DockerBuildImage() {
         ext.getImageId = { imageId }
-
+        onlyIf {
+            boolean runTask = previouslyBuiltImageExists() ? false : true
+            if (!runTask) { logger.quiet("${this.name}: UP-TO-DATE") }
```
bmuschko: This looks like a hack to me. Why can't we use the regular UP-TO-DATE check anymore?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: Don't we want to have a state per instance of the task type? I'd think so.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
cdancy: As soon as a test would fail the temporary folder would get blown away and all downstream tests, or those tests which have yet to run, would fail because the folder no longer exists. Reading up on TemporaryFolder this is expected behavior.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
cdancy: Docker version. docker-java is only reporting on what docker is outputing. I was testing on 1.8.3 which is supported. I think we want to at least support that as docker-java does.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
cdancy: @bmuschko it's a hack no doubt about it. Trying to use the outputs.upToDateWhen {} does not actually work inside of a custom task. Searching the internets I found others with the same problem and it's supposedly a known gradle issue. The workaround suggested was to use onlyIf {} as that works as expected.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
cdancy: @bmuschko the imageId is unique which is what I think we really care about. I don't think we care so much about a given user defined task name so much as what its inputs/outputs are. With this new code the user can define any task name they like but it's the inputs, which are created outside of the build by docker if you get what I mean, which need to be unique.

cdancy:
@bmuschko I'd like to get this PR vetted and merged and then tee up another release (be it minor or point). We've already got a few things in and I don't think we want to make the next release any bigger.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
bmuschko: I still don't get this. Every test cases creates a new test folder via `temporaryFolder.newFolder` so it shouldn't matter whether a previous test failed or not. Am I missing something here?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: @cdancy Can you please point me to a relevant issue report?

bmuschko: @cdancy I added functional test that verifies the correct UP-TO-DATE behavior.

bmuschko:
I merged this but there are plenty of open questions from my end. I am going to look at it tomorrow.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -121,13 +122,15 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
                 targetContainerId { startContainer.getContainerId() }
             }
 
-            task killContainer(type: DockerKillContainer) {
+            task removeContainer(type: DockerRemoveContainer) {
```
bmuschko: Why did we go for `DockerRemoveContainer` here instead?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: @cdancy Got the limitations by adding test cases. `upToDateWhen` only works if you have single task that stores the task execution history. However, it wouldn't work if you have two tasks of type `DockerBuildImage`. In those cases `upToDate` would still check the history of the task. That's why `onlyIf` is the better choice here.

PR #159, opened 2016-02-17
Docker client is configured with specific API version

dainiusvaznys:
Docker client is configured with specific API version -- required for registry authentication to know which format to use for X-Registry-Config header

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -23,14 +23,14 @@ import java.lang.reflect.Array
 import java.lang.reflect.Constructor
 import java.lang.reflect.Method
 
-class DefaultDockerClientSite implements DockerClientSite {
+class DockerThreadContextClassLoader implements ThreadContextClassLoader {
     public static final String MODEL_PACKAGE = 'com.github.dockerjava.api.model'
     public static final String COMMAND_PACKAGE = 'com.github.dockerjava.core.command'
 
     private final Map dockerClientConfiguration
     private def dockerClient
 
-    DefaultDockerClientSite(Map dockerClientConfiguration) {
+    DockerThreadContextClassLoader(Map dockerClientConfiguration) {
         this.dockerClientConfiguration = dockerClientConfiguration
     }
 
@@ -38,7 +38,7 @@ class DefaultDockerClientSite implements DockerClientSite {
      * {@inheritDoc}
      */
     @Override
-    def withDockerClient(Closure closure) {
+    void withClasspath(Closure closure) {
         ClassLoader originalClassLoader = getClass().classLoader
         try {
             if (dockerClient) {
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -27,10 +27,10 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
     public static final String MODEL_PACKAGE = 'com.github.dockerjava.api.model'
     public static final String COMMAND_PACKAGE = 'com.github.dockerjava.core.command'
 
-    private final Map<String, String> dockerClientConfiguration
+    private final Map dockerClientConfiguration
     private def dockerClient
 
-    DockerThreadContextClassLoader(Map<String, String> dockerClientConfiguration) {
+    DockerThreadContextClassLoader(Map dockerClientConfiguration) {
         this.dockerClientConfiguration = dockerClientConfiguration
     }
 
@@ -38,48 +38,26 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
      * {@inheritDoc}
      */
     @Override
-    void withClasspath(Set<File> classpathFiles, Closure closure) {
-        ClassLoader originalClassLoader = getClass().classLoader
+    void withClasspath(Closure closure) {
+        if (!dockerClient) {
+            dockerClient = new DockerClientFactory().getInstance(dockerClientConfiguration)
+        }
 
-        try {
-            Thread.currentThread().contextClassLoader = createClassLoader(classpathFiles)
-            closure.resolveStrategy = Closure.DELEGATE_FIRST
-            closure.delegate = this
+        closure.resolveStrategy = Closure.DELEGATE_FIRST
+        closure.delegate = this
+        closure(dockerClient)
+    }
 
-            if (!dockerClient) {
-                dockerClient = new DockerClientFactory().getInstance(dockerClientConfiguration)
+    private class DockerClientFactory {
+        private Class configClass
+        private Class builderClass
+
+        DockerClientFactory() {
+            classLoader.with {
+                configClass = loadClass('com.github.dockerjava.core.DockerClientConfig')
+                builderClass = loadClass('com.github.dockerjava.core.DockerClientBuilder')
             }
-
-            closure(dockerClient)
         }
-        finally {
-            Thread.currentThread().contextClassLoader = originalClassLoader
-        }
-    }
-
-    /**
-     * Creates the classloader with the given classpath files.
-     *
-     * @param classpathFiles Classpath files
-     * @return URL classloader
-     */
-    private URLClassLoader createClassLoader(Set<File> classpathFiles) {
-        new URLClassLoader(toURLArray(classpathFiles), ClassLoader.systemClassLoader.parent)
-    }
-
-    /**
-     * Creates URL array from a set of files.
-     *
-     * @param files Files
-     * @return URL array
-     */
-    private URL[] toURLArray(Set<File> files) {
-        files.collect { file -> file.toURI().toURL() } as URL[]
-    }
-
-    private static class DockerClientFactory {
-        private final Class configClass = DockerThreadContextClassLoader.loadClass('com.github.dockerjava.core.DockerClientConfig')
-        private final Class builderClass = DockerThreadContextClassLoader.loadClass('com.github.dockerjava.core.DockerClientBuilder')
 
         def getInstance(Map<String, String> configuration) {
             def configBuilder = configClass
@@ -103,12 +81,17 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
 
         private String getApiVersion(def configuration) {
             def client = createClient(configuration)
-            return client.versionCmd().exec().apiVersion
+            client.versionCmd().exec().apiVersion
         }
     }
 
-    static Class loadClass(String className) {
-        Thread.currentThread().contextClassLoader.loadClass(className)
+    private ClassLoader getClassLoader() {
+        URL[] urls = dockerClientConfiguration.classpath*.toURI()*.toURL()
+        dockerClient ? dockerClient.class.classLoader : new URLClassLoader(urls, ClassLoader.systemClassLoader.parent)
+    }
+
+    private Class loadClass(String className) {
+        classLoader.loadClass(className)
     }
 
     /**
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -72,24 +72,30 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
      * @param dockerClientConfiguration Docker client configuration
      * @return DockerClient instance
      */
-    private getDockerClient(DockerClientConfiguration dockerClientConfiguration) {
-        // Create configuration
-        Class dockerClientConfigClass = loadClass('com.github.dockerjava.core.DockerClientConfig')
-        Method dockerClientConfigMethod = dockerClientConfigClass.getMethod('createDefaultConfigBuilder')
-        def dockerClientConfigBuilder = dockerClientConfigMethod.invoke(null)
-        dockerClientConfigBuilder.withUri(dockerClientConfiguration.url)
+    private def getDockerClient(DockerClientConfiguration clientConfiguration) {
+        Class clientConfigClass = loadClass('com.github.dockerjava.core.DockerClientConfig')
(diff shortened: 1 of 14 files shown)


dainiusvaznys:
BTW I have functional test failures even when building the master branch, without my changes. So can't check this CI failure.

cdancy:
@dainiusvaznys were you able to run any of these tests once the change was made?

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -72,24 +72,30 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
-    private getDockerClient(DockerClientConfiguration dockerClientConfiguration) {
-        // Create configuration
-        Class dockerClientConfigClass = loadClass('com.github.dockerjava.core.DockerClientConfig')
-        Method dockerClientConfigMethod = dockerClientConfigClass.getMethod('createDefaultConfigBuilder')
-        def dockerClientConfigBuilder = dockerClientConfigMethod.invoke(null)
-        dockerClientConfigBuilder.withUri(dockerClientConfiguration.url)
-
-        if(dockerClientConfiguration.certPath) {
-            dockerClientConfigBuilder.withDockerCertPath(dockerClientConfiguration.certPath.canonicalPath)
+    private def getDockerClient(DockerClientConfiguration clientConfiguration) {
```
cdancy: In following with standard set by plugin no need to return def.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -72,24 +72,30 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
         }
 
-        def dockerClientConfig = dockerClientConfigBuilder.build()
+        def version = client(clientConfigBuilder).versionCmd().exec().apiVersion
+        clientConfigBuilder.withVersion(version)
+
+        return client(clientConfigBuilder)
+    }
+
+    private def client(def configBuilder) {
```
cdancy: Again, no need to return def. We also might want to rename this `createClient`. Looks like there is a bit of code duplication that we may be able to rid ourselves of by passing in parameters to this method.

cdancy:
@dainiusvaznys I like the idea here but on further thought I'm wondering if this is the best route to take. The way things are written now, each task, at the very least, will be making no less than 2 calls over the wire. Meaning: if I have a chain of 10 tasks, all from this plugin (which I personally have), there will be no less than 20 calls made to docker. That seems a bit much especially if you consider multiple work loads all going at the same time.

Perhaps we expose this via the extension object and allow users to pass it in at runtime? Users could call the `DockerVersion` task as say the first task in a chain, get the API version, and the set the value in the extension for all downstream tasks to use. That could work.

Maybe if there was a clean way, that made sense, to cache this information that might also work.

@bmuschko thoughts?

cdancy:
@dainiusvaznys looking over the[ docker-java docs](https://github.com/docker-java/docker-java#configuration) it looks like you can set the API version in your `~/.docker-java.properties` file. This I believe should get you what you want w/out the overhead I discussed above. 

IMO I think it still worth it to be able to pass the API version via extension/task. Perhaps modify this PR to do that?

dainiusvaznys:
I believe it should cause no additional inconvenience to plugin users in terms of maintaining `~/.docker-java.properties` or explicitly assigning docker remote API version. On the other hand recreating docker client each time within the class loader doesn't seem to make much sense unless I'm missing something. So this leads to docker client being cached in the class loader (API version being checked once per project) while also offering possibility to specify the version in the client code (which is also required for the integration/functional tests).
Please review.

cdancy:
@dainiusvaznys I like where you went with this. Were you able to run any of the functional tests against docker locally? Yes there may be a few failures, possibly related to not specifying a local registry, but most should work.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -66,37 +77,37 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
-        // Create configuration
-        Class dockerClientConfigClass = loadClass('com.github.dockerjava.core.DockerClientConfig')
-        Method dockerClientConfigMethod = dockerClientConfigClass.getMethod('createDefaultConfigBuilder')
-        def dockerClientConfigBuilder = dockerClientConfigMethod.invoke(null)
-        dockerClientConfigBuilder.withUri(dockerClientConfiguration.url)
-
-        if(dockerClientConfiguration.certPath) {
-            dockerClientConfigBuilder.withDockerCertPath(dockerClientConfiguration.certPath.canonicalPath)
+   private static class DockerClientFactory {
+        private final Class configClass = DockerThreadContextClassLoader.loadClass('com.github.dockerjava.core.DockerClientConfig')
```
cdancy: Indentation is a space off for this method.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -66,37 +77,37 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
+                createClient build()
+            }
         }
 
-        def dockerClientConfig = dockerClientConfigBuilder.build()
+       private def createClient(def configuration) {
+           def clientBuilder = builderClass
+                   .getMethod('getInstance', configClass)
+                   .invoke(null, configuration)
+           return clientBuilder.build()
```
cdancy: no need to use return

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -66,37 +77,37 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
 
-        // Create client
-        Class dockerClientBuilderClass = loadClass('com.github.dockerjava.core.DockerClientBuilder')
-        Method method = dockerClientBuilderClass.getMethod('getInstance', dockerClientConfigClass)
-        def dockerClientBuilder = method.invoke(null, dockerClientConfig)
-        dockerClientBuilder.build()
-    }
+       private String getApiVersion(def configuration) {
+           def client = createClient(configuration)
+           return client.versionCmd().exec().apiVersion
```
cdancy: Same as above: no need to use return

dainiusvaznys:
I was able to run functional tests just like the CI build so no problem here.
The reason for not passing DockerExtension to the DockerThreadContextClassLoader is to avoid one more tangle between packages com.bmuschko.gradle.docker.utils and com.bmuschko.gradle.docker.
Now that I'm looking at it I think it makes sense to remove AbstractDockerRemoteApiTask.classpath too and handle the setup from the plugin itself, i.e. follow the path of .url  and .certPath. What do you think?
I'll fix those styling issues, no problem. BTW on return keyword I tend to agree with http://groovy-lang.org/style-guide.html#_return_keyword_optional and omit it mainly from closures. Your code your call though.

cdancy:
In going with the standard set by the plugin we tend to omit `return` as it's not necessary.

I actually just gave things a go with this branch and ran into some issues. Mainly: any task which uses the method `DockerThreadContextClassLoader.loadClass` (I'm specifically referring to `DockerCreateContainer`) now fails as the class loaded in that context is done with a different ClassLoader and thus fails to coerce the object into something docker-java can receive. I pasted the stacktrace below. Running the functional tests would have shown this.

```
Caused by: org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'com.github.dockerjava.api.model.Volumes@4bff28e7' with class 'com.github.dockerjava.api.model.Volumes' to class 'com.github.dockerjava.api.model.Volumes'
	at com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer$_setContainerCommandConfig_closure2.doCall(DockerCreateContainer.groovy:242)
	at com.bmuschko.gradle.docker.utils.DockerThreadContextClassLoader.withClasspath(DockerThreadContextClassLoader.groovy:54)
	at com.bmuschko.gradle.docker.utils.ThreadContextClassLoader$withClasspath.call(Unknown Source)
	at com.bmuschko.gradle.docker.utils.ThreadContextClassLoader$withClasspath.call(Unknown Source)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask.runInDockerClassPath(AbstractDockerRemoteApiTask.groovy:40)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$runInDockerClassPath.callCurrent(Unknown Source)
	at com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer.setContainerCommandConfig(DockerCreateContainer.groovy:240)
	at com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer.this$5$setContainerCommandConfig(DockerCreateContainer.groovy)
	at com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer.runRemoteCommand(DockerCreateContainer.groovy:147)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$_start_closure1.doCall(AbstractDockerRemoteApiTask.groovy:35)
	at com.bmuschko.gradle.docker.utils.DockerThreadContextClassLoader.withClasspath(DockerThreadContextClassLoader.groovy:54)
	at com.bmuschko.gradle.docker.utils.ThreadContextClassLoader$withClasspath.call(Unknown Source)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask.runInDockerClassPath(AbstractDockerRemoteApiTask.groovy:40)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$runInDockerClassPath.callCurrent(Unknown Source)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$runInDockerClassPath.callCurrent(Unknown Source)
	at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask.start(AbstractDockerRemoteApiTask.groovy:34)
```

cdancy:
@dainiusvaznys assuming you have a docker instance that is reachable via http the below should suffice and let you know if there are big issues or not. Again, there may be some registry related failures but if the majority of the functional tests pass then we should be inching closer.

`./gradlew build -DdockerServerUrl=https://127.0.0.1:2375`

dainiusvaznys:
As it turns out functional tests failures are only visible when the reference docker daemon is running :) that's how I missed them.
There are still a lot of failures among them (docker version inconsistencies, killing container that's already stopped) but the happy path seems to be OK.
Please review.

cdancy:
@dainiusvaznys I know about the killing container issue and it is being addressed elsewhere so feel free to ignore. With that said, and taking a pass over this code, it feels as if we are touching too many files and are going outside of the scope of this PR. While I'm sure there is more than enough clean-up and proper naming we could implement lets try and keep the changes as minimal as possible and only to the matter at hand.

dainiusvaznys:
Can you be more specific? Roll back 'Test readability improvements', and 'Logging fixed', restore 'ThreadContextClassLoader' naming -- would that be what you're looking for?
Everything else I believe is strictly to the issue.

cdancy:
Yes to all of the above. I'd like to go back and build off of the PR https://github.com/dainiusvaznys/gradle-docker-plugin/pull/1 I sent to your branch. This gets us what we want and provides a clean solution to build off of. No name changes are required and keeps things short and sweet. All of your extra changes we can address in another PR.

bmuschko:
Seems like there we a lot of changes for just exposing the `apiVersion`. Sounds reasonable to me. I'd say let's minimize that and only change what relevant for exposing this new functionality. Everything else can follow in separate pull requests.

dainiusvaznys:
There are two problems with https://github.com/dainiusvaznys/gradle-docker-plugin/pull/1:
- In my setup I have end project which uses 'intermediate' plugin which applies 'docker' plugin; the 'intermediate' plugin creates and uses tasks out of 'docker' plugin by compiling statically against it (i.e. no dynamically loaded classes). So there's some indirection as you can see. In such a layout I still get ClassCastExceptions javax/ws/rs/client/ClientBuilder because of different classloaders, an issue that's addressed in my pull request as 'classloader kung-fu'
- ThreadContextClassLoader.withClasspath accepting Set<File> classpathFiles is misleading from design perspective as it is only considered for the first time and then cached as docker client. So it misleads in a way that even if you think you're passing a classpath for a docker job in fact you're not if that's a successive call. Which makes classpath passing a constructor issue, not this particular method's. Again, the issue is addressed in my pull request.

I do acknowledge readability and naming are subjective matters and I 'cleaned down' as requested however this pull request is the cleanest way to implement the feature as I see it. It is also a convenient one without forcing plugin users to go figure the API version themselves.

dainiusvaznys:
So gentlemen are you accepting it as it stands and I go ahead and resolve the merge conflicts or is this too much changes for you liking (in which case you might want to simply expose API version with some 5 LOC changed and leave it to the users)?

cdancy:
@dainiusvaznys I actually think there is an issue you stumbled upon and are trying to fix with the "kung fu" commit. That being: I think classes from the buildscript are bleeding into the plugin. I've run into a similar issue here in that xml-apis-2.0.4, from our own project, were bleeding into this plugin and its transitive dependencies requiring 1.4. I take this as a bug in the plugin as we don't want that sort of thing to happen which is why there is a supposed to be a single classloader. The commit I wanted to baseline on does get us closer to solving this problem, but if I understand your problem correctly, further from solving yours. I'm not sure there is much we can do on that end to be honest as I don't think we want to open the classloader up to the main thread as that defeats the purpose of what the plugin is "supposed to be doing". 

I'll have to think on this some more as you've brought to light multiple issues and are trying to put them into 1 commit which I don't particularly like to do. I'll admit you're right on what you'd like to do but I'd like to break things apart into multiple PR's to make reviewing easier. 

We also have vagrant functionality for testing now which should make running the integration tests a snap to see if anything break.

dainiusvaznys:
Sure take your time. Let me know if you need my assistance.

llamahunter:
Hmm.. I guess I should have looked for existing pull requests first.

https://github.com/bmuschko/gradle-docker-plugin/pull/182

cdancy:
The need for this I think has become largely irrelevant in the past year or so. If folks feel otherwise feel free to re-open for discussion.

PR #163, opened 2016-02-23
Fix for specifying host directory to copy into when resource is a sin…

cdancy:
…gle file

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 ### Version 2.6.7 (TBD)
 
 * Upgrade to Gradle 2.11.
+* Bug fix when copying single file from container and hostPath is set to directory for `DockerCopyFileFromContainer` - [Pull request 163](https://github.com/bmuschko/gradle-docker-plugin/pull/163)
+* Step reports are now printed to stdout by default for `DockerBuildImage` - [Pull request 145](https://github.com/bmuschko/gradle-docker-plugin/pull/145)
 
 ### Version 2.6.6 (February 27, 2016)
 
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -18,6 +18,7 @@ package com.bmuschko.gradle.docker.utils
 import com.bmuschko.gradle.docker.DockerRegistryCredentials
 import com.bmuschko.gradle.docker.tasks.DockerClientConfiguration
 import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+import org.gradle.api.GradleException
 import org.gradle.api.logging.Logger
 
 import java.lang.reflect.Array
@@ -346,7 +347,9 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
 
             invoke: {Object proxy, Method method, Object[] args ->
                 if ("onNext" == method.name) {
-                    logger.quiet(args[0].toString())
+                    def possibleStream = args[0].stream
+                    if (possibleStream)
+                        logger.quiet(possibleStream)
                 }
                 method.invoke(delegate, args)
             }
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -344,12 +344,12 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
         enhancer.setSuperclass(delegate.getClass())
         enhancer.setCallback([
 
-                invoke: {Object proxy, Method method, Object[] args ->
-                    if ("onNext" == method.name) {
-                        logger.info(args[0].stream)
-                    }
-                    method.invoke(delegate, args)
+            invoke: {Object proxy, Method method, Object[] args ->
+                if ("onNext" == method.name) {
+                    logger.quiet(args[0].toString())
                 }
+                method.invoke(delegate, args)
+            }
 
         ].asType(loadClass('net.sf.cglib.proxy.InvocationHandler')))
 
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -15,6 +15,7 @@
  */
 package com.bmuschko.gradle.docker
 
+import org.gradle.api.GradleException
 import org.gradle.testkit.runner.BuildResult
 import spock.lang.Requires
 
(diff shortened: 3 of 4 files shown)


cdancy:
@bmuschko I've attached to this PR the fix addressed in #158 where TemporaryFolder breaks all downstream tests. It's really frustrating attempting to run integration tests when one test fails and then the rest fail due to lack of finding a file. This fix that's and I'd really like to get it in if only for my own sanity. Thoughts?

bmuschko:
@cdancy Let's discuss that in more detail. `TemporaryFolder` creates a new temporary directory for every test case. I don't get why that would break other test cases.

cdancy:
Closing as changes here have been merged elsewhere.

bmuschko:
@cdancy Are you sure the changes to `DockerCopyFileFromContainer` et al. have been merged somewhere else? BTW: I reverted to `TemporaryFolder`.

cdancy:
@bmuschko they're not. I was actually just putting together a separate PR to make this bug fix against the newly released version in hopes of getting it in for the next version.

Where are you using TemporaryFolder? [DockerWorkflowFunctionalTest](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy) appears to have my changes still in. I just ran all tests from latest release against docker 1.8.3 and docker 1.6.2 and all went perfectly.

bmuschko:
@cdancy `temporaryFolder.newFolder` is used across the board in all test cases of `DockerWorkflowFunctionalTest`. You might want to pull from Git.

cdancy:
@bmuschko aaahhh yes I see it now and that I think is the proper solution. Before we were creating 1 instance of the TemporaryFolder and sharing that across the board which I believe was causing the breakage I was seeing. This ensures we get that new temp folder for each test. Very awesome.

Any chance I can whip together a PR for this bug fix and re-release? Or better off waiting for another point release?

bmuschko:
@cdancy `TemporaryFolder` is a single instance. However, the `@Rule` annotation make sure that a new temporary folder is created for each test method. I'd just remove all other changes from the PR and just keep the changes to `DockerCopyFileFromContainer`.

RE releasing. We can really release at any point in time. We don't need to gather x number of changes before we can release. Releasing with a single fix/change is totally fine. BTW: When you log into SnapCI can you modify the build pipeline, trigger operations there?

cdancy:
@bmuschko Ok. Re-opening to get this fix in.

cdancy:
Ok merged against master.

`BTW: When you log into SnapCI can you modify the build pipeline, trigger operations there?`

@bmuschko what are you referring to here?

bmuschko:
@cdancy Talking about this: https://snap-ci.com/bmuschko/gradle-docker-plugin/branch/master

cdancy:
Aaaaaahhhhh gotcha and thanks. @bmuschko what do you think about, as part of this point release, turning the 'logger.info' into 'logger.quiet' which is being discussed in #145? I think it makes sense to have this on by default as that's what users would expect in the "docker build image" context.

bmuschko:
@cdancy I think it's only a matter of time until somebody says that there's too much output from the plugin. :-) IMHO we should have a generic solution. If you want to switch it over to `logger.quiet` I am fine with it though I see that as a temporary change.

cdancy:
Ok. Let me put that in as I think it's valuable, only within the DockerBuildImage task, to have that extra verbosity to see the outputs of the "docker build process".

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -344,12 +344,12 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
         enhancer.setCallback([
 
-                invoke: {Object proxy, Method method, Object[] args ->
-                    if ("onNext" == method.name) {
-                        logger.info(args[0].stream)
-                    }
-                    method.invoke(delegate, args)
+            invoke: {Object proxy, Method method, Object[] args ->
+                if ("onNext" == method.name) {
+                    logger.quiet(args[0].toString())
```
cdancy: Fixed indentation and now using `args[0].toString()` as this is what [BuildImageResultCallback](https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/core/command/BuildImageResultCallback.java#L32) does. Using the old way (e.g. args[0].stream) caused [StreamCorruptedException](http://stackoverflow.com/questions/23262160/java-io-streamcorruptedexception-invalid-stream-header-54657374) in some cases.

rfhayashi: @cdancy, I can't test right now if that was really the case, but I think `toString` returns all the information related to the ResultItem (in a format like "Person[name=John Doe,age=33,smoker=false]"). I opted for using just the `stream` field so it reflected more the way docker output looks. Can you check if that is the case?

Can't think of why you're getting the StreamCorruptedException as `getStream` does nothing more than return the field value. If you can show me how to reproduce it, I can have a look tomorrow.

cdancy: Use my branch noted here and run the functional tests like so: `./gradlew functionalTest -PbootstrapDocker=true` 

Or really just checkout master, run the functional tests, if you want to see things fail ;)

cdancy: @rfhayashi looks like you're right. Let me go back and have a look at things.

`toString(): BuildResponseItem[stream=Step 1 : FROM ubuntu:12.04
,status=<null>,progressDetail=<null>,progress=<null>,id=<null>,from=<null>,time=0,errorDetail=<null>,error=<null>]

stream: Step 1 : FROM ubuntu:12.04`

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -344,12 +345,14 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
-                invoke: {Object proxy, Method method, Object[] args ->
-                    if ("onNext" == method.name) {
-                        logger.info(args[0].stream)
-                    }
-                    method.invoke(delegate, args)
+            invoke: {Object proxy, Method method, Object[] args ->
+                if ("onNext" == method.name) {
+                    def possibleStream = args[0].stream
+                    if (possibleStream)
+                        logger.quiet(possibleStream)
```
cdancy: @rfhayashi issue is due to stream potentially being null. A check for null, before putting to stdout, solves the problem.

RELEASE_NOTES.md
```diff
@@ -1,6 +1,8 @@
 ### Version 2.6.7 (TBD)
 
 * Upgrade to Gradle 2.11.
+* Bug fix when copying single file from container and hostPath is set to directory for `DockerCopyFileFromContainer` - [Pull request 163](https://github.com/bmuschko/gradle-docker-plugin/pull/163)
+* Step reports are now printed to stdout by default for `DockerBuildImage` - [Pull request 145](https://github.com/bmuschko/gradle-docker-plugin/pull/145)
```
bmuschko: If we go for `logger.quiet` then it should be done for all tasks for consistency.

cdancy: Not all tasks output information though. And in this specific case the output is something we have to purposefully cull out. I agree in that we should do it to be consistent but only where it makes sense. I don't see any value say for outputting the logs from DockerStopContainer but do see the value for DockerBuildImage.  The former gives us nothing while the latter tells us what the container is actually building so to speak.

bmuschko: Yeah, I meant the tasks that already use `logger.info`.

cdancy: Looking through the tasks they all seem to be using logger.quiet for bare minimum informational output.

PR #166, opened 2016-02-26
fixed docs on current java-docker version

mxab:


bmuschko:
Thanks for the PR. The plugin currently [uses 2.1.4](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy#L31). What lead you to believe that a different version is used?

mxab:
The PR is just a fix in your docs that currently state "2.2.4"

bmuschko:
Ah, thanks. I misread you change. Got it. We are currently in the process of upgrading the Docker Java version to 2.2.0 so I will take care of it.

mxab:
good to know
I was using the 2.2.1 version already as the 2.1.4 broke the image push due to a too strict jackson mapping rule
```dependencies { dockerJava  "com.github.docker-java:docker-java:2.2.1" }````

PR #172, opened 2016-03-04
REMOVE: DockerBuildImage up-to-date functionality as it has proven to…

cdancy:
…o difficult to tell when exactly an image is the same and we don't need to re-build.


cdancy:
@bmuschko PR for removing UP-TO-DATE check on `DockerBuildImage`. I'll leave this open for a bit incase you want to have a look and/or spot anything. All functional tests pass locally here so we should be good to go.

I'll give releasing a shot this weekend (maybe tonight) to get this fix as well as the others committed today.

bmuschko:
@cdancy LGTM

Issue #175, opened 2016-03-05
Buildscript classpath leaking into custom classloader


cdancy:
Add xml-apis-2.+ to the buldscript classpath and watch the plugin falllover. This was briefly talked about elsewhere (#159 )and we've had to do some gradle dependency substitution here to get around this. I consider this a bug as nothing added to the buildscript classpath should be interfearing with this plugin given its custom Classloader.

[@dainiusvaznys](https://github.com/dainiusvaznys) stumbled upon the issue with the aforementioned PR which reminded me that I was also getting around this "bug" and that it eventually needed to be addressed.

I've actually solved this problem with another plugin I wrote, given the ideas that @dainiusvaznys put forward in #159, that uses the same custom classloader design as this plugin. Change is fairly minimal but I'd like to get this in before we release 2.6.7. I'll look into sending a PR in later today.

cdancy:
Closing as I believe an OOTB solution by gradle may help to alleviate this pain in the not too distant future.

Issue #144, opened 2016-01-31
java.lang.ClassNotFoundException: org.bouncycastle.openssl.PEMParser


vskh:
Hi gents,

I'm getting following exception when trying to use com.bmuschko.docker-java-application.
Nothing special, mostly copy paste from documentation:

```
buildscript {
    repositories {
        jcenter()
    }
    dependencies {
        classpath 'com.bmuschko:gradle-docker-plugin:2.6.5'
    }
}
apply plugin:  'com.bmuschko.docker-java-application'
```

Then running gradle 2.8 like this:
`gradlew dockerBuildImage --stacktrace`
gives following exception:
```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':oauth-authorization-server:dockerBuildImage'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:69)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.PostExecutionAnalysisTaskExecuter.execute(PostExecutionAnalysisTaskExecuter.java:35)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:64)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:58)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:52)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:52)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:53)
        at org.gradle.api.internal.tasks.execution.ExecuteAtMostOnceTaskExecuter.execute(ExecuteAtMostOnceTaskExecuter.java:43)
        at org.gradle.execution.taskgraph.DefaultTaskGraphExecuter$EventFiringTaskWorker.execute(DefaultTaskGraphExecuter.java:203)
[...]

PR #176, opened 2016-03-05
Rework plugin to support single dockerClient with isolated ClassLoader

cdancy:
Addresses #175 . Some notes:

1) Single instance of dockerClient is created and used across the board.
2) Custom ClassLoader impl is simplified immensely.
3) DockerClientConfig is removed as it's no longer needed. And which leads me to...
4) url/certpath/classpath are removed from AbstractDockerTask. The last was required as part of this fix but are the first 2 (url and certpath) really being used on tasks? I think this is something that is common across all tasks and thus should stay in DockerExtension only.
5) All functional tests pass.

@bmuschko thoughts and review?

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -42,13 +42,20 @@ abstract class AbstractFunctionalTest extends Specification {
                 .collect { "'$it'" }
                 .join(", ")
 
-        // Add the logic under test to the test build
+        // Add the logic under test to the test build. 
+        //
+        // We are adding known versions of dependencies that will break the plugin 
+        // should they leak into our custom classloader. Should our custom classloader
+        // be configured correctly adding these to the buildscript classpath should
+        // cause no issue.
         buildFile << """
             buildscript {
                 repositories {
                     mavenCentral()
                 }
                 dependencies {
+                    classpath 'org.bouncycastle:bcpkix-jdk15on:1.47'
+                	classpath 'xml-apis:xml-apis:2.0.2'
                     classpath files($pluginClasspath)
                 }
             }
@@ -56,10 +63,6 @@ abstract class AbstractFunctionalTest extends Specification {
 
         buildFile << """
             apply plugin: com.bmuschko.gradle.docker.DockerRemoteApiPlugin
-
-            repositories {
-                mavenCentral()
-            }
         """
 
         setupDockerServerUrl()
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -45,6 +45,9 @@ abstract class AbstractFunctionalTest extends Specification {
         // Add the logic under test to the test build
         buildFile << """
             buildscript {
+                repositories {
+                    mavenCentral()
+                }
                 dependencies {
                     classpath files($pluginClasspath)
                 }
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
@@ -9,6 +9,10 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
 
     def setup() {
         project = ProjectBuilder.builder().withProjectDir(projectDir).build()
+
+        project.repositories {
+            mavenCentral()
+        }
     }
 
     def "Does not create tasks out-of-the-box when application plugin is not applied"() {
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
@@ -46,26 +46,20 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
     }
 
     private void configureAbstractDockerTask(Project project, DockerExtension extension) {
-        ThreadContextClassLoader dockerClassLoader = new DockerThreadContextClassLoader()
-
+        ThreadContextClassLoader dockerClassLoader = new DockerThreadContextClassLoader(extension, configurePluginClassPath(project))
         project.tasks.withType(AbstractDockerRemoteApiTask) {
-            Configuration config = project.configurations[DOCKER_JAVA_CONFIGURATION_NAME]
-
-            config.defaultDependencies { dependencies ->
-                dependencies.add(project.dependencies.create("com.github.docker-java:docker-java:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION"))
-                dependencies.add(project.dependencies.create('org.slf4j:slf4j-simple:1.7.5'))
-                dependencies.add(project.dependencies.create('cglib:cglib:3.2.0'))
-            }
-
             group = DEFAULT_TASK_GROUP
             threadContextClassLoader = dockerClassLoader
-
-            conventionMapping.with {
-                classpath = { config }
-                url = { extension.url }
-                certPath = { extension.certPath }
-            }
         }
+    }
+
+    private Set<File> configurePluginClassPath(Project project) {
+        project.repositories.addAll(project.buildscript.repositories.collect())
+        project.configurations.getByName(DOCKER_JAVA_CONFIGURATION_NAME).defaultDependencies { dependencies ->
+            dependencies.add(project.dependencies.create("com.github.docker-java:docker-java:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION"))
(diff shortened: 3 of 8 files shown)


cdancy:
@bmuschko any thoughts before I merge this? My only real concern, which may not be a concern at all, is with bullet point 4 above. I've never seen it advertised that one could set the url/certpath on the task itself so I don't think this is a big loss if any. Most folks I would assume are using the extension. Just want to confirm as this change means we are creating a single dockerClient, which is init'ed just once, and so setting any url/certpath further downstream no longer makes sense.

bmuschko:
@cdancy I briefly looked over it but didn't have the time to comment on it. It would be a breaking change so I'd say hold off on it. Can you please also explain what we are trying solve with this change? It's not clear to me. If this has something to do with classloader conflicts then I'd expect to see a test case for it.

Can you try to release without this PR? I think we should get out a new version asap to make sure people are not breaking on the previous issue.

cdancy:
We can release without and save for a future release. 

It's a bug fix, sort of, depending on how things were originally meant to be designed. It may just be an enhancement when looked at from another angle. What it does do is not allow the buildscript classpath to leak into the plugin classpath. Let me put together a test that throws some dependencies on the buildscript classpath (e.g. xml-apis-2.0) and ensure things run as expected.

cdancy:
@bmuschko anything to do before triggering the release phase on snap-ci? Do you have a list of what needs to be done or is literally everything automated?

bmuschko:
@cdancy Basically just have the proper release version checked in `build.gradle`. Release notes are finalized. That's pretty much it. Press the button on the release job. I will have to manually confirm the release on JCenter. Not sure if you could do that too. Will have to add you to it.

cdancy:
I've got an account on [bintray](https://bintray.com/cdancy) if you need to find me. Let me give this a go and see what happens

bmuschko:
@cdancy Looks like you need a commercial account now to add permissions for other people. Maybe there's a way to automate that as well. Not sure. I made the publication public.

cdancy:
Issues like [this one](https://github.com/bmuschko/gradle-docker-plugin/issues/144) will cease to be a problem. As noted above: I hit the same when xml-apis-2-+ is on the buildscript classpath.

codepoetmc:
I suspect, but have not fully confirmed this also may fix an issue that makes 2.6.7 no longer work with gradle 2.5 (minimum gradle version specified on project page). Gradle 2.5 seems to depend on groovy 2.3.10. When using gradle 2.5 and com.bmuschko:gradle-docker-plugin v2.6.7 missing method exception is thrown for org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.castToVargsArray (a groovy feature added in v2.4). Downgrading the com.bmuschko:gradle-docker-plugin dependency to 2.6.5 or upgrading gradle to a version that depends on groovy 2.4 seems to resolve the issue.

bmuschko:
@codepoetmc The requirement on Gradle 2.5 is due to the fact that the plugin uses an API that was introduced with 2.5.

cdancy:
@bmuschko instead of adding tests, for which this change will literally touch everything, I instead added dependencies to the buildscript classpath for functional tests that are known to break the plugin should they exist. As such, all downstream functional tests will now be affected by this change and should they break anything most tests should fail. But of course they don't given this commit.

As part of this change the user no longer needs to specify an external repository block to pick up our dependencies. Instead we will use the repositories defined within the buildscript block to resolve these. This is more/less a side-effect of the changes being proposed in this PR and one that users would expect anyway.

src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
```diff
@@ -9,6 +9,10 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
 
     def setup() {
         project = ProjectBuilder.builder().withProjectDir(projectDir).build()
+
+        project.repositories {
```
bmuschko: The test case doesn't require any external dependencies. No need to define a repository.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -42,10 +42,20 @@ abstract class AbstractFunctionalTest extends Specification {
...
+        // should they leak into our custom classloader. Should our custom classloader
+        // be configured correctly adding these to the buildscript classpath should
+        // cause no issue.
         buildFile << """
             buildscript {
+                repositories {
+                    mavenCentral()
+                }
                 dependencies {
+                    classpath 'org.bouncycastle:bcpkix-jdk15on:1.47'
```
bmuschko: This shouldn't be define on this level of the test. It should be an individual test case just for this purpose.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -42,10 +42,20 @@ abstract class AbstractFunctionalTest extends Specification {
...
-        // Add the logic under test to the test build
+        // Add the logic under test to the test build. 
+        //
+        // We are adding known versions of dependencies that will break the plugin 
+        // should they leak into our custom classloader. Should our custom classloader
+        // be configured correctly adding these to the buildscript classpath should
+        // cause no issue.
         buildFile << """
             buildscript {
+                repositories {
```
bmuschko: The plugin itself doesn't require an external repository on the `buildscript` level.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -29,22 +29,23 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
-        try {
-            Thread.currentThread().contextClassLoader = createClassLoader(classpathFiles)
-            closure.resolveStrategy = Closure.DELEGATE_FIRST
-            closure.delegate = this
-            closure(getDockerClient(dockerClientConfiguration))
-        }
-        finally {
-            Thread.currentThread().contextClassLoader = originalClassLoader
-        }
+    void withClosure(Closure closure) {
```
bmuschko: The method name is not very speaking. I don't know what the closure works on or what it really does without looking at the implementation.

src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
```diff
@@ -46,28 +46,22 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
...
         }
     }
 
+    private Set<File> configurePluginClassPath(Project project) {
+        project.repositories.addAll(project.buildscript.repositories.collect())
+        project.configurations.getByName(DOCKER_JAVA_CONFIGURATION_NAME).defaultDependencies { dependencies ->
+            dependencies.add(project.dependencies.create("com.github.docker-java:docker-java:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION"))
+            dependencies.add(project.dependencies.create('org.slf4j:slf4j-simple:1.7.5'))
+            dependencies.add(project.dependencies.create('cglib:cglib:3.2.0'))
+        }.files
```
bmuschko: The method call `.files` will resolve all dependencies right away during Gradle's configuration time. This should be avoided. Dependencies should only be resolved when they are actually used e.g. running a task.

src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
```diff
@@ -46,28 +46,22 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
...
-            conventionMapping.with {
-                classpath = { config }
-                url = { extension.url }
-                certPath = { extension.certPath }
-            }
         }
     }
 
+    private Set<File> configurePluginClassPath(Project project) {
+        project.repositories.addAll(project.buildscript.repositories.collect())
```
bmuschko: See discussion in other thread. I don't think it's a good idea.

bmuschko:
Can you explain how your changes actually prevent the classloading issue? After briefly looking through the code that's still not clear to me.

Side note: I am currently working on the plugin classpath isolation in Gradle core which would prevent Gradle core runtime libraries leak into the plugin classpath. With that we'd not need a custom classloader anymore and could just use the classes as is. What still might happen though is that other plugins pull in conflicting versions of the libraries provided by Docker Java. I'd hold off on this until we get that functionality out-of-the-box. It's likely going to be available with Gradle 2.14.

bmuschko:
Gradle 2.14 is available as release candidate. I'd suggest you try out that version which avoids leaking internal Gradle dependencies to the plugin classpath. As a result we might be able to go without the custom classloading mechanism. Nevertheless, dependencies from other plugins could still conflict with dependencies of Docker Java.

cdancy:
@bmuschko I'll give things a go. We definitely don't want any clobbering from other plugins as I've hit that issue myself. The docker-java guys were supposed to be putting together a fat/shadow jar. Using that might also save us from the dependency clobbering that can happen at the expense of packing everything into a single binary.

cdancy:
Closing as this has become largely unnecessary IMO.

PR #181, opened 2016-03-17
DockerLogsContainer implementation

### LINKED ISSUE #132: Creation of DockerLogsContainer task
Use of this task, especially within a dev-ops environment, can be critical as one does not generally know when a given container, and its app therein, is "live". Exposing this task is an easy win for those needing to know when a container, and its app, have reached the "live" point via some log message.

In keeping with the spirit of the naming convention in place, it seems this name is as good as any. I can take on the work here.

double16:
This task implements `docker logs`. Fixes #132

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.bmuschko.gradle.docker.tasks.container
+
+import org.gradle.api.InvalidUserDataException
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
+
+/**
+ * Copies the container logs into standard out/err, the same as the `docker logs` command. The container output
+ * to standard out will go to standard out, and standard err to standard err.
+ */
+class DockerLogsContainer extends DockerExistingContainer {
+    /**
+     * Set to true to follow the output, which will cause this task to block until the container exists.
+     * Default is unspecified (docker defaults to false).
+     */
+    @Input
+    @Optional
+    Boolean follow
+
+    /**
+     * Set to true to copy all output since the container has started. For long running containers or containers
+     * with a lot of output this could take a long time. This cannot be set if #tailCount is also set. Setting to false
+     * leaves the decision of how many lines to copy to docker.
+     * Default is unspecified (docker defaults to true).
+     */
+    @Input
+    @Optional
+    Boolean tailAll
+
+    /**
+     * Limit the number of lines of existing output. This cannot be set if #tailAll is also set.
+     * Default is unspecified (docker defaults to all lines).
+     */
+    @Input
+    @Optional
+    Integer tailCount
+
+    /**
+     * Include standard out.
+     * Default is true.
+     */
+    @Input
+    @Optional
+    boolean stdOut = true
+
+
+    /**
+     * Include standard err.
+     * Default is true.
+     */
+    @Input
+    @Optional
+    boolean stdErr = true
+
+    /**
+     * Set to the true to include a timestamp for each line in the output.
+     * Default is unspecified (docker defaults to false).
+     */
+    @Input
+    @Optional
+    Boolean showTimestamps
+
+    /**
+     * Limit the output to lines on or after the specified date.
+     * Default is unspecified (docker defaults to all lines).
+     */
+    @Input
+    @Optional
+    Date since
+
+    @Override
+    void runRemoteCommand(dockerClient) {
+        logger.quiet "Logs for container with ID '${getContainerId()}'."
+        def logCommand = dockerClient.logContainerCmd(getContainerId())
+        setContainerCommandConfig(logCommand)
+        logCommand.exec(threadContextClassLoader.createLoggingCallback(logger))?.awaitCompletion()
+    }
+
+    private void setContainerCommandConfig(logsCommand) {
+        if (getFollow() != null) {
+            logsCommand.withFollowStream(getFollow())
+        }
+
+        if (getShowTimestamps() != null) {
+            logsCommand.withTimestamps(getShowTimestamps())
+        }
+
+        logsCommand.withStdOut(getStdOut()).withStdErr(getStdErr())
+
+        if (getTailAll() != null && getTailCount() != null) {
+            throw new InvalidUserDataException("Conflicting parameters: only one of tailAll and tailCount can be specified")
+        }
+
+        if (getTailAll() != null) {
+            logsCommand.withTailAll()
+        } else if (getTailCount() != null) {
+            logsCommand.withTail(getTailCount())
+        }
+
+        if (getSince()) {
+            logsCommand.withSince((int) (getSince().time / 1000))
+        }
+    }
+}
+
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -339,6 +339,40 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
         createCallback("${COMMAND_PACKAGE}.PullImageResultCallback")
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    def createLoggingCallback(Logger logger) {
+        Class callbackClass = loadClass("${COMMAND_PACKAGE}.LogContainerResultCallback")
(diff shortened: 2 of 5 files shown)


cdancy:
@double16 thanks for the PR! A few things:

1. Any particular reason you created a separate functional test file? We've been putting all functional tests inside `DockerWorkFlowFunctionalTest`.
2. For loading classes from docker-java please follow the model we set within `ThreadContextClassLoader` and it's impl `DockerThreadContextClassLoader`.
3. Consider creating an inner class for the `tail` variables. Having 1 variable mean potentially N number of things will be confusing for users. Take a look at `DockerCreateContainer` to see how we've done similar things in the past.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -0,0 +1,92 @@
...
+    private void setContainerCommandConfig(logsCommand) {
+        if (follow != null) {
+            logsCommand.withFollowStream(follow)
+        }
+
+        if (showTimestamps != null) {
+            logsCommand.withTimestamps(showTimestamps)
+        }
+
+        logsCommand.withStdOut(true).withStdErr(true)
```
cdancy: Should make these variables and default to true. If it's an option then it should be exposed to users to toggle whichever they may need.

double16:
I'll address these and one more thing ... I forgot to add documentation. :blush:

cdancy:
On further thinking: lets keep the functional tests where you have them now. Our all-in-one functional test class is getting quite big and I think making it bigger, if only to keep tests in one place, is not the way to go.

double16:
I've addressed your comments. For the tail, I opted to go with two tail options similar to the docker-java API and throw an exception if both are specified.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        if (getShowTimestamps() != null) {
+            logsCommand.withTimestamps(getShowTimestamps())
+        }
+
+        logsCommand.withStdOut(getStdOut()).withStdErr(getStdErr())
+
+        if (getTailAll() != null && getTailCount() != null) {
+            throw new InvalidUserDataException("Conflicting parameters: only one of tailAll and tailCount can be specified")
```
cdancy: Throw GradleException instead. It's standard to expect as much from a gradle plugin.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -0,0 +1,150 @@
...
+
+    /**
+     * Limit the output to lines on or after the specified date.
+     * Default is unspecified (docker defaults to all lines).
+     */
+    @Input
+    @Optional
+    Date since
+
+    private createLoggingCallback() {
```
cdancy: Stick this method into `DockerThreadContextClassLoader` and add its interface definition to `ThreadContextClassLoader`. Look at any of the other tasks to see how we do this.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
double16: InvalidUserDataException extends GradleException, so being more specific isn't hurting handling the more generic case, right? According to Gradle API, InvalidUserDataException is appropriate here:
```
A InvalidUserDataException is thrown, if a user is providing illegal data for the build.
```

cdancy: Ahhhh gotcha. I assumed it was java exception. Ok lets keep it. The standard we've been keeping thus far is to use the generic GradleException in all places. Having something a bit more specific can't hurt.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+        ].asType(threadContextClassLoader.loadClass('net.sf.cglib.proxy.InvocationHandler')))
+
+        enhancer.create()
+    }
+
+    @Override
+    void runRemoteCommand(dockerClient) {
+        logger.info "Logs for container with ID '${getContainerId()}'."
```
cdancy: Lets use `logger.quiet` here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -0,0 +1,150 @@
...
+        def delegate = callbackClass.newInstance()
+
+        Class enhancerClass = threadContextClassLoader.loadClass('net.sf.cglib.proxy.Enhancer')
+        def enhancer = enhancerClass.getConstructor().newInstance()
+        enhancer.setSuperclass(callbackClass)
+        enhancer.setCallback([
+
+            invoke: {Object proxy, java.lang.reflect.Method method, Object[] args ->
+                if ("onNext" == method.name) {
+                  def frame = args[0]
```
cdancy: Are we sure `frame` wont be null? We had an issue before within [createPrintStreamProxyCallback](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy#L350) where the returned 'stream' was actually null causing failures for users.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -339,6 +339,40 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
+        enhancer.setSuperclass(callbackClass)
+        enhancer.setCallback([
+
+            invoke: {Object proxy, java.lang.reflect.Method method, Object[] args ->
+                if ("onNext" == method.name && args.length && args[0]) {
+                  def frame = args[0]
+                  switch (frame.streamType as String) {
+                    case "STDOUT":
+                    case "RAW":
+                        System.out.print(new String(frame.payload))
```
cdancy: lets change this to logger.quiet

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -339,6 +339,40 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
+                        break
+                    case "STDERR":
+                        System.err.print(new String(frame.payload))
```
cdancy: and this to logger.error

cdancy:
@double16 just a couple minor things related to logging and I think we are set. Have you run the integration tests yet?

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
double16: That means normal output is only shown when Gradle is run with `--info` or `--debug`. From my point of view, if I'm using `DockerLogsContainer` I intend to see container output as a part of the normal Gradle run. You sure you want this to be logger output?

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
double16: Following the logic for standard out, if it continues to go to stdout, for consistency I think this should go to stderr. If standard out goes to the logger, then logger.error makes sense here

double16:
@cdancy I have run all of the tests for the project and they pass, except for three that push to DockerHub and private repos, which I didn't configure.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
cdancy: logger.quiet will be shown by default. logger.info will only be shown if the `-i` flag is given. This is why we have a simple logger.quiet message for each task to give some indication something has happened. Is this not happening in your case?

double16: Sorry, I was confused. logger.quiet is showing up for me by default. I've made the requested logging changes.

cdancy:
@double16 lgtm. Squash and rebase then I'll merge.

double16:
@cdancy squashed and ready to go

cdancy:
Thanks @double16!

cdancy:
@double16 have you used this task at all? I'm more wondering how you are using it assuming you're doing so. My thoughts surround the capturing of the logs themselves and if you're doing so through some mechanism. I'm thinking of putting together some sort of `sink` for this task or some way to capture the output so folks can process it and/or make some decision further downstream. With all that in mind I'm wondering how you are going about things and if you have any thoughts surrounding such a thing.

double16:
I am currently using this task in a Bamboo plan. Bamboo captures the Gradle output and saves the logs so if something goes wrong (or it succeeds), the logs can be inspected. I've really only used the logs to review when something goes wrong.

I avoid making decisions based on log output. I rely on output files with a specific purpose such as JUnit XML, or JSON, etc.

Allowing a sink to be specified sounds like a good idea, as long as it's optional. I'd really like to keep the functionality to output to stdout/stderr so CI tools like Bamboo or Jenkins can capture the output.

cdancy:
Aaahhhh a fellow Bamboo user! Our use case here is more for checking when a given container is "live" by waiting for our app to print out an "init is complete" message.

I'd definitely keep the functionality as-is, as we've already released and folks will have come to depend on it. Any additions would be purely optional at this point.

Thanks for the info.

PR #182, opened 2016-03-18
Extension variable for docker remote api version.

llamahunter:
Provide a way to set the docker remote api version on the underlying docker-java library.
The default version seems to be UNKNOWN_VERSION, which usually is ok, but if you are
talking to a v2 registry, the api version is used as a proxy to determine how to send
authentication headers.  If unset, v2 registry authentication will fail.  See also
https://github.com/docker-java/docker-java/issues/506

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
@@ -21,6 +21,7 @@ class DockerExtension {
     String url = 'http://localhost:2375'
     File certPath
     DockerRegistryCredentials registryCredentials
+    String apiVersion
 
     void registryCredentials(Closure closure) {
         registryCredentials = new DockerRegistryCredentials()
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
@@ -64,6 +64,7 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
                 classpath = { config }
                 url = { extension.url }
                 certPath = { extension.certPath }
+                apiVersion = { extension.apiVersion }
             }
         }
     }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -40,6 +40,13 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     @Optional
     File certPath
 
+    /**
+     * The docker remote api version
+     */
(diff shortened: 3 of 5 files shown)


cdancy:
Once we flesh out some remaining issues (mainly the classpath issue noted in #176 ) we will circle back and pick parts of #159 out. I like the idea he was going with in that we make an initial call to get the API and, cache that, and use for all future downstream tasks within a given script. This removes user intervention from setting it and will ensure we pass along the correct version every time.

cdancy:
Not pulling in this PR. As I noted above: we will circle back and pick the parts out of #159 as that is the direction, WRT setting API version, that we want to go.

llamahunter:
Ok, well, it's sort of a pressing issue.  I have my own build, so I'm ok for now, but I imagine others may have authentication problems using gradle-docker-plugin with v2 registries.

cdancy:
Re-opening as upon further thought, and with a couple potential changes coming soon, this may be something folks want to leverage.

cdancy:
@llamahunter if you provide some functional tests (i.e. showing setting the version works and showing it fails when set to something invalid) we can look to getting it merged before I do another release probably this Sunday.

llamahunter:
Ok... well, there's not really any checking all the way down to docker-java on 'invalid' versions.  It's just treated as a string.  I wrote a little test that checks that the API version set is the API version used, but I can't figure out how to get the functional tests to run against a docker machine that has TLS enabled.

llamahunter:
Hmm.. well, maybe it doesn't really need to talk to the remote server. So, removed that check. Anyway, the test feels a little bit 'fragile', as it has to dig through the internal implementation details of docker-java to verify that the apiVersion was set correctly.  Here's the test case, to see if you *really* want it.  I chose a (currently) nonsensical api version just to show it works.

<pre>
package com.bmuschko.gradle.docker

import org.gradle.testkit.runner.BuildResult
import spock.lang.Requires

class DockerApiVersionFunctionalTest extends AbstractFunctionalTest {
    def "Can set Docker API version"() {
        buildFile << """
            import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask

            docker {
                apiVersion = 3.45
            }

            task dockerApiVersion(type: DockerApiVersionTask)

            class DockerApiVersionTask extends AbstractDockerRemoteApiTask {
                @Override
                    logger.quiet('Docker API Version: ' +
                        dockerClient.dockerCmdExecFactory.dockerClientConfig.version.asWebPathPart())
                }
            }
        """

        when:
        BuildResult result = build('dockerApiVersion')

        then:
        result.output.contains('Docker API Version: v3.45')
    }
}
</pre>

llamahunter:
Also, turns out that docker-java does eventually parse the apiVersion, but if it doesn't understand it, it just ignores it, which is treated the same as UNKNOWN_VERSION, a legal value.

cdancy:
@llamahunter yeah docker API does that in a few places actually and gives the impression that you did something correct when in fact you didn't. Ok ... let me give things a look over for myself and if nothing sticks out we'll take this as-is.

PR #186, opened 2016-03-24
Check if directory exists before trying to create it

liftyourgame:
mkdirs returns false if it doesn't need to create any directories.

I suggest we check if the path exists first so that we don't end up throwing any exception if the directory already exists


cdancy:
@liftyourgame good catch and thanks for the PR!

Issue #190, opened 2016-04-02
Remove requirement on specifying repositories for internal dependencies


cdancy:
To configure the plugin currently one has to do the following:

```
  repositories {
      mavenCentral()
  }

  buildscript {
      repositories { jcenter() }
      dependencies {
	      classpath 'com.bmuschko:gradle-docker-plugin:2.6.7'
      }
  }

  apply plugin: 'com.bmuschko.docker-remote-api'
```
Users, when applying a plugin, would not think that external repositories would be needed to resolve the plugin and its dependencies. This is needed due to the fact that we resolve our internal deps in an ad-hoc fashion to create a custom classloader. However, what users would expect is that the respositories defined within the buildscript-repository closure would be enough to resolve all the plugins dependencies. 

```
  buildscript {
      repositories {
	      jcenter()
	      mavenCentral()
      }
      dependencies {
	      classpath 'com.bmuschko:gradle-docker-plugin:2.6.7'
      }
  }

  apply plugin: 'com.bmuschko.docker-remote-api'
```
This is along the lines of what a user would expect to do when applying a plugin. When can do this easily enough within the plugin by adding whatever respositories are within the buildscript-repository closure to our custom configuration created internally.

@bmuschko thoughts?

bmuschko:
I think this is not a good idea. While this might look convenient, bigger organizations will not appreciate any pre-defined repositories in a plugin. In fact it's a bad practice. There are multiple problems here:

1) I usually see people running into the plugin that are completely new to Gradle. They don't understand the difference between the `buildscript` repositories and the top-level repositories and think they are the same. Another problem is that people simply don't read the documentation.

2) Some organizations will not want to resolve external dependencies more the same location that they resolved the plugin from. They might have different `repositories`. Predefining it for the top-level `repositories` block will lead to potential side-effects. If at all this should be somehow an option in Gradle core. I know that there a more improvements coming to the plugin mechanism in the next 6 months.

3) The plugin is actually available in 3 repositories: Maven Central, JCenter and the Gradle plugin portal. Many people do not define their plugin with the `buildscript` block. Instead they use the plugin DSL. If they do they will still not be able to resolve Docker Java dependencies as the Gradle plugin portal does not contain any third-party libraries.



    plugins {
        id "com.bmuschko.docker-remote-api" version "2.6.7"
    }

cdancy:
[...]

PR #191, opened 2016-04-02
internal dependencies are now resolved from buildscript repos.

cdancy:
Address #190


cdancy:
After further discussion on #190 we will not move forward with this PR.

Issue #198, opened 2016-04-18
Custom Classloader and AbstractDockerRemoteApiTask


marbon87:
I have the following class to run a docker exec command several times, until it runs succesful or a maximum number of retries is reached:
```
class DatabaseStartWait extends AbstractDockerRemoteApiTask {

    @Input
    @Optional
    def timeBetweenRetries

    @Input
    Integer maxRetries

    @Input
    String databaseName

    String containerId

    Integer exitCode

    DatabaseStartWait() {
        maxRetries = 20
        timeBetweenRetries = 4000
    }

    void targetContainerId(Closure containerId) {
        conventionMapping.containerId = containerId
    }

    @Override
    void runRemoteCommand(dockerClient) {
        String id = getContainerId()

        for (int i = 0; i < maxRetries; i++) {
            logger.debug "$i. try: Create Docker command to connect to database: $databaseName"
            def execCreateCmdResponse = dockerClient
                    .execCreateCmd(id)
                    .withAttachStdout(true)
                    .withCmd('su', '-c', "db2 'CONNECT TO $databaseName'", '-', 'db2inst1')
                    .exec()

            logger.quiet "$i. try: Remote command execution-ID: $execCreateCmdResponse.id"

            ClassLoader dockerClientClassLoader = dockerClient.getClass().getClassLoader()
            Class callbackClass = dockerClientClassLoader.loadClass("com.github.dockerjava.core.command.ExecStartResultCallback")
            def callback = callbackClass.getDeclaredConstructor(OutputStream.class, OutputStream.class).newInstance(System.out, System.err);

            dockerClient.execStartCmd(execCreateCmdResponse.getId())
                    .withDetach(false)
                    .exec(callback)
                    .awaitCompletion()

            def execResponse = dockerClient
                    .inspectExecCmd(execCreateCmdResponse.getId())
                    .exec()

            exitCode = execResponse.getExitCode()
            logger.debug "$i. try: ExitCode: $exitCode"

            if (exitCode == 0) {
                logger.debug "Database is started after $i. try"
                break;
            }
            sleep(timeBetweenRetries)
        }
        if (exitCode > 0) {
            throw new GradleException("Could not connect to database in the defined time")
        }

        logger.quiet "exitCode on database-connect: $exitCode"
    }

    @Input
    String getContainerId() {
        containerId
    }
}
```

I need to included the following, to get a callback-instance, that is loaded from the correct classloader
```
[...]

PR #199, opened 2016-04-19
Set originalClassLoader as parent for new URLClassLoader

marbon87:
functional test included


cdancy:
@marbon87 see my comments on #198. Instead of trying to side-step our custom classloader the better PR to put together would be one adding the `DockerExecContainer` task following the semantics we've already set.

PR #207, opened 2016-04-27
Adding Task for docker exec

lukaslentner:
Documentation was not updated

PR #221, opened 2016-06-05
Docker exec task

cdancy:
Fleshing out PR from #207


cdancy:
@lukaslentner I fleshed out your PR so that we can get the task in the next release.

Issue #232, opened 2016-06-23
Cannot wait for application in container to start before running, say, integration tests


aflowers001:
when building a spring boot application and then looking to run integration tests against the application deployed in a docker container i faced an issue whereby I had to wait for the application to be started in the container before commencing the tests.

I did not want to pollute the tests with and docker specific waits so I looked for a way to monitor the container for a specific event, but could not find anything that suited.

I got around this by updating a local deployment of the plugin with a new task that monitored the docker container stdout log looking for a specific string that indicated that the application had started up.

Is there a better alternative at all? If not I will tidy up my task and submit it for review. 

The basic premise was that it read the log, checked for the string it was after, then went to sleep and repeated until it found the required string indicating the container was ready for use.

cdancy:
@aflowers001 that's the way I've solved similar problems and have instructed others to do so as well. The standard we've set with this plugin is to provide a 1-to-1 (docker-command to gradle-task) and so we don't typically take any custom tasks so to speak. However, this is a common enough issue/question and I might entertain an enhancement to the `DockerLogsContainer` task that polls for some log output and finishes when it finds it. Feel free to put together a PR and ask any questions.

orzeh:
@aflowers001 also check the official [Docker Compose doc](https://docs.docker.com/compose/startup-order/) - there are links to some useful utilities that you can use to solve your problem.

@cdancy I agree that this is a common problem, so we should propose some "advised" solution with regards to docker-gradle-plugin.

cdancy:
We now have the ability to poll logs within a container and/or receive callbacks for each log output triggered. On top of this docker now supports the ability to do so natively. Think we can close this now IMO. Feel free to re-open for further discussion.

PR #234, opened 2016-06-27
Added support for simple monitoring of the container logs and waiting for specific message

aflowers001:
Added support for simple monitoring of the container logs in order to have the build wait for a specified string in the logs before continuing.

This can be used, for example, to have the execution of integration tests wait for the application deployed within a container to start.

There is a new task type, DockerWaitForLogMesageInContainer, which has 3
input settings,
- logSnippet - the string to look for in the logs, mandatory
- sleep - how long to sleep between requests for the log from the container, optional
- timeout - how long, in total, we are willing to wait for the message in the logs, optional

Looks to resolve issue #232


cdancy:
@aflowers001 I like the idea here though I don't think we need to put things into a separate class/task. Instead, lets user an inner class to capture the properties required here. If the properties are set, and thus the inner classes is not null, then we can proceed with polling. Something like the below should get you started:

    LogPoller poller;
    ..
    ..
    ..
    void poll(String regex, int sleep, int timeout) {
        poller = new LogPoller(regex, sleep, timeout);
    }
    ..
    @Override
        if (poller) {
            // poll code
        } else {
           // business as usual
        }
    }
  
Have a look at `DockerCreateContainer` to see how we use inner classes.

orzeh:
@aflowers001, @cdancy I think it would be better to do this in a more generic way. We should provide new task which would let user to use different _wait for container ready_ strategies.

Suppose I`d like to test my microservice which depends on two other - some database and let say some utility one. Tests should start when all components are _up_ and _ready_.  _Ready_ state is defined as follows:
* database container - it accepts TCP connections on port 1234
* utility container - it created read only file named _proc.pid_ in volume mounted to host 
* container under test - _STARTED_ is printed on _stdout_

The plugin should allow more or less DSL like this:
````groovy
task wait4Containers(type: DockerWaitForContainerIsReady) {
   dependsOn startContainers
   timeout = 100
   sleep = 10
   log {
       targetContainerId { testedContainer.getContainerId() }
       logSnippet = 'STARTED'
   }
   file {
       targetContainerId { utilityContainer.getContainerId() }
       path = "/path/to/volume/proc.pid"
   }
   tcp {
      targetContainerId { dbContainer.getContainerId() }
      port = 1234 
   }
}
````

What do you think?

**Comment by cdancy** [comment]
@orzeh this type of task I think would be better suited for a downstream plugin based on this. It's very docker-compose'ish and I don't think belongs in a "base" plugin like this.

However, I do like the basic premise of what you did. I've talked previously about putting together a "community edition" of a plugin which would use this as the base but build on top of it with custom functionality. Folks could then submit their ideas/PR's for more complicated tasks there. Something we could talk about more offline if ever you're interested.

**Comment by orzeh** [comment]
@cdancy you are right. This could be part of a _gradle-docker-test-plugin_ or something like that.

Actually, I've planned to start work on a plugin which could help me in testing my applications (microservices) with Docker and Gradle, so I'm interested in this topic. I`ll let you know if anything will be ready. Any ideas and thoughts are appreciated!

I like the concept of "community edition" of this plugin - I`ve got several useful tasks to share ;)

**Comment by aflowers001** [comment]
Hi,

A little late in commenting, day job panic getting in the way :-).

As the concept looks fine, I will take a look at the use of inner classes.

File: src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainer.groovy [hunk 1]
```diff
@@ -15,12 +15,103 @@
...
+      if(poller.timeout <= 0) {
+        throw new InvalidUserDataException("timeout should be greater than zero")
+      }
+  
+      if(poller.regex == null || poller.regex == '') {
+        throw new InvalidUserDataException("regex should be set to a valid string")
+      }
+    }
+    
+    void poll(dockerClient) {
```
cdancy: Should be private

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainer.groovy
```diff
@@ -15,12 +15,103 @@
...
         logger.quiet "Starting container with ID '${getContainerId()}'."
         def containerCommand = dockerClient.startContainerCmd(getContainerId())
         containerCommand.exec()
+
+        if (getPoller()) {
+            poll(dockerClient)
+        }
+    }
+    
+    void poller(Closure closure) {
```
cdancy: Lets instead use the actual parameters to the LogPoller class to make it absolutely obvious what needs to be passed in.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainer.groovy
```diff
@@ -15,12 +15,103 @@
...
+  
+      def loggingCallback = threadContextClassLoader.createLoggingCallback(sink)
+  
+      boolean finished = false
+      int howLongWaiting = 0
+  
+      while(!finished) {
+        logCommand.exec(loggingCallback)?.awaitCompletion()
+  
+        finished = sink.toString().contains(poller.regex)
```
cdancy: Shouldn't we want to `match` against the regex instead of using `contain`?

cdancy:
@aflowers001 added a few comments. I'm feeling like this functionality belongs within the `DockerLogsContainer` task and not here. Each task in this plugin serves a very specific point and no more. Anything "log" related I'd like to see within the task defined for it.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerStartContainer.groovy
aflowers001: Yep - thought about that after pushing, it says regex so it should BE a regex

otrosien:
Any chance of getting this feature into master in the near future?

cdancy:
@otrosien this functionality needs to go within the `DockerLogsContainer` task as I noted above.

aflowers001:
Been off doing other stuff, but will take a look at getting this moved into the DockerLogsContainer task

cdancy:
@aflowers001 before reviewing these changes lets wait until @orzeh finishes #307. His work on implementing the reactive method `onNext` will allow a callback of sorts for the `DockerLogsContainer` task for each message received.

cdancy:
Closing as the our `onNext` implementation should make this more/less possible. If folks want to somehow put in a hook to short-circuit the log polling that might be helpful to some.

PR #242, opened 2016-07-13
Fixing docker-java api compatibility issue.

roberttaylor426:
Right now specifying a memory limit on a container is broken (and apparently has been for some time). See https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/core/command/CreateContainerCmdImpl.java#L789.

This change addresses the compatibility issue and also renames the associated plugin config property, bringing it back into line with the naming used by the [Docker Remote Api](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/).

This change was tested on my own project against docker-java 3.0.0.


cdancy:
Thanks for the fix @roberttaylor426 ! I don't use this feature myself so have never encountered the hiccup.

PR #245, opened 2016-07-15
Allow to define devices when creating a container

### LINKED ISSUE #243: Allow to define a device mapping when creating a container
When defining a container we want to configure which devices from the host are accessible in the container. With the docker `run` command this would be the `--device` option.

mrhaki:
* To expose devices from the host system to a container.
* The docker-java library has support for a Device class to configure this.
* Extra devices property for DockerCreateContainer.
* Added code to dynamically load a Device class.

* Fixes #243

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
@@ -365,7 +365,7 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
         Class deviceClass = loadClass("${MODEL_PACKAGE}.Device")
         try {
             // The parse method is  available in newer Docker java library versions.
-            Method method = deviceClass.getMethod("parse", String.class)
+            Method method = deviceClass.getMethod("parse", String)
             method.invoke(null, deviceString)
         } catch (NoSuchMethodException) {
             // For older Docker java library versions we must parse the device string ourselves.
@@ -519,13 +519,12 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
     
     private def parseDevice(String deviceString) {
         List<String> tokens = deviceString.tokenize(':')
-        int numberOfTokens = tokens.size()
 
         String permissions = 'rwm'
         String source = ''
         String destination = ''
         
-        switch(numberOfTokens) {
+        switch(tokens.size()) {
             case 3:
                 if (validDeviceMode(tokens[2])) {
                     permissions = tokens[2]
@@ -559,12 +558,12 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
             return false
         }
 
-        for (char ch in deviceMode) {
-            final String mode = String.valueOf(ch)
-            if (!validModes.get(mode)) {
+        for (mode in deviceMode) {
+            if (!validModes[mode]) {
                 return false // wrong mode
+            } else {
+                validModes[mode] = false
             }
-            validModes.put(mode, false)
         }
 
         return true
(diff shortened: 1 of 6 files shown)


orzeh:
LGTM :+1: 
@cdancy review and merge if it's ok

cdancy:
@mrhaki what is the [reason for supporting older](https://github.com/bmuschko/gradle-docker-plugin/pull/245/files) versions of docker-java?

mrhaki:
@cdancy The 3.0.0 docker-java doesn't have the Device.parse method yet. It is already in the master branch, but not yet released.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -499,4 +516,57 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
+    }
+    
+    private boolean validDeviceMode(String deviceMode) {
+        Map<String, Boolean> validModes = [r: true, w: true, m: true]
+
+        if (!deviceMode) {
+            return false
+        }
+
+        for (char ch in deviceMode) {
```
cdancy: Can we do this in a more terse/groovy way?

	deviceMode.each {
		if (!validModes.get(it)) { return false
		} else { validModes.put(it, false) }
	}

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -499,4 +516,57 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
     private Class loadBindClass() {
         loadClass("${MODEL_PACKAGE}.Bind")
     }
+    
+    private def parseDevice(String deviceString) {
+        List<String> tokens = deviceString.tokenize(':')
+        int numberOfTokens = tokens.size()
```
cdancy: Do we need to create this extra variable? Can we not just use the size in the switch itself:

    switch(tokens.size()) {
        ...
    }

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -361,6 +361,23 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
      * {@inheritDoc}
      */
     @Override
+    def createDevice(final String deviceString) {
+        Class deviceClass = loadClass("${MODEL_PACKAGE}.Device")
+        try {
+            // The parse method is  available in newer Docker java library versions.
+            Method method = deviceClass.getMethod("parse", String.class)
```
cdancy: No need to use `String.class`. Simply use `String` and groovy should magically make things happen ;)

cdancy:
@mrhaki fair enough. A couple minor nit-picky syntactical issues to address and we should be good to merge.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
orzeh: this variable better express the intent, IMO it's ok

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
mrhaki: I made the code more terse, but still with the for loop, because otherwise the return statement doesn't return anything.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
mrhaki: The variable is not used anymore and tokens.size() also reads well, so I've removed the variable.

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
mrhaki: Removed the .class

cdancy:
LGTM and tests pass as expected. Thanks for the PR @MrHaki !

PR #250, opened 2016-07-27
Issue #237 Java system properties are not taken into account

### LINKED ISSUE #237: Java system properties are not taken into account
According to `FAQ` section `Executing the plugin’s test suite with custom configuration` there are few system properties that can configure Docker Engine location

* dockerServerUrl
* dockerCertPath

But seems they are not working, i.e. if I use `-DdockerServerUrl=..` when invoking `Gradle` the value is ignored.

eshepelyuk:
- added dependency on `JUnit System Rule` testing framework
- fallback chain for `dockerUrl`, `dockerCertPath` and `apiVersion` properties override and integrationg tests for API properties override
- DockerExtension doesn't populate `dockerUrl`, thus falling back to `Java Docker Library` defaults
- README adjustments

resolves #237

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -8,7 +8,7 @@ image:https://d3oypxn00j2a10.cloudfront.net/0.12.6/img/nav/docker-logo-loggedout
 
 Gradle plugin for managing link:https://www.docker.io/[Docker] images and containers using via its
 link:http://docs.docker.io/reference/api/docker_remote_api/[remote API]. The heavy lifting of communicating with the
-Docker remote API is handled by the link:https://github.com/docker-java/docker-java[Docker Java library]. Please
+Docker remote API is handled by the {dockerJavaLink}. Please
 refer to the library's documentation for more information on the supported Docker's client API and Docker server version.
 
 *This plugin requires Gradle >= 2.5 to work properly.*
@@ -132,6 +132,13 @@ The plugin defines the following extension properties in the `docker` closure:
 |`certPath`      |File      |null                     |The path to certificates for communicating with link:https://docs.docker.com/articles/https/[Docker over SSL].
 |`apiVersion`    |String    |null                     |The remote API version. For most cases this can be left null.
 |=======
+
+[WARNING]
+====
+Default value of `docker.url` property is deprecated.
+It will be replaced to `unix:///var/run/docker.sock` in upcoming releases.
+Please adjust your build scripts accordingly.
+====
 
 Image pull or push operations against the public Docker Hub registry or a private registry may require authentication.
 You can provide those credentials in the `registryCredentials` closure:
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -1,4 +1,7 @@
-Gradle Docker plugin
+= Gradle Docker plugin
+:dockerJavaUrl: https://github.com/docker-java/docker-java
+:dockerJavaLink: {dockerJavaUrl}[Docker Java library]
+:dockerJavaCfgLink: {dockerJavaUrl}#configuration[Docker Java library]
 ====================
 
 image:https://d3oypxn00j2a10.cloudfront.net/0.12.6/img/nav/docker-logo-loggedout.png[Docker Logo]
@@ -142,10 +145,28 @@ You can provide those credentials in the `registryCredentials` closure:
 |`email`         |String    |null                        |The registry email address.
 |=======
 
+=== Extension properties override
 
-==== Extension examples
+Properties defined within `docker` closure are providing flexible fallback mechanism
+allowing to reuse _the same build script_ on various environments with different `Docker Engine` settings,
+e.g. on CI servers, on Windows/Linux/Mac envs etc.
 
-===== Working with a TLS-enabled Docker instance
+Fallback tries to retrieve values from various source in following order
+(with higher precedence on top of the list)
+
+. Project properties, i.e. passed via command line with `-P` flag or defined in build script.
+. Java system variables, i.e. passed via command line with `-D` flag.
+. Environment variables.
+. Properties of currently executed `Docker` task.
+. `docker` closure defined on project or top project level.
+. Fallback to {dockerJavaLink} values.
+
+When overriding via project properties, system variables and environment properties -
+property names are following {dockerJavaCfgLink} conventions.
+
+=== Extension examples
+
+==== Working with a TLS-enabled Docker instance
 
 Starting with Docker version 1.3, TLS is enabled by default. Please consult the Docker documentation link:https://docs.docker.com/articles/https/["Running Docker
 with https"] to set up your certificate. The following example demonstrates how to configure the plugin to use those certificates.
@@ -166,7 +187,7 @@ docker {
 }
 ----
 
-===== Working with a Docker instance without TLS
+==== Working with a Docker instance without TLS
 
 The following example assumes that you disabled TLS on your Docker instance. You can do so by setting `DOCKER_TLS=no` in the file
  `/var/lib/boot2docker/profile`. Additionally, this code snippet shows how to set the user credentials.
@@ -178,7 +199,31 @@ docker {
 }
 ----
 
-=== Usage examples
+==== Overriding `docker` extension properties via command line
+
+Assuming we're having `build.gradle` with empty `docker` closure, we could do following
+
+Override Docker Engine location via project property::
++
+  $ gradle -PDOCKER_HOST=tcp://localhost:9999 dockerBuildImage <1>
++
+<1> tcp://locahost:9999 will be used
+
+Override Docker Engine location via environment varaible::
++
+  $ export DOCKER_HOST=tcp://locahost:5555
+  $ gradle dockerBuildImage <1>
++
+<1> tcp://locahost:5555 will be used
+
+Precedence rules for property override::
++
+  $ export DOCKER_HOST=tcp://locahost:5555
+  $ gradle -PDOCKER_HOST=tcp://localhost:9999 dockerBuildImage <1>
++
+<1> tcp://locahost:9999 will be used, since project properties have precedence over environment variables
+
+=== Task usage examples
 
 The following usage examples demonstrate code for common use cases. More scenarios can be found in the link:https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy[functional tests].
 
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
@@ -19,7 +19,7 @@ import org.gradle.api.file.FileCollection
 
 class DockerExtension {
     FileCollection classpath
-    String url = 'http://localhost:2375'
+    String url
     File certPath
     String apiVersion
 
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
@@ -19,7 +19,7 @@ import org.gradle.api.file.FileCollection
 
 class DockerExtension {
     FileCollection classpath
-    String url
+    String url = 'http://localhost:2375'
     File certPath
     String apiVersion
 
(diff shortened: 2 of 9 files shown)


cdancy:
Lots of changes here outside of the intended feature. Please remove those and submit a separate PR for that work.

There seems to be a lot of repeated code here within the Enum `ConfigOverrideOrder` and then how it's expected to be used. I've not given this too much thought but it feels like we should be able to do things in a simpler way?

eshepelyuk:
Hello @cdancy 

> Lots of changes here outside of the intended feature

I've rebased PR onto 3.0.2 release and removed everything that may seem *outside*. Could you please review again and comment inline, so I can try to explain intentions to any modifications.

> There seems to be a lot of repeated code here within the Enum ConfigOverrideOrder and then how it's expected to be used. I've not given this too much thought but it feels like we should be able to do things in a simpler way?

It seems you have a certain vision of how the code should look like. Can you share it so I can apply suggestions ?

src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy
```diff
@@ -57,16 +65,26 @@ class DockerThreadContextClassLoader implements ThreadContextClassLoader {
...
+    private File getCertPath(DockerClientConfiguration taskConfig) {
+        String certPath = project.properties.getOrDefault(DOCKER_CERT_PATH_PROPERTY_NAME,
+            System.getProperty(DOCKER_CERT_PATH_PROPERTY_NAME, System.getenv(DOCKER_CERT_PATH_PROPERTY_NAME)))
+        certPath ? new File(certPath) : (taskConfig.certPath ?: dockerExtension.certPath)
+    }
+
+    private String getApiVersion(DockerClientConfiguration taskConfig) {
+        String apiVersion = project.properties.getOrDefault(DOCKER_API_VERSION_PROPERTY_NAME,
+            System.getProperty(DOCKER_API_VERSION_PROPERTY_NAME, System.getenv(DOCKER_API_VERSION_PROPERTY_NAME)))
+        apiVersion ?: (taskConfig.apiVersion ?: dockerExtension.apiVersion)
```
orzeh: maybe we should move props extracting logic into a method:
````groovy
String getConfigProperty(String key) {
   project.properties.getOrDefault(key, System.getProperty(key, System.getenv(key)))
}
````

File: src/main/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoader.groovy [hunk 2]
```diff
@@ -19,19 +19,27 @@ import com.bmuschko.gradle.docker.DockerExtension
...
+    static final String DOCKER_CERT_PATH_PROPERTY_NAME = "DOCKER_CERT_PATH"
+    static final String DOCKER_API_VERSION_PROPERTY_NAME = "api.version"
+
     public static final String MODEL_PACKAGE = 'com.github.dockerjava.api.model'
     public static final String COMMAND_PACKAGE = 'com.github.dockerjava.core.command'
 
     private DockerExtension dockerExtension
 
-    public DockerThreadContextClassLoader(DockerExtension dockerExtension) {
+    private Project project
```
orzeh: could be `final`

eshepelyuk:
@orzeh suggestions are applied in a28734b

eshepelyuk:
@cdancy @orzeh Any feedback ?

src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
```diff
@@ -19,7 +19,7 @@ import org.gradle.api.file.FileCollection
 
 class DockerExtension {
     FileCollection classpath
-    String url = 'http://localhost:2375'
+    String url
```
orzeh: This breaks backward compatibility, so I have doubts if we should change it.

orzeh:
I've added comment about default value of `host` property - I think we shouldn't break backward compatibility. Besides this it LGTM.
@cdancy anything to add?

cdancy:
@orzeh @eshepelyuk I think we should hold off on this implementation. It feels overly complicated IMO and something I think we can accomplish with far less code. 

@eshepelyuk I get the idea, and where you're trying to go, it just feels too bloated. I haven't given this too much thought but the idea in my head is that we could do something like the below:

```
String getProperty(String... props) {
    def value = null
    for(String prop : props) {
        value = getProperty(prop)
        if (value) { break }
    }
    value
}

String getProperty(String prop) {
    def value = System.getEnv(prop)
    if (!value) {
        value = System.getProperty(prop)
        if (!value) {
            // some other search
        }
    }
    value
}
```

Very rough idea, and it could probably be made smaller with groovy magic, but that was the basic idea.

orzeh:
@cdancy but this is done in:
````
private String getConfigPropertyWithFallback(String key) {
   project.properties.getOrDefault(key, System.getProperty(key, System.getenv(key)))
}
````
I agree that @eshepelyuk implementation of tests is not easy to follow on first sight, but after a while it looks OK to me.

**Comment by cdancy** [comment]
@orzeh it's the repeating code in `ConfigOverrideOrder` that doesn't feel right. I like the concept just feels like we could do the same but simpler.

**Comment by eshepelyuk** [comment]
This is a *test* class used only for integration testing.

**Comment by cdancy** [comment]
@eshepelyuk I know but it's on us to maintain it and carry it forward and looking at this a year from now, should we need to make a change, may not be exactly intuitive for anyone else coming on board or even for ourselves.

With that said, @orzeh if you're Ok with this (which I believe you are), we can merge it in. @eshepelyuk there was one outstanding comment from orzeh to address. Any thoughts on it?

**Comment by eshepelyuk** [comment]
Hello @cdancy @orzeh about comment 
>>I've added comment about default value of host

One of the intention of this PR was to follow conventions and fall back to defaults of `docker-java-library` as much as possible.

`docker-java-library` provides convenient default value for the `DOCKER_HOST` property (`unix:///var/run/docker.sock`) which is the value used by `Docker Engine` on `Linux` computers.

From the other hand, current default value of `DOCKER_HOST` used by the plugin (`http://localhost:2375`) is quite odd, because

* it is not working for `Linux` (unix socket is used)
* neither it's acceptable for `Docker Toolbox` for `Windows` and `Mac OS` (2376 port is used and `Docker Engine` is not accessible via `localhost`, VM IP should be used instead)

I.e., currently plugin users are forced to explicitly setup proper `DOCKER_HOST` value in their `build.gradle`  via `docker.url` extension property.

Changing this property default value from `http://localhost:2375` to `null` gives us benefit of using defaults of `docker-java-library`, thus providing at least `Linux` users with out-of-box experience while applying plugin to their build script.

Additionally, this PR will release `Docker Toolbox` for `Win` and `Mac` users from the need to explicitly setup `docker.url` because `Docker Toolbox` sets up `DOCKER_HOST` environment variable properly.

As a result, plugin users for any OS don't need to setup `docker.url` in their build scripts for the most of the cases.

So, I believe we should sacrifice backward compatibility in this case in favor of less boilerplate plugin setup.

**Comment by cdancy** [comment]
@eshepelyuk I'm actually running this PR now to ensure all works as expected. I definitely do NOT want to break backward compatibility. A lot of folks, for better or worse, may have their scripts/plugins not setting the URL as they assume it will default to non-tls http://localhost:2375. We simply can't break this assumption at this point.

**Comment by eshepelyuk** [comment]
Unfortunately, I didn't get your point about TLS
* protocol of the URL defined by `docker.url` extension propery doesn't affect security because internally we replace both `http` and `https` to `tcp` before passing to `docker-java-library`
* If communication is secure is defined only by inspecting the presence of `docker.certPath`. See `DockerThreadContextClassLoader#getDockerClient()`
* `http://localhost:2375` is some sort of odd and artificial URL, to use it plugin users should have reconfigured their `Docker Engines` and I hardly believe they do it.

Basically, I believe that change won't affect existing users at all, but instead simplify experience of new ones.

Am I missing any scenario ?

**Comment by cdancy** [comment]
Meaning a lot of users could/were/are setting the URL to non-tls enabled daemon because it's _easier_ for dev/testing/etc. @eshepelyuk make no mistake: I agree with you 100% but I can't break backwards compatibility for those users/scripts/code who assume we will set this by default.

We can put in a message within the README.md to say that this functionality is deprecated and will be removed in a future release as we do want to encourage users to move away from this _standard_ that was set.

Did you want to fix things up? I planned on doing a release today but I will wait for you to address this, if you want me to, as I think it would be a great feature that folks could take advantage of.

**Comment by eshepelyuk** [comment]
Hi, I can try to revive old behaviour in an hour or so. Are you willing to wait ?

**Comment by cdancy** [comment]
@eshepelyuk let me give things a run to ensure all is Ok and then I will merge.

**Comment by eshepelyuk** [comment]
Sure, thank you.

**Comment by cdancy** [comment]
Running this branch skips the following tests:

com.bmuschko.gradle.docker.DockerJavaApplicationPluginFunctionalTest > Can create image for Java application and push to private registry SKIPPED

com.bmuschko.gradle.docker.DockerWorkflowFunctionalTest > Can build an image and push to private registry SKIPPED

File: README.asciidoc [hunk 4]
```diff
@@ -131,6 +134,13 @@ The plugin defines the following extension properties in the `docker` closure:
 |`apiVersion`    |String    |null                     |The remote API version. For most cases this can be left null.
 |=======
 
+[WARNING]
+====
+Default value of `docker.url` property is deprecated.
+It will be replaced to `unix:///var/run/docker.sock` in upcoming releases.
+Please adjust your build scripts accordingly.
```
cdancy: I think a more sensible default, at least for now, would be for `tcp://localhost:2375`. The purpose of using the remote API is to connect to engines regardless of their OS. Defaulting to unix assumes one is using a unix based OS and thus connecting to unix only docker-engine. TCP url on the other hand is more generic.

eshepelyuk: On Linux, one should enable tcp explicitly, since by default `Docker Engine` only listens to unix socket. And `docker-java` defaults to `unix://` as well.

Not sure, the plugin thusl provide own defaults in this case.

cdancy: That's assuming you're connecting to a local engine. You may not have docker installed at all locally and just want to connect to some remote instance to do work (we do that a lot here for CI related purposes).

eshepelyuk:
@cdancy May I ask if there any instructions how to run functional tests, especially how to run single test ?

cdancy:
@eshepelyuk there are directions on the very bottom of the README.md file on how to run functional tests.

README.asciidoc
eshepelyuk: In the latter case you've described `tcp://localhost:2375` is inappropriate too.

cdancy: > In the latter case you've described tcp://localhost:2375 is inappropriate too

How so?

In either event we're now back to square one again in that we can't because code/scripts/plugins/etc are expecting this to be the default and work as-is.

eshepelyuk:
I'm giving up for today. I was unable to run functional tests now on Windows to check.
Will try to get back to it tomorrow.
Can I ask you to check if those tests are not skipped on `master` branch code ?

cdancy:
@eshepelyuk they are not which is why I brought them up. You do need to have vagrant installed or at the very least point the tests at your docker instance (which needs to be non-tls at the moment).

eshepelyuk:
I think I have to do more work for this PR, i.e. make changes in functional tests to stop using custom variables `dockerHost` and `dockerCertPath` in `com.bmuschko.gradle.docker.TestConfiguration` and use `docker-java` conventions names, i.e. DOCKER_HOST / DOCKER_CERT_PATH, like production code does.


I've tried with my local `Docker` but I still got errors while initializing `TestPrecondition`. 
I'm using Windows with MSYS2, s maybe it's related issue.

Unfortunately have to dig into this later :(

README.asciidoc
eshepelyuk: You've written
>>That's assuming you're connecting to a local engine. You may not have docker installed at all locally and just want to connect to some remote instance to do work (we do that a lot here for CI related purposes).

If default will be `tcp://localhost:2375` - it will not work, because `Docker Engine` runs on other PC and is not accesible via `localhost` URL.

redi-foran: I agree with @eshepelyuk.  We should be coding to default docker behavior/configuration.  This will make it much easier for a docker newbie to start running and using this.  If he/she decides that further configuration is needed, then it should be done explicitly in docker and other dependent configurations.
Further, by specifying `localhost` you've limited this to running locally and not a remote machine anyway.  If that's the  case, then why not use a unix socket which is what docker is doing anyway rather than making more configuration changes.

cdancy: @redi-foran I agree in principle but we have backwards compatibility for existing users to think of. I'm in favor of defaulting to secure by default, as that's what docker does, but not without at least a few minor releases under our belt and giving folks ample time to adjust.

PR #254, opened 2016-08-03
Added several ports expose for java application

### LINKED ISSUE #253: Expose several ports for java application
How to expose several ports for java application?

```java
docker {
    javaApplication {
        port = 8080
    }
}

```
Port option for this task accepts only single integer value.

Salmondx:
resolve #253

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -18,6 +18,13 @@ package com.bmuschko.gradle.docker
 class DockerJavaApplication {
     String baseImage = 'java'
     String maintainer = System.getProperty('user.name')
+    @Deprecated
     Integer port = 8080
+    Set<Integer> ports = []
     String tag
+
+    Integer[] getPorts() {
+        return ports.size() > 0 ? ports : [port]
+    }
+
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
@@ -77,7 +77,7 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
             maintainer { dockerJavaApplication.maintainer }
             addFile({ tarTask.archivePath.name }, { '/' })
             entryPoint { determineEntryPoint(project, tarTask) }
-            exposePort { dockerJavaApplication.port }
+            exposePort { dockerJavaApplication.getPorts() }
         }
     }
 
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -268,7 +268,8 @@ The plugin defines the following extension properties in the `javaApplication` c
 |Property name   |Type      |Default value                                         |Description
 |`baseImage`     |String    |java                                                  |The Docker base image used for Java application.
 |`maintainer`    |String    |Value of `user.home` system property                  |The name and email address of the image maintainer.
-|`port`          |Integer   |8080                                                  |The Docker image entry point port used for the Java application.
+|`port`          |Integer   |8080                                                  |The Docker image entry point port used for the Java application (Deprecated)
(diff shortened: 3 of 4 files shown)


orzeh:
@Salmondx LGTM :+1: 
My only concern is that the property name `port` is now a little bit misleading, but on the other hand it's backward compatible. 

@cdancy do we prefer backward compatibility over clean API? if yes, please merge :smile:

bmuschko:
The variable name `port` is not very descriptive. It should be `ports` and we should probably provide different method signatures e.g. one that takes `Integer...` and one that takes `Iterable<String>`. I'd deprecate the old method/property and remove in a couple of versions.

Salmondx:
i've added `ports` property to DockerJavaApplication. Is this implementation is ok?

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -18,6 +18,17 @@ package com.bmuschko.gradle.docker
 class DockerJavaApplication {
     String baseImage = 'java'
     String maintainer = System.getProperty('user.name')
+    @Deprecated
     Integer port = 8080
+    Integer[] ports = []
     String tag
+
+    Integer[] getPorts() {
```
cdancy: Lets change the implemenation to something like the below.

    Integer[] getPorts() {
        def openedPorts = ports.clone()
        if (port && !openedPorts.contains(port)) { openedPorts.add(port) }
        openedPorts
    }

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -77,7 +77,7 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
             maintainer { dockerJavaApplication.maintainer }
             addFile({ tarTask.archivePath.name }, { '/' })
             entryPoint { determineEntryPoint(project, tarTask) }
-            exposePort { dockerJavaApplication.port }
+            exposePort { dockerJavaApplication.ports }
```
cdancy: Lets use the `getPorts()` method so that groovy won't try and access just the Integer array but instead use the function which takes into account both the Integer array and standalone Integer port.

cdancy:
@Salmondx added a couple comments. Once those are addressed please `squash + rebase` and I think we should be good to merge.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
Salmondx: @cdancy I've thought about this variant, and imho it's not so clean for end users cause they can use port and ports variable simultaneously. And if we want to deprecate port in future release, I don't think that's a good idea (several build script fixes pre and before plugin release).
But if you want, I'll add this.

cdancy: @Salmondx on second thought I think your point out weighs the points I was about to make ;) Lets leave your implementation. The only thing I would change is that when you're returning `port`do something like `return [port]` as what you're doing is returning the single instance instead of the expected array.

Salmondx:
@cdancy Done

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -18,6 +18,17 @@ package com.bmuschko.gradle.docker
 class DockerJavaApplication {
     String baseImage = 'java'
     String maintainer = System.getProperty('user.name')
+    @Deprecated
     Integer port = 8080
+    Set<Integer> ports = []
     String tag
+
+    Integer[] getPorts() {
+        if (ports.size() > 0) {
```
cdancy: Think we can use the ternary operator here to make this more terse `ports ? ports : [port]`

Salmondx:
Is it's ok now? Can you merge?

cdancy:
Integration tests run as expected. LTGM. Thanks for the PR @Salmondx !

Issue #238, opened 2016-07-07
Starting container with HostConfig was deprecated since v1.10 and removed in v1.12


kboom:
Hi All,

Just started to use the plugin. I have docker 1.11 installed on my PC.
When running DockerStartContainer task I'm getting the error like in the issue title.
Is this a known bug? How to resolve it? I don't think nobody has noticed it yet as it's major.

Here's my config if it can be of any help:
```
    docker {
        url = 'http://192.168.99.100:2376'
        certPath = file(System.getProperty('docker.cert.path') ?: "${System.getProperty('user.home')}/.docker/machine/machines/default")
    }

    task createDockerfile(type: Dockerfile) {
        destFile = project.file("$buildDir/docker/Dockerfile")
        from 'java:8u91'
        exposePort 8080
        defaultCommand 'java', '-jar', 'app.jar'
    }

    task buildImage(type: DockerBuildImage, group: 'docker') {
        dependsOn 'bootRepackage', 'createDockerfile'

        doFirst {
            copy {
                from "$buildDir/libs"
                into "$buildDir/docker"
                include 'orangebox-storage-1.0.0.jar'
                rename 'orangebox-storage-1.0.0.jar', 'app.jar'
            }
        }

        inputDir = file("$buildDir/docker")
    }

    task createContainer(type: DockerCreateContainer, group: 'docker') {
        dependsOn buildImage
        targetImageId { buildImage.getImageId() }
    }

    task startContainer(type: DockerStartContainer, group: 'docker') {
        dependsOn createContainer
        targetContainerId { createContainer.getContainerId() }
    }

    task stopContainer(type: DockerStopContainer, group: 'docker') {
        targetContainerId { createContainer.getContainerId() }
    }
```

Thanks!

orzeh:
I`m afraid it is [docker-java problem](https://github.com/docker-java/docker-java/issues/617), so we have to wait for them to fix it. 

You can try to specify Docker API version used, but I`m not sure if it will help.

cdancy:
@orzeh has it right. I was actually discussing this very thing [on this thread](https://github.com/docker-java/docker-java/issues/615). Unfortunately setting the API version has no effect ;(

ericweinberg:
I don't suppose there's any chance of not using the docker-java library?  Between this problem and the l[ack of support for docker native](https://github.com/docker-java/docker-java/issues/537), I think they're going to hurt adoption of this plugin.

cdancy:
[...]

PR #259, opened 2016-08-14
Bump docker-java to version 3.0.2

cdancy:


cdancy:
@orzeh I'd like your thoughts on this:

Bumping to docker-java 3.0.2 works but if executing against docker 1.12 (i.e. api 1.24) things break due to the issue #238. One can get around this by forcing, via the extension, the apiVersion to `1.23` which is the latest docker-java supports. I'm hesitant to merge this code as it will force bleeding edge users to a specific apiVersion if they want to use this plugin which I think is Ok it's just cumbersome to have to force them to set an extra config. I was thinking of adding a `DISCLAIMER` message on the README to give folks the heads up. 

Any thoughts one way or the other?

cdancy:
Actually this goes for our current version as well so maybe not such a big deal.

KostyaSha:
@cdancy try update to latest docker-java.

Issue #257, opened 2016-08-09
Applying plugin changes project repositories


stempler:
I'm having some trouble with the changes introduced in [this pull request](https://github.com/bmuschko/gradle-docker-plugin/pull/218).

In my Groovy project where I use the docker plugin this leads to the Gradle plugins Maven repository (http://plugins.gradle.org/m2) being added before any repositories that I specify in my build script.

Simplified build script:

```groovy
plugins {
  id 'groovy'
  id 'application'
  id 'com.bmuschko.docker-java-application' version '3.0.1'
}

repositories {
  // this needs to be defined before jcenter/MavenCentral for retrieving JAI
  maven {
    url 'http://download.osgeo.org/webdav/geotools/'
  }
  jcenter()
}
...
```

Usually that would not be a problem, but for Java Advanced Imaging (`javax.media:jai_core:1.1.3`) the order of repositories checked is quite important. If I update the Docker plugin to 3.0.0/3.0.1 for any of my projects having this dependency, the builds will fail, because Gradle tries to retrieve the dependency from the Gradle plugin repository where the POM is available, but not the JAR.

```
* What went wrong:
Could not resolve all dependencies for configuration ':compileClasspath'.
> Could not find jai_core.jar (javax.media:jai_core:1.1.3).
  Searched in the following locations:
      https://plugins.gradle.org/m2/javax/media/jai_core/1.1.3/jai_core-1.1.3.jar
```

Gradle complains when I try to move the `repositories` section before the `plugins` section, so the changes in #218 will always add the buildscript repositories to the project's repositories, before I can add them in the build script.

Why does the plugin need to adapt the project's repositories at all? Shouldn't the plugin use the buildscript repositories and dependencies?
If it is really necessary, can it be made configurable? Or maybe made to apply at a later stage, when repositories might have been added in the build script? (e.g. with `project.afterEvaluate { ... }`)

stempler:
If anyone else is having that problem, workaround is adding the repositories that should be checked first to the buildscript repositories:

```
buildscript {
  repositories {
    // these need to before jcenter/MavenCentral for retrieving JAI
    maven {
      url 'http://download.osgeo.org/webdav/geotools/'
    }
  }
}

plugins {
  id 'groovy'
  id 'application'
  id 'com.bmuschko.docker-java-application' version '3.0.1'
}

repositories {
  jcenter()
}
...
```

orzeh:
[...]

PR #260, opened 2016-08-14
Ensure addition of buildscript repos is executed after evalution.

cdancy:
Fix for #257

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
@@ -43,7 +43,9 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
         // if no repositories were defined fallback to buildscript
         // repositories to resolve dependencies as a last resort
         if (project.repositories.size() == 0) {
-            project.repositories.addAll(project.buildscript.repositories.collect())
+            project.afterEvaluate {
+                project.repositories.addAll(project.buildscript.repositories.collect())
+            }
         }
         
         Configuration config = project.configurations[DOCKER_JAVA_CONFIGURATION_NAME]


cdancy:
All functional tests pass as expected.

bmuschko:
My personal take on this functionality still is that it shouldn't part of this plugin or functionality provided by a different, cross-cutting plugin.

cdancy:
@bmuschko I understand your reasoning from past conversations but logically it does not seem to make sense for a plugin to get its dependencies/repositories from that of the thing we are building. There is a reason we (gradle) are separating the 2 in the first place. I grant you that there are use cases where some repos (e.g. corporate/etc), be they buildscript or otherwise, may be blocked and so we need to take another route to resolve artifacts. We still have that now we are just falling back to using the buildscript repos if no other project repos are defined.

src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
```diff
@@ -43,7 +43,9 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
         // if no repositories were defined fallback to buildscript
         // repositories to resolve dependencies as a last resort
         if (project.repositories.size() == 0) {
```
stempler: I still don't quite understand the rationale why the repositories are added, but should the check on the number of repositories then be done also in `afterEvaluate`? Otherwise it will apply even if the user added project repositories. The only possibility I can see right now for project repositories to have been added at that point would be if another plugin, that has been applied before the Docker plugin, added project repositories already.

orzeh: @cdancy I agree with @stempler that the repository count check should be inside `afterEvaluate` closure.

cdancy: Planned on doing another release this weekend as docker-java put out another release this morning fixing the "start container with post data" issue. I'll sneak this in.

PR #265, opened 2016-08-22
Print error messages received from Docker engine when build fails

### LINKED ISSUE #264: Plugin should report full error messages sent to docker client
I'm trying to construct a docker image, and the "buildImage" task fails with the following:
`* What went wrong:
Execution failed for task ':ordersImage:buildImage'.
> java.lang.reflect.InvocationTargetException-->null`

This obviously doesn't tell me anything useful.  If I add "--stacktrace", I get the following from the root cause:
`DockerClientException: Could not build image: lstat ordersService.war: no such file or directory`

I know why this is giving this error message, but the point is, I shouldn't have had to wade through the stacktrace to find this. Any error messages sent back to the embedded docker client should be reported to the user.

orzeh:
CGLIB proxies created in DockerThreadContextClassLoader catch InvocationTagretException and throw underlying exception, which contain error message from Docker engine. 
Fix #264

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -5,11 +5,32 @@ import com.bmuschko.gradle.docker.TestPrecondition
 import org.gradle.testkit.runner.BuildResult
 import spock.lang.Requires
 
-import static org.gradle.testkit.runner.TaskOutcome.SKIPPED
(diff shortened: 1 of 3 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -1,16 +1,36 @@
 package com.bmuschko.gradle.docker.tasks.image
-
```
cdancy: Typo in removing the space between package and imports?

cdancy:
@orzeh anything more we want to do here? LGTM.

orzeh:
no. please merge

PR #269, opened 2016-08-27
Various fixes and cleanups.

cdancy:


cdancy:
@orzeh care to look things over to see if you spot anything? Any minor fixes/cleanups we need to add that you can think of? Plan on doing another release tomorrow.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -484,7 +484,7 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
     private File createDockerfile(File imageDir) {
         File dockerFile = new File(imageDir, 'Dockerfile')
         dockerFile << """
-FROM ubuntu:12.04
+FROM ubuntu:14.04
```
orzeh: I just wonder if we can use `alpine` to reduce download size and speed up things?

cdancy: Giving that a go now...

PR #284, opened 2016-09-28
Added `simple-docker-plugin` with tests and docco

liviutudor:
I'm providing in this pull request to offer a simple plugin on top of the existing ones which offers a very simple approach to building and publishing a docker image for a java application.
The plugin simply creates all the tasks for creating a docker image for a java application and allows for some simple customization which I found myself to use all the time (version/tag as well as adding files to a docker image). It reduces the boilerplate code needed for creating this to just a couple of lines.
There are details in the docco but happy to answer any questions.


orzeh:
@liviutudor I don't think we should add such opinionated plugin here. Maybe you should release it as a separate plugin?

@cdancy WDYT?

cdancy:
@orzeh I agree completely. I actually meant to respond to this PR before but forgot about it (my apologies). Definitely belongs in a more opinionated plugin. @liviutudor maybe this is something you can host locally on an Artifactory instance or within your own branch and then publish to jcenter?

liviutudor:
Thanks guys. I have already spawned a nebula-docker-plugin in the Netflix nebula "ecosystem" however I thought it might make sense to have this outside the Netflix OSS stack, but I agree that the plugin is quite opinionated and might not have place here.
Thanks again.

PR #287, opened 2016-10-06
Added default unix url

develmusa:
For all not very experienced Docker users it would be an great time saver-


cdancy:
@develmusa thanks for the doc clarification!

PR #302, opened 2016-12-01
Implementation of reactive-stream 'onError' method/closure for all tasks.

cdancy:
@orzeh if you have time for a review I'd appreciate it. This is a simplified take on #153 to provide an `onError` closure for all tasks to take advantage of. The idea is to implement those methods from [reactive-streams-jvm](https://github.com/reactive-streams/reactive-streams-jvm/#2-subscriber-code) that make sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ out
 *.iml 
 *.ipr
 *.iws
+/.nb-gradle/
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -208,6 +208,27 @@ task buildImage(type: DockerBuildImage) {
 }
 ----
 
+=== Reactive-Streams
+
+As needed, we will implement reactive methods as described in link:https://github.com/reactive-streams/reactive-streams-jvm[reactive-streams]. We implement these here as optional closures for all tasks. Currently the only supported method is `onError`.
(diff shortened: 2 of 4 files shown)


orzeh:
I think it's a step in the right direction. However, I would change the handler name to `onError()`. Looking from a higher perspective it would be great to add another handlers like `onNext()` to for e.g. `DockerBuildImage` or `DockerLogsContainer` and `onComplete()` to `DockerInspectImage`. These handler names are the same as those in [Reactive Streams](https://github.com/reactive-streams/reactive-streams-jvm/#2-subscriber-code) specification. The benefit is we use common names developer might be familiar with.

cdancy:
@orzeh `onError` sounds good to me. Following the reactive model was exactly what I had in mind. `onError` I think is a good place to start as it's what others have asked for (myself included) and would solve a lot of "problems". We can implement the other reactive methods as needed.

orzeh:
:+1: 
I've started some work on that. I would prepare PR in near future.

cdancy:
@orzeh anything else we want to add here for `onError` implementation? The code is simple enough to understand from my point of view. Not sure if more tests/examples would help or are needed.

PR #307, opened 2016-12-05
Implement onNext and onComplete

orzeh:
This is my work-in-progress implementation of `onNext` method.

Currently only `DockerBuildImage`, `DockerLogsContainer` and `DockerListImages` can take advantage of defining `onNext` closure.


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerListImages.groovy
```diff
@@ -16,13 +16,12 @@
...
 
 import com.bmuschko.gradle.docker.response.ResponseHandler
-import com.bmuschko.gradle.docker.response.image.ListImagesResponseHandler
 import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 
 class DockerListImages extends AbstractDockerRemoteApiTask {
-    private ResponseHandler<Void, List<Object>> responseHandler = new ListImagesResponseHandler()
+    private ResponseHandler<Void, List<Object>> responseHandler
```
cdancy: Not really related to what you put together but I was thinking we just remove `ResponseHandler`. Ben had a use-case for this at one point but that never came to fruition. Don't think we really need it and this is a better solution IMO. Thoughts?

cdancy:
I like it so far. The only thing I'd like to talk about is how we can possibly make this work for all tasks. I hinted at this previously but maybe we define a standard for all tasks to implement (perhaps AbstractDockerRemoteApiTask extends from AbstractReactiveStreamsTask). For tasks which are not really "reactive" for the `onNext` method maybe we just pass along the returned response from the `docker-java` invocation thereby allowing downstream users to query for whatever information they want instead of the small amount of things we provide (hope that makes sense).

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerListImages.groovy
orzeh: Because there is a setter, I plan to leave it as is to be backward compatible in case someone provided custom handler. I will add deprecation and we can remove it in future release.

orzeh:
I agree in 100%. I've added `onNext` to `AbstractDockerRemoteApiTask` because defining separate class/interface is IMO _over-engineering_. For _non reactive_ tasks we should just return response as you suggest - see `DockerListImages` in this PR.

cdancy:
> class/interface is IMO over-engineering

Fair enough. I guess my thinking was that we could define an abstract task, which in turn had analogous abstract methods for tasks to override, that would force all downstream tasks to implement this functionality in exactly the same way if only to keep the look/feel of how to implement this similar and easy to understand for folks coming on board in the future.

In either event this is great and will be a huge win for folks. I assume you're going to knock out the rest of the tasks as part of this PR? If you need help just yell ;)

orzeh:
I plan to finish this in next days

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
```diff
@@ -57,15 +57,21 @@ class DockerCopyFileFromContainer extends DockerExistingContainer {
...
-            // if compressed leave file as is otherwise untar
-            if (compressed) {
-                copyFileCompressed(tarStream, hostDestination)
-            } else {
-                copyFile(tarStream, hostDestination)
+            if(onNext) {
+                onNext(tarStream)
+            }
+
+            if(!tarStream.isClosed()) {
```
cdancy: Should we do one or the other here? Meaning if the user has defined an `onNext` closure then we don't attempt to handle anything. Otherwise the user could leave things in an undefined state.

cdancy:
@orzeh new code looks good. One minor comment to talk about. 

As part of our "Reactive" documentation do you think it makes sense to document how each task implements the `onNext` methods. Not sure what this would look like but our tasks are implementing the method differently, and for good reason, I would just hate to have to field questions/ISSUE(s) from users who don't understand how things work and/or why we've implemented things in a given way.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
orzeh: You're right. I'll change this.

orzeh:
@cdancy I think we should describe in `README.md` how things works for the most _"reactive"_ tasks like building image or fetching container logs. For others there should be a link to [functional tests](https://github.com/bmuschko/gradle-docker-plugin/blob/15da459b944e8bc16897b7eecab6a603b2bd1c02/src/functTest/groovy/com/bmuschko/gradle/docker/DockerReactiveMethodsFunctionalTest.groovy).

BTW as my English is poor, may I ask you to help with the doc?

cdancy:
> BTW as my English is poor, may I ask you to help with the doc?

Of course! Let me know when you think you've got things all wrapped up and I'll put together the documentation on the `README.md`.

orzeh:
I think I've finished this. @cdancy please review and in spare time write some docs :smile:

cdancy:
@orzeh looks good so far. I'll send in a PR to your branch for the docs.

Any thoughts on tasks which don't have a sensible use-case for `onNext`? I'm thinking specifically of `DockerCopyFileToContainer`. Maybe allow for it, as I suspect users may do so, but instead pass in nothing? Something like a no-op. Just talking out loud here but I think we should do _something_ just not sure what that looks like yet.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -48,29 +48,42 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
...
     @Optional
     Closure onError
 
+    @Input
+    @Optional
+    Closure onNext
+
+    @Input
+    @Optional
+    Closure onComplete
```
cdancy: @orzeh the more I think about this the more I think it makes more sense to create an `AbstractReactiveStreamsTask`. This class will extend that and the `AbstractReactiveStreamsTask` will extend DefaultTask. I think, from a documentation perspective, it will be easier to talk at length about these tasks, and how we implement them, vs where things are now and it being intertwined with our current docs. 

Don't think is over-engineering as we are just adding 1 more class. But this will give us a single place to point devs to for information on how and/or why we are implementing these methods.

orzeh: OK, I'll refactor and we will see.

cdancy:
@orzeh how do you feel about it?

orzeh:
I think this implementation is better. Now we have good separation of concerns.

If there are no issues left, you can merge it.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractReactiveStreamsTask.groovy
```diff
@@ -0,0 +1,65 @@
...
+            } else {
+                throw possibleException
+            }
+        }
+        if(!commandFailed && onComplete) {
+            onComplete()
+        }
+    }
+
+    abstract void runProcessing()
```
cdancy: Rename to `runTask`?

orzeh: Hmm, task _is_ already running...

cdancy: @orzeh on second thought: lets just revert https://github.com/bmuschko/gradle-docker-plugin/pull/307/commits/196fd99c0a1807e5a27737ea8d89038958155f5c. The fact that we are even having this conversation I think is reason enough to back it out. If the naming is confusing to us, and it's adding yet another layer for folks to dig through, maybe it wasn't the best idea. If you're on board lets do that and then we are good to merge and then put out another release.

orzeh: Hmmm, I think extracting _"reactive"_ logic into its own class was a __good step__. The name of the abstract method is IMO secondary issue. It can be _whatever_. That being said I can change it to `runReactiveStream`, leave it as is or you can propose another one. Maybe `startEmitting` or `startSignalling`?

cdancy: @orzeh Ok. Lets do `runReactiveStream` and let that be it. I'll merge once done. We can always go back and change later.

cdancy:
@orzeh this really is great and I'm excited to use these new features. This will further remove loads of custom code we have here and be able to use more OOTB features. Thanks again!

PR #311, opened 2016-12-07
Support multiple variables per single ENV cmd

matez:
Add `environmentVariable(Map<String, String>)` and `environmentVariable(Closure)` methods to Dockerfile task. These two will generate ENV command in form

`ENV "foo"="bar" "baz"="bah"`

Original method `environmentVariable(String, String)` keep generating

`ENV foo bar`


cdancy:
@matez thanks for the PR! Any thoughts on combining `EnvironmentVariableInstruction` and `MultiEnvironmentVariableInstruction` into 1 class? I'm thinking we keep the name `EnvironmentVariableInstruction` but replace its contents with your Multi impl. Then massage the methods that use `EnvironmentVariableInstruction` the old way to work with your impl. Hope that makes sense.

matez:
I've added `MultiEnvironmentVariableInstruction` just to be able to extend `MapInstruction`. I'll update it to have only `EnvironmentVariableInstruction` supporting both approaches.

matez:
I can do

```groovy
static class EnvironmentVariableInstruction extends MapInstruction {
        EnvironmentVariableInstruction(String key, String value) {
            super([(key): value])
        }

        EnvironmentVariableInstruction(Map envVars) {
            super(envVars)
        }

        EnvironmentVariableInstruction(Closure envVars) {
            super(envVars)
        }

        @Override
        String getKeyword() {
            "ENV"
        }
    }
```

but it'll change old behavior for `environmentVariable(String, String)` from `ENV foo bar` to `ENV "foo"="bar"`. I am not sure if such change is breaking or not ... since dunno when docker introduced ENV  cmd with multiple varibales

matez:
Since 1.4.0 https://github.com/docker/docker/blob/master/CHANGELOG.md#140-2014-12-11

cdancy:
@matez I think that should be Ok as we don't support docker versions that old anyway. So long as all integration tests pass we should be good.

@orzeh any thoughts one way or the other?

matez:
81a19b5 shouldn't change any previous behavior ... There is only `EnvironmentVariableInstruction` used in all `environmentVariable` methods.

cdancy:
@matez honestly I had a completely different way of going about doing this in my head but I prefer you're take on things. LGTM! 

@orzeh second pair of eyes for a code review?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -18,6 +18,7 @@ package com.bmuschko.gradle.docker.tasks.image
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import org.gradle.internal.impldep.com.google.common.collect.Maps
```
cdancy: I don't see this import being used anywhere. Further more I think we would want to shy away from using anything internal to gradles API.

matez: Totally agree ... missed that :(

orzeh:
I don't see any issues. @matez thanks for this PR!

@cdancy please merge.

cdancy:
All integrations tests pass as expected (couple failures but un-related and will cleanup in a separate PR). Thanks @matez !

Issue #338, opened 2017-01-26
Dockerfile task failing when FROM instruction is not the first instruction


thijslemmens:
Behavior:
When some instructions, and then adding a FROM instruction, the validation fails when the task is executed.

Improvement:
Separate out the FROM instruction in a separate field. Run that instruction first and then the list of other instructions. That way, you can change the FROM at any time, it will always be the first instruction.

Why:
I use a chain of builds, and I add the imageId of a previous build later in the process  (because it is not available at the time of constructing the task). Now I use a workaround by storing all the instructions and adding them at the moment I have the "FROM" instruction. It works, but it makes my code less readable.

If you like the improvement, I can send a pull request.

cdancy:
@thijslemmens yes please do send in a PR!

PR #339, opened 2017-01-26
Added possibility to use FROM instructions at any time in Dockerfile …

thijslemmens:
…task. It will always end on top of the actual file

Looks to solve #338

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
@@ -31,22 +31,36 @@ class Dockerfile extends DefaultTask {
 
     @TaskAction
     void create() {
+        Instruction fromInstruction = findFromInstruction()
         verifyValidInstructions()
 
         getDestFile().withWriter { out ->
+            out.println fromInstruction.build()
             getInstructions().each { instruction ->
-                out.println instruction.build()
+                if(instruction.keyword != 'FROM'){ //FROM instruction already executed
+                    out.println instruction.build()
+                }
             }
+        }
+    }
+
+    private Instruction findFromInstruction(){
+        Instruction fromInstruction
+        for(Instruction instruction: instructions){
+            if(instruction.keyword == 'FROM'){
+                fromInstruction = instruction
+            }
+        }
+        if(fromInstruction == null){
+            throw new IllegalStateException('Please specify a FROM instruction for your Dockerfile')
+        } else {
+            return fromInstruction
         }
     }
 
     private void verifyValidInstructions() {
         if(getInstructions().empty) {
             throw new IllegalStateException('Please specify instructions for your Dockerfile')
-        }
-
-        if(!(getInstructions()[0].keyword == 'FROM')) {
-            throw new IllegalStateException('The first instruction of a Dockerfile has to be FROM')
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -31,23 +31,37 @@ class Dockerfile extends DefaultTask {
...
+                }
+            }
+        }
+    }
+
+    private Instruction findFromInstruction(){
+        Instruction fromInstruction
+        for(Instruction instruction: instructions){
+            if(instruction.keyword == 'FROM'){
+                fromInstruction = instruction
```
cdancy: We can just return at this point. No point in letting the for loop continue.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -31,23 +31,37 @@ class Dockerfile extends DefaultTask {
...
             }
         }
+        if(fromInstruction == null){
+            throw new IllegalStateException('Please specify a FROM instruction for your Dockerfile')
+        } else {
+            return fromInstruction
+        }
```
cdancy: If we return from inside the for loop we can just thrown the `IllegalStateException` if we reach this point.

cdancy:
@thijslemmens couple minor comments to address but looks good so far.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
thijslemmens: My idea was using the last from instruction, making it possible to override. I added also some extra docs to the from methods for this.

cdancy: hmmm ... Ok I'll go with that. I can't imagine the use-case off the top of my head but it sounds reasonable to allow for this to happen.

thijslemmens: I do not have a use-case for it right now, but it was easy enough not to block it.

thijslemmens:
Now I can see a problem with my own suggestion:
You can have multiple times FROM in one Dockerfile
https://docs.docker.com/engine/reference/builder/#/from states
FROM can appear multiple times within a single Dockerfile in order to create multiple images. Simply make a note of the last image ID output by the commit before each new FROM command.

I cannot imagine that I would want to use it, but it is something that my code does not support.

cdancy:
@thijslemmens hmmm ... let me think on this. I've never actually built multiple images from a single Dockerfile and in fact did not know you code even do so ;)

thijslemmens:
I asked a bit around, and I have not found anyone else that even knew of the multiple 'FROM' feature.
That said, I figured out that I could have solved the issue with the delayed from with a closure, more elegantly, so maybe, this PR is just an extra limitation.
That said, we could improve the error feedback to an earlier stage when using the from() methods and the instruction(String) method. In that case we can already detect weather the first instruction is a FROM or not. We can already fail in the pre-execution phases.

cdancy:
@thijslemmens lets close this for now as you're able to move forward and the multiple FROM's options is still not yet clear to me how that would work given this context. Maybe best to just let things fail as they normally do.

PR #348, opened 2017-02-21
Improve README. Add missing '/' in default registryCredentials

aleskiewicz:
Default links in documentation were missing the closing slash. It should be: 'https://index.docker.io/v1/' as in gradle-docker-plugin/src/main/groovy/com/bmuschko/gradle/docker/DockerRegistryCredentials.groovy


cdancy:
@aleskiewicz thanks for the doc fix!

PR #355, opened 2017-03-08
Using the class name for the plugin

alvarosanchez:


orzeh:
@alvarosanchez It was obvious bug, thanks for the fix!

PR #362, opened 2017-04-18
Inspec exec container

### LINKED ISSUE #361: Exit codes for TaskType:DockerExecContainer
I have tried to use this for my project, one of my requirement is to exec into a container and run a script, check for exit codes if the scripts fails(return a non-zero exit code) and fail the task. Is there a way to check exit codes for taskType:DockerExecContainer if cmd[] fails?

orzeh:
Adds new task DockerInspectExecContainer which allows to inspect exec instance.

This should fix #361

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.groovy
@@ -20,38 +20,45 @@ import org.gradle.api.tasks.Optional
 
 class DockerExecContainer extends DockerExistingContainer {
 
-  	@Input
-  	String[] cmd
+    @Input
+    String[] cmd
 
-  	@Input
-  	@Optional
-  	Boolean attachStdout = true
+    @Input
+    @Optional
+    Boolean attachStdout = true
 
-  	@Input
-  	@Optional
-  	Boolean attachStderr = true
+    @Input
+    @Optional
+    Boolean attachStderr = true
 
+    String execId
 
-  	@Override
-	void runRemoteCommand(dockerClient) {
-  		logger.quiet "Executing '${getCmd()}' on container with ID '${getContainerId()}'."
-    	def execCallback = onNext ? threadContextClassLoader.createExecCallback(onNext) : threadContextClassLoader.createExecCallback(System.out, System.err)
-    	def execCmd = dockerClient.execCreateCmd(getContainerId())
-    	setContainerCommandConfig(execCmd)
-    	dockerClient.execStartCmd(execCmd.exec().getId()).withDetach(false).exec(execCallback).awaitCompletion();
-  	}
+    DockerExecContainer() {
+        ext.getExecId = { execId }
+    }
 
-  	private void setContainerCommandConfig(containerCommand) {
-  	    if (getCmd()) {
-  	    	containerCommand.withCmd(getCmd())
-  	    }
+    @Override
+    void runRemoteCommand(dockerClient) {
+        logger.quiet "Executing '${getCmd()}' on container with ID '${getContainerId()}'."
+        def execCallback = onNext ? threadContextClassLoader.createExecCallback(onNext) : threadContextClassLoader.createExecCallback(System.out, System.err)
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectExecContainerResponseHandler.groovy
```diff
@@ -0,0 +1,21 @@
...
+
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.Logging
+
+/**
+ * Logs inspect exec container result.
+ *
+ * @see com.bmuschko.gradle.docker.tasks.container.DockerInspectExecContainer
+ */
+class InspectExecContainerResponseHandler {
```
cdancy: I'd actually push back against adding this as I think we want to move away from using this form of callback in favor of the work you did with reactive-streams which IMO is a much better way to go about such things. Thoughts?

orzeh: This is just a fallback which prints result to stdout (via logger) when `onNext` is not defined. IMO it is consistent with current implementation of other tasks.

orzeh:
@cdancy I've moved it to separate task because of two reasons:
* the _exec_ can take some time to finish
* user may not be interested in _exec_ exit code and other details at all

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectExecContainerResponseHandler.groovy
cdancy: Just getting back around to this now. I really think we should remove this in favor of using an `onNext` block and if none is passed than create one. Within a few releases I'd like to remove any use of the "response handler" objects/concept as I believe the reactive-streams impl is a superior solution. Keeping it in, and/or adding to it, may give the wrong impression that we want to keep this around.

Aside from that: everything else looks good.

cdancy:
Would it make sense to put this functionality within the `DockerExecContainer` task and provide some sort of hook to make it work? I know it would break our "1 task" principle but it might be worth it in this case.

src/main/groovy/com/bmuschko/gradle/docker/response/container/InspectExecContainerResponseHandler.groovy
orzeh: ok, I'll rework this and remove _response handler_

orzeh:
It would be hard to define nice DSL in `DockerExecContainer` task. I think it is better to have this logic separated as it is more flexible solution.

cdancy:
@orzeh fair enough and I actually agree though I was secretly hoping you would have provided a compelling enough reason (I couldn't think of one) for us to break the "1 task" rule ;)

orzeh:
@cdancy I've removed logging handler. Please review.

cdancy:
All tests pass as expected. Thanks again @orzeh !

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.groovy
```diff
@@ -20,38 +20,45 @@ import org.gradle.api.tasks.Optional
...
+    Boolean attachStdout = true
 
-  	@Input
-  	@Optional
-  	Boolean attachStderr = true
+    @Input
+    @Optional
+    Boolean attachStderr = true
 
+    String execId
```
cdancy: Just merged this PR but wondering if we should make this private and provide a method to retrieve it's value? Might give the impression that it can or should be set by developers.

orzeh: good point, I'll fix it soon.

seb-koch:
Is there an example on how to use this new task? Struggling to get this up and running.

orzeh:
@seb-koch Just see [functional test](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/DockerInspectExecContainerFunctionalTest.groovy)

PR #363, opened 2017-04-24
Remove an erroneous claim

eljobe:
The following snippet does not actually explain how to configure
credentials. I suspect this was a simple copy/paste error.


cdancy:
@eljobe you're absolutely right and thanks for the PR!

PR #370, opened 2017-05-14
Logs container preserve leading space

double16:
Using `trim()` on log messages was intended to remove double line spacing. It also removes leading space, which makes stack traces and other indented text hard to read.


cdancy:
Our first PR with travis wired up and functional/Integration tests all working as expected! Tis a beautiful thing ;)

Anyway ... totally agree and thanks for the PR @double16 !

PR #379, opened 2017-05-18
Expose method imageId for DockerPullImage task to better fall inline …

cdancy:
…with DockerBuildImage expectations.

The previous way (which still works) to grab the imageId from the `DockerPullImage` task was to do something like:

    targetImageId { pullImage.repository + ":" + pullImage.tag }

Now developers can do:

    targetImageId { pullImage.getImageId() }

This shortens things up immensely especially if you have long verbose task names (like we do here ;)). This better falls inline with developers expectations having used the `DockerBuildImage` task.


orzeh:
LGTM :+1:

cdancy:
Thanks @orzeh !

PR #380, opened 2017-05-18
Allow to build image with multiple tags

### LINKED ISSUE #286: Ability to tag a docker build with multiple tags
Docker 1.10 added the ability to tag a build with multiple tags. Doing so is convenient as it reduces the number of tasks needed in a project (for example when building and using both "latest" and "<version>" tags)

It would be nice if the plugin could support this. For docker 1.10+ it would just use the multi-tag approach. For docker < 1.10 it could run the build step then run a series of Actions afterwards that tag the image behind the scenes.

orzeh:
Adds `tags` property to `DockerBuildImage` task and deprecates `tag` property.
To preserve backward compatibility if `tag` is defined, `tags` are ignored.
Fix #286

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -1,9 +1,7 @@
 package com.bmuschko.gradle.docker.tasks.image
 
 import com.bmuschko.gradle.docker.AbstractFunctionalTest
-import com.bmuschko.gradle.docker.TestPrecondition
 import org.gradle.testkit.runner.BuildResult
-import spock.lang.Requires
 import spock.lang.Unroll
 
 class DockerBuildImageFunctionalTest extends AbstractFunctionalTest {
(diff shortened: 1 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -55,13 +53,26 @@ class DockerBuildImageFunctionalTest extends AbstractFunctionalTest {
...
+
+
     private String imageCreation() {
         """
             import com.bmuschko.gradle.docker.tasks.image.Dockerfile
             import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
 
             task dockerfile(type: Dockerfile) {
-                from 'ubuntu:12.04'
+                from 'alpine'
```
cdancy: +1

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -95,9 +104,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
...
         }
 
-        if (getTag()) {
+        if(getTag()) {
             logger.quiet "Using tag '${getTag()}' for image."
             buildImageCmd.withTag(getTag())
+        } else if (getTags()) {
+            def tagListString = getTags().collect {"'${it}'"}.join(", ")
+            logger.quiet "Using tags ${tagListString} for image."
+            buildImageCmd.withTags(getTags())
```
cdancy: Can we allow for both? Maybe something like:

```
def allTags = [] + getTags()
if (getTag()) { allTags << getTag() }
if (allTags) {
    logger.quiet "Using tags ${tagListString} for image."
    buildImageCmd.withTags(allTags)
}
```
May be a better more succinct way to write this but I can def see folks defining a default tag and then possibly through a `doFirst` block add additional tags at runtime.

cdancy:
I take that back. Doing so is going to promote bad behavior and force us to maintain both ways for longer than is necessary. Merging now.

orzeh:
> Doing so is going to promote bad behavior and force us to maintain both ways for longer than is necessary
Exactly!

PR #384, opened 2017-06-05
Add networkAliases option to DockerCreateContainer

jeysal:
Allow assigning network aliases to containers, enabling network access inside the network via these set domain names.
https://docs.docker.com/engine/userguide/networking/configure-dns/

```groovy
project.task('createDockerContainer', type: DockerCreateContainer) {
    // ...
    networkMode = 'some-network'
    networkAliases = ['some-alias']
}
```

I'm assuming this doesn't require an additional test case since we're just delegating to docker-java (networkMode for instance isn't tested either).

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -47,7 +47,7 @@ abstract class AbstractFunctionalTest extends Specification {
         setupDockerServerUrl()
         setupDockerCertPath()
         setupDockerPrivateRegistryUrl()
-        
+
         buildFile << """
             task dockerVersion(type: com.bmuschko.gradle.docker.tasks.DockerVersion)
         """
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -115,6 +115,10 @@ abstract class AbstractFunctionalTest extends Specification {
         generateRandomUUID()
     }
 
+    protected String createUniqueNetworkName() {
+        generateRandomUUID()
+    }
+
     private String generateRandomUUID() {
         UUID.randomUUID().toString().replaceAll('-', '')
     }
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -494,6 +494,70 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
         result.output.contains("Devices : [/dev/sda:/dev/xvda:rwm]")
     }
 
+    def "Can build an image, create a container and assign a network and alias"() {
+        File imageDir = temporaryFolder.newFolder('images', 'minimal')
+        File dockerFile = createDockerfile(imageDir)
+
+        String uniqueContainerName = createUniqueContainerName()
+        String uniqueNetworkName = createUniqueNetworkName()
+
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
+            import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+            import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+            import com.bmuschko.gradle.docker.tasks.container.DockerInspectContainer
+
+            class DockerCreateNetworkTask extends AbstractDockerRemoteApiTask {
+                @Override
+                void runRemoteCommand(dockerClient) {
+                    dockerClient.createNetworkCmd().withName("${uniqueNetworkName}").exec()
+                }
+            }
+            class DockerRemoveNetworkTask extends AbstractDockerRemoteApiTask {
+                @Override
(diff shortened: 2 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -103,6 +102,10 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
 
     @Input
     @Optional
+    String[] networkAliases
+
+    @Input
+    @Optional
```
cdancy: What is this change for?

cdancy:
@jeysal thanks for the PR!

> networkMode for instance isn't tested either

What do you mean by this? That docker-java is not testing network related functions?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
jeysal: I added the networkAliases property to the task that can be set to assign aliases to the container.
Other containers in the network will then be able to reach it using that alias, e.g. `http://some-alias/`.

jeysal: In case git's weird diff rendering is confusing you here, this is really just
```
+ @Input
+ @Optional
+ String[] networkAliases
```

jeysal:
> What do you mean by this? That docker-java is not testing network related functions?

Sorry for being unclear, I meant that there is no test for `networkMode` in gradle-docker-plugin, because it is really just passed to `docker-java` which handles it.
For that reason, I'm assuming we don't need to test `networkAliases` either because docker-java knows how to deal with it.

cdancy:
@jeysal would still want a test to ensure things are as you'd expect them to be. I assume you could get away with writing a test to start the container and then do an "info" on said container to check the network. Let me know if you need any help.

jeysal:
I've added a test using using `DockerInspectContainer` and `onNext` to check that network and alias are assigned.
Let me know if there's still something missing.

jeysal:
I'll check why the build failed when I have time.

cdancy:
Ran this PR on my mac and it worked as expected (docker ver 17.03.1-ce) while on my linux box (docker ver 1.12.6) it failed due to: `network some-network not found`.

jeysal:
Interesting - I only tried it on my Linux machine and it worked.
I was surprised that docker didn't mind the missing network but I'm even more surprised this behavior appears to be system dependent.
I'll look into this - might have to create a network first and tear it down afterwards...

jeysal:
I've added network setup and teardown to the test. I had to extend the abstract task because the plugin has no tasks for network yet, so unfortunately the test is pretty large now. But I guess this is an appropriate "workflow" test close to what users might do - I've actually used the plugin for something like this before.

CI still doesn't pass though - @cdancy can you check what the error might be on 1.12.6? Travis log is not very helpful.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -494,6 +494,70 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
...
+                dependsOn buildImage, createNetwork
+                targetImageId { buildImage.getImageId() }
+                containerName = "${uniqueContainerName}"
+                networkMode = "${uniqueNetworkName}"
+                networkAliases = ["some-alias"]
+                cmd = ['/bin/sh']
+            }
+
+            removeNetwork.mustRunAfter createNetwork
+            createNetwork.finalizedBy removeNetwork
```
cdancy: This is your problem. You're tearing down the network just after you create it! :)

Do something like `inspectContainer.finalizedBy removeNetwork` and you should be good to go. Just ran this setup on my linux machine and it worked as expected.

cdancy:
Made a comment on how to fix. The test setup is fine. You should actually send in another PR to add support for creating and tearing down networks in a separate PR when you have the time. Think lots of folks might find that useful.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
jeysal: Whoops, that was not meant to be `createNetwork`. Thanks for the comment, changing it RN.

jeysal:
Alright, build is passing, thanks for the help @cdancy! 🎉
Might implement some network tasks for the plugin when I've got some spare time.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -115,6 +115,10 @@ abstract class AbstractFunctionalTest extends Specification {
         generateRandomUUID()
     }
 
+    protected String createUniqueNetworkName() {
```
cdancy: Can we not use method `generateRandomUUID()`?

jeysal: Did this to be consistent with container names, which also have an abstraction that simply delegates to `generateRandomUUID` right above this method.

cdancy: Ok fair enough. Lets keep it then.

cdancy:
Thanks for the PR @jeysal !

PR #405, opened 2017-07-05
Fix publishing JavaDoc

orzeh:
This PR should allow to publish JavaDoc to `gh-pages` branch from Travis job.

@bmuschko can you define those two secure variables in Travis:
* `GRGIT_USER=bmuschko`
* `GRGIT_PASS=[GITHUB_TOKEN]`

and create GitHub [access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/) with `public_repo` scope so Travis can push to this repository?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
       script: skip
       deploy:
         provider: script
-        script: ./gradlew -s assemble bintrayUpload gitPublishPush
+        script: ./gradlew -s assemble bintrayUpload gitPublishPush -Dorg.ajoberstar.grgit.auth.username=bmuschko -Dorg.ajoberstar.grgit.auth.password=$GH_TOKEN
         on:
           tags: true
 
--- a/doc/RELEASE.md
+++ b/doc/RELEASE.md
@@ -14,13 +14,13 @@ The [`gradle-git-publish`](https://github.com/ajoberstar/gradle-git-publish) Gra
 [Travis CI](https://travis-ci.com) service is used as our current [CI/CD](https://en.wikipedia.org/wiki/CI/CD) server. Build and deploy jobs are configured in [.travis.yml](.travis.yml) file. Please refer its [documentation](https://docs.travis-ci.com/) for more details.
 
 ## Bintray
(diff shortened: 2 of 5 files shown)


cdancy:
LGTM

bmuschko:
@orzeh I already the environment variables a while ago. However, the variable defining the GitHub token is called `GH_TOKEN`. You can just hard-code the user variable for now.

cdancy:
@orzeh can you still define `release.stage=final` in the `gradle.properties` file but comment it out with a note saying it needs to be enabled/passed at release time? I'd still like to have it defined even if it is commented out.

cdancy:
@orzeh merging this in now as I plan to do another release with latest `docker-java`.

cdancy:
Think we have to revert this commit. Attempting to release with it popped:

https://travis-ci.org/bmuschko/gradle-docker-plugin/builds/251195000

cdancy:
@orzeh so something odd is happening. I've been playing around trying to get this to work and it created tag `v3.0.10` but created release `v3.0.11`? I can't seem to figure out where it's getting the default patch version to use so that it can be bumped.

@bmuschko looks like there was a `3.0.11-dev*` version released on bintray we may want to delete as well.

orzeh:
@cdancy OK, I know what was wrong: when we added `release.scope=patch` in `gradle.properties` built-in strategy called `RebuildVersionStrategy` [can't be used](https://github.com/ajoberstar/gradle-git/blob/master/src/main/groovy/org/ajoberstar/gradle/git/release/semver/RebuildVersionStrategy.groovy#L54). Before our changes, it was used on Travis release job: git head was on tag and repo was clean -> inferred version was the same as tag. 

After you reverted this PR you set default version strategy to `FINAL`. Then on Travis release job this strategy was selected as fallback, because - as before `RebuildVersionStrategy` can't be used and git repo was clean -> version was increased at `patch` because `release.scope`. Thus tag `3.0.10` and version `3.0.11`. It's a little bit complicated, but this is how it works :smile: .

It will fix it here.

cdancy:
Thanks @orzeh !!! My apologies if I got thinks out of whack as I was just trying to do a release (which I eventually did) and read the docs at the same time and fix a fire drill here at work :)

orzeh:
@cdancy no problem :) I pushed new commits to  `FIX-PUBLISH-DOC` branch and going to submit PR one more time.

bmuschko:
@cdancy I just deleted 3.0.11-dev.0+f5b8053 from Bintray.

cdancy:
+1

PR #418, opened 2017-07-13
Corrected JavaDoc for `arg` method

pauxus:


cdancy:
@pauxus good eye thanks for the doc fix!!!

Issue #415, opened 2017-07-11
multiple ENVs are not set the same way as like single ENV instructions


childnode:
When
```groov
dockerfile(Type:Dockerfile) {
    environmentVariable "foo" "bar"
}
```

we get 
```Dockerfile
ENV FOO bar
```

but with
```groov
dockerfile(Type:Dockerfile) {
    environmentVariable [ "foo":"bar" ]
}
```

we get 
```Dockerfile
ENV "FOO"="bar"
```

This is caused by the [DockerFile.MultiItemJoiner](https://github.com/bmuschko/gradle-docker-plugin/blob/v3.0.11/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy#L510)

referring to https://docs.docker.com/engine/reference/builder/#env and for consistency with the SIngleItemJoiner the code should be
```patch
     static class MultiItemJoiner implements ItemJoiner {
         @Override
         String join(Map map) {
-             map.inject([]) { result, entry -> result << "\"$entry.key\"=\"$entry.value\"" }.join(' ')
+             map.inject([]) { result, entry -> result << "$entry.key=$entry.value" }.join(' ')
         }
     }
```

anyway: this code will (no more) cover spaces in values. but the singleItemJoiner doesn't care about that neither.
both don't care about escaping characters
...another two points we might address

~Marcel

orzeh:
@childnode Thanks for finding this bug! Could you send PR with the fix?

childnode:
just with the line change as described in the patch? sure … next night ;)

orzeh:
@childnode It would be great if you address all issues, i.e. support for spaces in values and escape characters :smile: but lets start with this one line.

childnode:
sorry for the delay, but have had some troubles to get the tests started ... and to get a free evening ;)
so I stumbled over some cases in 
https://docs.docker.com/engine/reference/builder/
1. single `ENV <key> <value>`  instruction
1.a) simple case with a simple word value without any linebreaks. Special Chars (quotes) doesn't matter and are passed as defined
1.b) if value contains a linebreak: ❓undefined if it has to be quoted or (current assumption) one the newline has to be escaped
2. multiple `ENV <key>=<value>[ <key>=<value>]*` instructions behaves a bit different
2.a) simple case with a simple word value without linebreaks. Special Chars (quotes) have to be escaped only if the value is enwrapped in quotes due to spaces
[...]

PR #419, opened 2017-07-18
Issue #415 - multiple ENVs are not set the same way as like single ENV instructions

childnode:
Issue https://github.com/bmuschko/gradle-docker-plugin/issues/415

> When
> ```groov
> dockerfile(Type:Dockerfile) {
>     environmentVariable "foo" "bar"
> }
> ```
> 
> we get 
> ```Dockerfile
> ENV FOO bar
> ```
> 
> but with
> ```groov
> dockerfile(Type:Dockerfile) {
>     environmentVariable [ "foo":"bar" ]
> }
> ```
> 
> we get 
> ```Dockerfile
> ENV "FOO"="bar"
> ```

referring to https://docs.docker.com/engine/reference/builder/#env

- [x] for consistency with the SIngleItemJoiner the code should not always add these quotes

1. single `ENV <key> <value>`  instruction
- [x] 1.a) simple case with a simple word value without any linebreaks. Special Chars (quotes) doesn't matter and are passed as defined
- [ ] 1.b) if value contains a linebreak: ❓undefined if it has to be quoted or (current assumption) one the newline has to be escaped
2. multiple `ENV <key>=<value>[ <key>=<value>]*` instructions behaves a bit different
- [x] 2.a) simple case with a simple word value without linebreaks. Special Chars (quotes) have to be escaped only if the value is enwrapped in quotes due to spaces
- [x] 2.b) if value contains a linebreak or space: the value has to be quoted, linebreak is escaped / translated into a single \
- [x] 3. ❓undefined if keys are allowed to contain whitespaces, I assume not but quotes are allowed
  * as per [definition in Windows](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682653(v=vs.85).aspx):
    > The name of an environment variable cannot include an equal sign (=).
    …
    > The maximum size of a user-defined environment variable is 32,767 characters. There is no technical limitation on the size of the environment block.
    …
    > The environment block is an array of null-terminated Unicode strings. The list ends with two nulls (\0\0).
   * which matches the statements from the [open group](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html)(via [so](https://stackoverflow.com/a/2821183/529977))
     > … names shall not contain the character '='.
     > … the value shall be composed of characters from the portable character set (except NUL and as indicated below).
     > Environment variable names … consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in [Portable Character Set](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names.

- [x] 4. this applies to ENV and LABEL

some constraints: 
- [x] if user already quotes the string, we must not repeat this

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
@@ -116,8 +116,8 @@ RUN echo deb http://archive.ubuntu.com/ubuntu precise universe >> /etc/apt/sourc
 CMD ["echo", "some", "command"]
 EXPOSE 8080 14500
 ENV ENV_VAR_KEY envVarVal
-ENV "ENV_VAR_A"="val_a"
-ENV "ENV_VAR_B"="val_b" "ENV_VAR_C"="val_c"
+ENV ENV_VAR_A=val_a
+ENV ENV_VAR_B=val_b ENV_VAR_C=val_c
 ADD http://mirrors.jenkins-ci.org/war/1.563/jenkins.war /opt/jenkins.war
 COPY http://hsql.sourceforge.net/m2-repo/com/h2database/h2/1.4.184/h2-1.4.184.jar /opt/h2.jar
 ENTRYPOINT ["java", "-jar", "/opt/jenkins.war"]
@@ -125,7 +125,7 @@ VOLUME ["/jenkins", "/myApp"]
 USER root
 WORKDIR /tmp
 ONBUILD RUN echo "Hello World"
-LABEL "version"="1.0"
+LABEL version=1.0
 """
     }
 
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
@@ -43,11 +43,11 @@ class Dockerfile extends DefaultTask {
     }
 
     private void verifyValidInstructions() {
-        if (getInstructions().empty) {
+        if(getInstructions().empty) {
             throw new IllegalStateException('Please specify instructions for your Dockerfile')
         }
 
-        if (!(getInstructions()[0].keyword == 'FROM')) {
+        if(!(getInstructions()[0].keyword == 'FROM')) {
             throw new IllegalStateException('The first instruction of a Dockerfile has to be FROM')
         }
     }
@@ -75,9 +75,11 @@ class Dockerfile extends DefaultTask {
      * Example:
      *
      * <pre>
-     * task createDockerfile(type: Dockerfile) {*     instruction 'FROM ubuntu:14.04'
+     * task createDockerfile(type: Dockerfile) {
+     *     instruction 'FROM ubuntu:14.04'
      *     instruction 'MAINTAINER Benjamin Muschko "benjamin.muschko@gmail.com"'
-     *}* </pre>
+     * }
+     * </pre>
      *
      * @param instruction Instruction as String
      */
@@ -91,7 +93,11 @@ class Dockerfile extends DefaultTask {
      * Example:
      *
      * <pre>
-     * task createDockerfile(type: Dockerfile) {*     instruction { 'FROM ubuntu:14.04' }*     instruction { 'MAINTAINER Benjamin Muschko "benjamin.muschko@gmail.com"' }*}* </pre>
+     * task createDockerfile(type: Dockerfile) {
+     *     instruction { 'FROM ubuntu:14.04' }
+     *     instruction { 'MAINTAINER Benjamin Muschko "benjamin.muschko@gmail.com"' }
+     * }
+     * </pre>
      *
      * @param instruction Instruction as Closure
      */
@@ -404,7 +410,6 @@ class Dockerfile extends DefaultTask {
 
     static interface Instruction {
         String getKeyword()
-
         String build()
     }
 
@@ -421,9 +426,10 @@ class Dockerfile extends DefaultTask {
 
         @Override
         String getKeyword() {
-            if (instruction instanceof String) {
+            if(instruction instanceof String) {
                 parseKeyword(instruction)
-            } else if (instruction instanceof Closure) {
+            }
+            else if(instruction instanceof Closure) {
                 parseKeyword(instruction())
             }
         }
@@ -434,9 +440,10 @@ class Dockerfile extends DefaultTask {
 
         @Override
         String build() {
-            if (instruction instanceof String) {
-                instruction
-            } else if (instruction instanceof Closure) {
+            if(instruction instanceof String) {
+               instruction
+            }
+            else if(instruction instanceof Closure) {
                 instruction()
             }
         }
@@ -455,9 +462,10 @@ class Dockerfile extends DefaultTask {
 
         @Override
         String build() {
-            if (command instanceof String) {
+            if(command instanceof String) {
                 "$keyword $command"
-            } else if (command instanceof Closure) {
+            }
+            else if(command instanceof Closure) {
                 "$keyword ${command()}"
             }
         }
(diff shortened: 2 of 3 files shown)


childnode:
oh, I saw I accidentially changes some whitespaces ... fixed, pushed

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -501,20 +501,31 @@ class Dockerfile extends DefaultTask {
...
     }
 
     static class MultiItemJoiner implements ItemJoiner {
         @Override
-        String join(Map map) {
-            map.inject([]) { result, entry -> result << "\"$entry.key\"=\"$entry.value\"" }.join(' ')
+        String join(Map<String, String> map) {
+            map.inject([]) { result, entry ->
+                // TODO: clarify if assert should be used even in productive code and if this would be a valid one either
+                assert !entry.key.matches('.*(?: |(?:\r{0,1}\n)).*')
```
childnode: so, that's what has to be discussed as mentioned: Are keys valid even they contain whitespaces?

cdancy: Lets just take a step back here: what is an environment variable if it doesn't have some key? Is it even valid? What use would it serve to query for an empty string? `System.getenv("")`. I don't think we should allow empty keys.

childnode: mhh ... updated the task above... I think, we should not verify anything here, just handle what will lead to an invalid Dockerfile per definition.
If anyone adds "invalid" characters to a name or variable, this might be intended. As long the definition is correct and we don't produce invalid Dockerfiles, we should not care about because it's like a user writes stupid things into this file manually. Else, we have to be accurate enough to also verify if someone passes a pre-build Dockerfile for reuse ... "not my job" ;)

so, what I see:
* we care about whitespaces because they will change the
i.e. doing the same as with values, if any whitespace is included, enwrap it with quotes to be safe 
* we care about empty keys, because they will break the Dockerfile syntax too, because
`environmentVariable "" "bar"`
  => 
  > ENV bar
  <br>
  is invalid, resp. inverts what the user expects (value becomes the name, value is empty (in best case))

cdancy: Lets throw an exception here with an appropriate error message.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -501,20 +501,31 @@ class Dockerfile extends DefaultTask {
...
         }
     }
 
     static class SingleItemJoiner implements ItemJoiner {
         @Override
-        String join(Map map) {
-            map.inject([]) { result, entry -> result << "$entry.key $entry.value" }.join(' ')
+        String join(Map<String, String> map) {
+            map.inject([]) { result, entry ->
+                assert !entry.key.matches('.*(?: |(?:\r{0,1}\n)).*')
```
cdancy: Lets throw an exception here with an appropriate error message thereby letting the user know what exactly popped and why.

src/test/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileTest.groovy
```diff
@@ -7,47 +7,66 @@ import static com.bmuschko.gradle.docker.tasks.image.Dockerfile.*
...
+        new RunCommandInstruction({ 'apt-get update && apt-get clean' })                | 'RUN'           | 'RUN apt-get update && apt-get clean'
+        new DefaultCommandInstruction('ping google.com')                                | 'CMD'           | 'CMD ["ping google.com"]'
+        new DefaultCommandInstruction({ 'ping google.com' })                            | 'CMD'           | 'CMD ["ping google.com"]'
+        new ExposePortInstruction(8080)                                                 | 'EXPOSE'        | 'EXPOSE 8080'
+        new ExposePortInstruction({ 8080 })                                             | 'EXPOSE'        | 'EXPOSE 8080'
+        new ExposePortInstruction(8080, 9090)                                           | 'EXPOSE'        | 'EXPOSE 8080 9090'
+        new ExposePortInstruction({ [8080, 9090] })                                     | 'EXPOSE'        | 'EXPOSE 8080 9090'
+        new EnvironmentVariableInstruction('OS', 'Linux')                               | 'ENV'           | 'ENV OS Linux'
+        // TODO: negative test:
```
cdancy: Does this still need to be addressed?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
```diff
@@ -116,16 +116,16 @@ RUN echo deb http://archive.ubuntu.com/ubuntu precise universe >> /etc/apt/sourc
 CMD ["echo", "some", "command"]
 EXPOSE 8080 14500
 ENV ENV_VAR_KEY envVarVal
-ENV "ENV_VAR_A"="val_a"
-ENV "ENV_VAR_B"="val_b" "ENV_VAR_C"="val_c"
```
cdancy: Should we keep the old way as well since it's technically valid?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
cdancy: @childnode any thoughts here?

childnode: yes, sorry, was on vacation for some days .. As mentioned above, I would only like to escape / encaps whitespaces in quotes and would not like to throw any exception as long we only generate the file as accurate as possible, but are not responsible for validity in any system context, aren't we? So I would replace this assert by a replace as like for the value.

P.S: reading my own comments again .. we need to throw an exception if a blank key is given.

src/test/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileTest.groovy
childnode: you mean the negative test case? Yes, we should test the exceptional case I think ;)

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
childnode: yes, that's exactly what I prepared locally .. just pushin in that moment

childnode:
If you have another idea how to handle the exception cases in spock framework .. feel free to poke me ;) I have just tried to make it work without reading lot of spock documentation but failed to use some thrown() assertions

src/test/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileTest.groovy
```diff
@@ -2,52 +2,97 @@ package com.bmuschko.gradle.docker.tasks.image
...
+        catch(Exception e) {
+            LOG.info "exception catched with message `${e.message}`"
+            if(!expectedBuiltInstruction.equals(e.class)) {
+                LOG.log(Level.WARNING, e.stackTrace?.join('\n\t'))
+            }
+            actual = e.class
+        }
+
+        instructionInstance.keyword == expectedKeyword
+        actual == expectedBuiltInstruction
```
cdancy: What exactly is going on with this try/catch block?

I would also change the log to `Exception caught with message`

cdancy:
@childnode one comment to address and I think we should be good to go.

@orzeh second pair of eyes before we merge?

src/test/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileTest.groovy
childnode: 1. I had to catch the exception thrown in build. The expected type of exception is defined in `expectedBuiltInstruction`. If it doesn't match, the assertion fails like:
```
actual == expectedBuiltInstruction
|      |  |
|      |  ""
|      false
class java.lang.IllegalArgumentException
```
Without the catch, the exception would be thrown and will not produce a clean spock test result assertion

2. the LOG will show the stacktrace of the exception, if it's not the expected type to get some informations about what is going wrong .. indeed ... we might remove that

Last but not least: I'll change the message from Germish to English ;)

cdancy:
Thanks for the PR @childnode !

PR #422, opened 2017-07-23
Add basic network related tasks

jeysal:
Add tasks to create, inspect and remove docker networks.
Did not add the tasks to the README table yet, since the groovydoc would need to be published first.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
@@ -1,0 +1,22 @@
+package com.bmuschko.gradle.docker.tasks.network
+
+import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
+import org.gradle.api.tasks.Input
+
+class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
+    @Input
+    String networkId
+
+    DockerCreateNetwork() {
+        ext.getNetworkId = { networkId }
+    }
+
+    void runRemoteCommand(final dockerClient) {
+        logger.quiet "Creating network '$networkId'."
+        final network = dockerClient.createNetworkCmd().withName(networkId).exec()
+
+        if (onNext) {
+            onNext.call(network)
+        }
+    }
+}
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerInspectNetwork.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerInspectNetwork.groovy
@@ -1,0 +1,13 @@
+package com.bmuschko.gradle.docker.tasks.network
+
+class DockerInspectNetwork extends DockerExistingNetwork {
+    @Override
+    void runRemoteCommand(final dockerClient) {
+        logger.quiet "Inspecting network '${getNetworkId()}'."
+        final network = dockerClient.inspectNetworkCmd().withNetworkId(getNetworkId()).exec()
+
+        if (onNext) {
+            onNext.call(network)
+        }
+    }
+}
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerRemoveNetwork.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerRemoveNetwork.groovy
@@ -1,0 +1,10 @@
+package com.bmuschko.gradle.docker.tasks.network
+
+class DockerRemoveNetwork extends DockerExistingNetwork {
+    @Override
+    void runRemoteCommand(final dockerClient) {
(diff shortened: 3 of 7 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
```diff
@@ -0,0 +1,18 @@
...
+    @Input
+    String networkId
+
+    DockerCreateNetwork() {
+        ext.getNetworkId = { networkId }
+    }
+
+    void runRemoteCommand(final dockerClient) {
+        logger.quiet "Creating network '$networkId'."
+        dockerClient.createNetworkCmd().withName(networkId).exec()
```
cdancy: Is this a `void` call?

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerInspectNetwork.groovy
```diff
@@ -0,0 +1,10 @@
+package com.bmuschko.gradle.docker.tasks.network
+
+class DockerInspectNetwork extends DockerExistingNetwork {
+    @Override
+    void runRemoteCommand(final dockerClient) {
+        logger.quiet "Inspecting network '${getNetworkId()}'."
+        final network = dockerClient.inspectNetworkCmd().withNetworkId(getNetworkId()).exec()
+        onNext network
```
cdancy: Lets do something like:

```
        if (onNext) {
            onNext.call(network)
        }
```

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerRemoveNetwork.groovy
```diff
@@ -0,0 +1,10 @@
+package com.bmuschko.gradle.docker.tasks.network
+
+class DockerRemoveNetwork extends DockerExistingNetwork {
+    @Override
+    void runRemoteCommand(final dockerClient) {
+        final networkId = getNetworkId()
+        logger.quiet "Removing network '$networkId'."
+        dockerClient.removeNetworkCmd(networkId).exec()
```
cdancy: Is this a `void` call?

cdancy:
Hey @jeysal thanks for the handful of tasks! Few comments to address before merging. All in all code looks good though.

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
jeysal: Nope https://github.com/docker-java/docker-java/blob/f6d225d2c2e65aa058da5b94c2350cbe5d5f96c9/src/main/java/com/github/dockerjava/api/command/CreateNetworkResponse.java
I'll `onNext` the response.

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerRemoveNetwork.groovy
jeysal: Yep, this one is https://github.com/docker-java/docker-java/blob/f6d225d2c2e65aa058da5b94c2350cbe5d5f96c9/src/main/java/com/github/dockerjava/api/command/RemoveNetworkCmd.java#L14

jeysal:
Thanks for reviewing, I've made the requested changes 👍
Let's see what CI says, should not have broken anything though.

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerRemoveNetwork.groovy
cdancy: hmmm ... I wonder if we should start here with setting a standard. If the user were to define `onNext` here, but we have nothing to hand off, than it's more/less a no-op. I wonder if we should pass in say `true` just to give the user some feedback. @orzeh any thoughts?

jeysal: imo, that wouldn't be clean because `onNext` on Observables etc should usually carry a meaningful value.
You can already get success feeback from `onComplete`, right?

cdancy: That had slipped my mind and you're correct. Ok lets take it as is then.

cdancy:
@jeysal would you mind adding these tasks to the README.md file as well? I understand the groovydocs are not yet published but it should be easy enough to make a guess on the links for when they are published. Once this gets in, and I take care of the other currently in-flight PR, I plan on releasing. Probably before end of week at the latest.

jeysal:
Done 👍

cdancy:
Merged now. Thanks for the set of network tasks @jeysal !

Issue #357, opened 2017-03-21
Allow Dockerfile task to be UP-TO-DATE


ayedo:
First of all: thanks for your awesome work!

I was wondering: why does the Dockerfile task set the following in the constructor:

`   Dockerfile() {
        outputs.upToDateWhen { false }
}`

I'm using the plugin to start a postgres database during the build in order to generate a jOOQ meta model. Incremental building does not work since in the end, the build chain depends on the Dockerfile task, which is never upToDate.

I tried overwriting the directive in my task, but without success.

orzeh:
@ayedo I think this is worth considering. I can imagine use cases when generated `Dockerfile` (task output) does not change for subsequent builds. We could remove the `outputs.upToDateWhen { false }` statement and allow developers to define task inputs explicit using [standard Gradle syntax](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks). 

@cdancy WDYT?

@ayedo are you interested in submitting PR?

cdancy:
@orzeh I agree. Think this was probably left in from when we attempted a go at making docker image reproduction "UP-TO-DATE".

remmeier:
would be very helpful if this is fixed

cdancy:
@remmeier submit a PR?

cdancy:
Removed in master. Next release will have change.

childnode:
currently we suspect this change to cause images are not rebuild properly with the application plugin (`com.bmuschko.docker-java-application`) since a dockerDistTar anyway doesn't respect the distTar state and therefor this change in Dockerfile leads to deprecated Dockerfile even the distTar has changed

kaserf:
Hi, I ran into a problem recently where I would change my "Dockerfile" task (specifically adding some environment variables and changing the entry point), however when I rebuild a task that depends on the "Dockerfile" task, the build task thinks that the "Dockerfile" task is up to date.

Could it be that this is a regression that was introduced by the fix for this issue? It sounds a bit like what @childnode is experiencing as well.

cdancy:
@kaserf what version of the plugin are you using?

kaserf:
@cdancy 3.0.9

cdancy:
@kaserf that's why. Try upgrading to `3.0.12` and let me know how things workout. Fix for this was merged within the past couple releases.

kaserf:
Just tried with `3.0.12`, same problem.

kaserf:
[...]

PR #433, opened 2017-08-08
Fix Dockerfile task up-to-date logic

orzeh:
Should fix issues reported in #357

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
@@ -28,7 +28,7 @@ class Dockerfile extends DefaultTask {
     File destFile = project.file("$project.buildDir/docker/Dockerfile")
 
     Dockerfile() {
-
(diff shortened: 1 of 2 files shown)


cdancy:
Let me see if I understand the logic here correctly: we are collecting all the instructions as an input property and so if this should change further downstream, meaning the instructions, this property will become out-of-date and re-trigger the build?

orzeh:
@cdancy yes. See the test case.

cdancy:
LGTM. Merging now.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -28,7 +28,7 @@ class Dockerfile extends DefaultTask {
     File destFile = project.file("$project.buildDir/docker/Dockerfile")
 
     Dockerfile() {
-
+        inputs.property("instructionsCache", { instructions.collect { it.build() }.join('\n') })
```
childnode: time passes by ... so my comment https://github.com/bmuschko/gradle-docker-plugin/issues/357#issuecomment-321316089 is now outdated, but perhaps still valid?!

PR #434, opened 2017-08-09
Some test additions to Issue 357+433

childnode:
addressing discussion in #357 and fix #433 

added some more regression tests
- [x] open question: is `dockerFile() { outputs.upToDateWhen { moon.inZenith } }` a valid use case now the base issue is fixed by #433 ?

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
@@ -1,7 +1,6 @@
 package com.bmuschko.gradle.docker
 
 import org.gradle.testkit.runner.BuildResult
-import org.gradle.testkit.runner.TaskOutcome
 import spock.lang.Requires
 
 class DockerJavaApplicationPluginFunctionalTest extends AbstractFunctionalTest {
@@ -27,68 +26,6 @@ ENTRYPOINT ["/${projectName}-1.0/bin/${projectName}"]
 EXPOSE 8080
 """
         result.output.contains('Author           : ')
-    }
-
-    def "Can build image for Java application with default configuration and changes on changed application package"() {
-        String projectName = temporaryFolder.root.name
-        createJettyMainClass()
-        writeBasicSetupToBuildFile()
-        writeCustomTasksToBuildFile()
-
-        when:
-        BuildResult result = build('dockerBuildImage')
-
-        then:
-        File dockerfile = new File(projectDir, 'build/docker/Dockerfile')
-        File distTarFile = new File(projectDir, "build/distributions/${projectName}-1.0.tar")
-        distTarFile.exists()
-        long distTarFileLastTouched = distTarFile.lastModified()
-        dockerfile.exists()
-        dockerfile.text ==
-"""FROM java
-MAINTAINER ${System.getProperty('user.name')}
-ADD ${projectName}-1.0.tar /
-ENTRYPOINT ["/${projectName}-1.0/bin/${projectName}"]
-EXPOSE 8080
-"""
-
-        when:
-        result = build('dockerBuildImage')
-
-        then:
-        TaskOutcome.UP_TO_DATE == result.tasks.find{it.path == ":dockerBuildImage"}.outcome
-
-        when:
-        temporaryFolder.newFile('file1.txt')
-        temporaryFolder.newFile('file2.txt')
-        buildFile << """
-        distTar {
-            from '${temporaryFolder}/file1.txt'
-            from '${temporaryFolder}/file2.txt'
-        }
-        """
-        result = build('dockerBuildImage')
-
-        then:
-        result.tasks.each { println "$it.path $it.outcome" }
-
-        distTarFileLastTouched < distTarFile.lastModified()
-        TaskOutcome.SUCCESS == result.tasks.find{it.path == ":dockerBuildImage"}.outcome
-
-        when:
-        buildFile << "version = 1.1"
-        build('dockerBuildImage')
-
-        then:
-        dockerfile.exists()
-        dockerfile.text ==
-"""FROM java
-MAINTAINER ${System.getProperty('user.name')}
-ADD ${projectName}-1.1.tar /
-ENTRYPOINT ["/${projectName}-1.1/bin/${projectName}"]
-EXPOSE 8080
-"""
-
     }
 
     def "Can create image for Java application with user-driven configuration"() {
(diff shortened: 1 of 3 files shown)


cdancy:
I think supporting this is still a requirement as some users may want this functionality and we should obey the gradle contract. @childnode are we thinking the current changes by @orzeh don't allow for this?

childnode:
> @childnode are we thinking the current changes by @orzeh don't allow for this?

that's what my test tells me ... waiting for travis testing 
https://github.com/bmuschko/gradle-docker-plugin/pull/434/commits/0601fdf6115c57c7c83c3ac4eb6797f79dc3aa68

cdancy:
If that's the case then we should be able to do something within the `Dockerfile` task to detect if the user has set this or not and proceed accordingly. Care to add that functionality to this PR?

orzeh:
@childnode great work! Thanks for that! Support for custom _UP-TO-DATE_ logic like in the failing test can be added in two ways:
* _built in_ logic should be opt-in, i.e. user must explicitly set some boolean property which in setter will register task inputs
* we have to add build lifecycle listener (`project.afterEvaluate` ?) and check if task has `outputs.upToDateWhen { }` defined. If not, register task input property.

For me 2nd one is better. WDYT? Or maybe you have better ideas?

cdancy:
Option #2 seems to be the logical choice.

childnode:
you're welcome. option 2 sounds good for me too, but I have no glue how to implement this ... that's too much gradle interna for me atm ;)

cdancy:
@childnode @orzeh think I've got something that would work. Give this a go within `Dockerfile` and let me know how it works out:

```
    Dockerfile() {

        // if user has defined their own `upToDate` spec then
        // skip implementing ours.
        if (outputs.getUpToDateSpec().getSpecs().empty) {
            inputs.property("instructionsCache", { instructions.collect { it.build() }.join('\n') })
        }
    }
```

orzeh:
@cdancy how would `outputs` property already be initialized in constructor?

cdancy:
@orzeh I know but it worked when I gave it a go here. I'm not sure how gradle is doing things behind the scenes but when I set an `outputs.upToDateWhen{ false }` on a task that check was false. When I removed it the check was true.

cdancy:
@childnode have another look at this if you can given the above workaround I suggested. It _may_ work. If not I think we can release without this and simply put a disclaimer somewhere for the time being.

childnode:
mhh ... I can't see that the (proposed patch)[#issuecomment-322183151] will change anything for me locally ;/ ... if it works for you, I can just rebase and drop in that patch and let's see what travis says ,)

If not ... yeah, that will be somewhat a regression of Issue #357

childnode:
what was to be proved:
❌ https://github.com/bmuschko/gradle-docker-plugin/pull/434/commits/9095417da1322d7c029bb25cf9a9244598a0a0eb

so, since there are no other ideas ... we have to fall back to your "disclaimer notice" ;)

I even don't see a chance to get this since it seems the spec like `upToDateWhen { true }` transforms the `DefaultTaskOutputs` Spec into a Composite / `AndSpec` with the user given Spec / Closure (see `gradle-api-4.0.jar!/org/gradle/api/internal/tasks/DefaultTaskOutputs.class:62` ff)
I even wonder sometimes why your code works ... it's groovy I know. Why? `Spec<? super TaskInternal> getUpToDateSpec()` doesn't have any interface like `.getSpecs()` unless is it implemented via  `DefaultTaskOutputs.class` with an `org/gradle/api/specs/CompositeSpec.class:59`
This way, I found there would be a shortcut for your patch:
```
-         if (getUpToDateSpec().getSpecs().empty) {
+         if (getUpToDateSpec().isEmpty()) {
```
see `gradle-api-4.0.jar!/org/gradle/api/specs/CompositeSpec.class:63`

Last but not least: along with `gradle-api-4.0.jar!/org/gradle/api/internal/tasks/DefaultTaskOutputs.class:130` 
> `    public void cacheIf(Spec<? super Task> spec) {…}`

I assume there are more cases than `upToDateWhen` to be respected ...

my second but last thought here ... if `upToDateWhen {}` is implemented per default as an logical and of any given closures/specs ... can't we "just" add ours (somehow) the same way?

Else: is this a core bug in gradle not respecting the upToDateWhen spec when input properties are defined?!

childnode:
I moreover assume .. I'm a stupid moron ;)
I should have written a correct test .. [after reading the docs](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/TaskOutputs.html#upToDateWhen(groovy.lang.Closure))
> You can add multiple such predicates. The task outputs are considered out-of-date when any predicate returns false.

1. we don't need to change anything in Dockerfile ... it works as expected
2. you never can override the upToDate by just returning true, this code will have no effect
2. I fixed the tests, rebased it ... If you like, you might merge them

cdancy:
@childnode I assume this latest test failure is unexpected?

src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
```diff
@@ -28,6 +29,68 @@ EXPOSE 8080
...
+ADD ${projectName}-1.0.tar /
+ENTRYPOINT ["/${projectName}-1.0/bin/${projectName}"]
+EXPOSE 8080
+"""
+
+        when:
+        result = build('dockerBuildImage')
+
+        then:
+        TaskOutcome.UP_TO_DATE == result.tasks.find{it.path == ":dockerBuildImage"}.outcome
```
childnode: This fails in https://travis-ci.org/bmuschko/gradle-docker-plugin/jobs/265243034#L736
```
TaskOutcome.UP_TO_DATE == result.tasks.find{it.path == ":dockerBuildImage"}.outcome
                       |  |      |     |                                    |
                       |  |      |     :dockerBuildImage=SUCCESS            SUCCESS
                       |  |      [:compileJava=UP_TO_DATE, :processResources=NO_SOURCE, :classes=UP_TO_DATE, :jar=UP_TO_DATE, :startScripts=UP_TO_DATE, :distTar=UP_TO_DATE, :dockerCopyDistResources=UP_TO_DATE, :dockerDistTar=UP_TO_DATE, :dockerBuildImage=SUCCESS]
                       |  org.gradle.testkit.runner.internal.FeatureCheckBuildResult@68c9d179
                       false
```

childnode:
it was someway expected that there is something wrong when I wrote it ... *dough* yeah, it moved out of the scope. This addresses what I mentioned in 357: when a given file that is added to the image has changed, the input Dockerfile was not changed but the image has to.
- [x] This is now fixed half-way in the case when the filename changes.
- [ ] Perhaps (that's what I tried to test) it is (not) fixed in the case, the filename keeps the same, but the file added changed. 
- [ ] I also assumed when writing the test, that a rebuild (dockerBuildImage) secondly becomes UP-TO-DATE, because Dockerfile is not re-build neither the file has changed.
That is - the test failure as far as I can reproduce - not what it works like actually: https://github.com/bmuschko/gradle-docker-plugin/pull/434/commits/2419283eeee2c53ff1dbc7ed102ca8097b7a2533#r133552747
The second `dockerBuildImage` becomes successfull even when nothing changed... for unknown reasons

cdancy:
Ok lets take these as-is and we can follow up with a point release if need be once `3.1.0` is live.

PR #435, opened 2017-08-09
The first instruction of a Dockerfile has to be FROM except for Docker later than 17.05

childnode:
https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy#L50

is no longer valid since [docker 17.05+](https://github.com/moby/moby/releases/tag/v17.05.0-ce)
> Allow using build-time args (ARG) in FROM https://github.com/docker/docker/pull/31352

So: while 
```
task dockerFile(type: Dockerfile) {
    arg 'from=alpine'
    from '\$from'
}
```
is not valid until 17.05, later on it works.

I just hacked some ugly code ... mainly rebasing `Dockerfile` on `AbstractDockerRemoteApiTask`, implementing `runRemoteCommand()`, 🤞 cross the fingers that this `@TaskAction` runs before `Dockerfile.create()`, parsing the version in ugly 3 lines 

Feel free to blame ... but perhaps someone could get me on track how to get the version correctly or does have another idea how we should handle this issue.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
@@ -49,13 +49,40 @@ import com.bmuschko.gradle.docker.tasks.image.Dockerfile
 task ${DOCKERFILE_TASK_NAME}(type: Dockerfile) {
     maintainer 'Benjamin Muschko "benjamin.muschko@gmail.com"'
 }
+
+task ${DOCKERFILE_TASK_NAME}simple(type: Dockerfile) {
+    from '$TEST_IMAGE_WITH_TAG'
+}
+
+task ${DOCKERFILE_TASK_NAME}1705(type: Dockerfile) {
+    arg 'from=$TEST_IMAGE_WITH_TAG'
+    from '\$from'
+}
 """
+        def outcome = new File(projectDir, 'build/docker/Dockerfile')
+
         when:
         BuildResult buildResult = buildAndFail(DOCKERFILE_TASK_NAME)
 
         then:
-        buildResult.output.contains('The first instruction of a Dockerfile has to be FROM')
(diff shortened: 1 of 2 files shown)


orzeh:
@childnode IMHO in that case it would be better to just remove the validation from plugin and let Docker daemon do the job for us. @cdancy ?

cdancy:
@orzeh agreed. @childnode feel free to rip that out now while you're poking around. This was a recent check put in place to stop things early but the build will ultimately fail for users on older versions anyway so it's not entirely necessary.

childnode:
should I rebase / fixup the branch?

cdancy:
@childnode yes please do a squash/rebase of your commits.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
```diff
@@ -49,13 +49,40 @@ import com.bmuschko.gradle.docker.tasks.image.Dockerfile
...
+        outcome.exists()
+        def lastOutModified = outcome.lastModified()
+
+        when:
+        buildResult = build("${DOCKERFILE_TASK_NAME}simple")
+
+        then:
+        TaskOutcome.SUCCESS == buildResult.tasks.first().outcome
+        outcome.exists()
+        lastOutModified < outcome.lastModified()
```
cdancy: Looks like this line is failing in CI?

childnode:
ok ... I'll try to reproduce this again locally ... perhaps I missed something to update before running the tests ...

cdancy:
@childnode any chance we can clean this up sooner rather than later? I'd like to get a new release out ASAP.

childnode:
ahh, I see .. I assume this will be fixed by #433 .. I'm going to run some tests

cdancy:
> But unfortunatly .lastModified() is not reset when file has been changed the second time

@childnode yeah that's odd. In either event this should suffice for now. Thanks again for the PR!

Issue #436, opened 2017-08-11
How to attach a container to a docker network.


edesdan:
Hi I'm wondering how to attach a container to a specific docker network. I'm don't see any option in the DockerCreateContainerTask. Also tried networkAliases but no success.

cdancy:
@edesdan while I have yet to personally have a use-case for this it should work with `networkAliases` if I"m not mistaken. A developer just recently contributed the various network tasks we have but they are only for create, delete, inspect. Though if a network exists I would think `networkAliases` would do the trick. Can you post an example of what you're trying to do and snippet of the stacktrace (assuming there is one)?

cdancy:
@edesdan actually going over the `docker-java` project, and looking at this [ISSUE](https://github.com/docker-java/docker-java/issues/601), it appears that creating a container with `networkMode=<network-name>` is the proper way to go. Give that a go and let me know how things work out.

edesdan:
@cdancy thanks for the advice/help. It works setting the 'networkMode' to the name of the docker network you create. I got confused by the name of the variable, it should be called just 'network', networkMode is a little bit confusing I guess.  Thanks!

orzeh:
@edesdan feel free to submit PR with either some documentation or changed variable name. IMO `networkMode` is confusing.

edesdan:
@orzeh yeah I'll try to submit a PR with a name variable change if there are no objections/contraindications . Do we have any specific guideline to submit a PR?

cdancy:
@edesdan only guidelines is that all tests pass. I actually dont think we have a test wired for this use-case so adding one would be great.

As for new name... I'm up to suggestions. Maybe just `networkName`?

edesdan:
ok I'm forking the project, I will prepare a PR for you to review. I think just network should be fine. From the official docker documentation it is just 'network' 

docker run --network=<NETWORK> . If it is easy for me to understand how to add a test I will do that for this scenario.

cdancy:
Agreed.

edesdan:
Just to keep track and to have another opinion on that I opened this issue here: https://github.com/docker-java/docker-java/issues/899

mursilsayed:
Passing one network works fine for me. I have a use case in which I want to pass multiple docker networks to a container.

I can do that with docker cli by specifying --network network1 --network network2
 How can I pass multiple networks while creating docker containers via the plugin?

cdancy:
[...]

PR #440, opened 2017-08-15
change networkMode to network (issue #436)

edesdan:
@cdancy @orzeh As discussed the PR for #436. I haven't added any test because there was a test already to cover that scenario.


cdancy:
@edesdan thanks for the PR!

edesdan:
@cdancy no prob, I 'll try to contribute more to the project, I like it and it is very useful. I think the codebase could be improved... Thanks!

cdancy:
>  I think the codebase could be improved.

It certainly can and anywhere you see a shortcoming of any kind don't be afraid to open an ISSUE and call it out. We're all just trying to get to the best most maintainable code for all to use.

Issue #383, opened 2017-05-30
Ensure all tests clean up after themselves


cdancy:
After running integration tests we have loads of images and containers hanging around. We shouldn't be doing this and really need to set a standard either for cleaning up as we go or to have some sort of finally task that executes after integration tests have finished running and clean up then.

edesdan:
I think this is very annoying. I had a quick look and maybe we can do something like this:
inside the AbstractFunctionalTest class we can store all created containers/images names and then we can clean up after tests. We need a way to use a docker client to remove all the containers based on container names. Something like this:
```

package com.bmuschko.gradle.docker

import org.gradle.testkit.runner.BuildResult
import org.gradle.testkit.runner.GradleRunner
import org.junit.Rule
import org.junit.rules.TemporaryFolder
import spock.lang.Specification

abstract class AbstractFunctionalTest extends Specification {
    @Rule
    TemporaryFolder temporaryFolder = new TemporaryFolder()

    File projectDir
    File buildFile

    static final String TEST_IMAGE = "alpine"
    static final String TEST_IMAGE_TAG = "3.4"
    static final String TEST_IMAGE_WITH_TAG = "${TEST_IMAGE}:${TEST_IMAGE_TAG}"

    String dockerServerUrl

    def containers = []

    def setup() {
        projectDir = temporaryFolder.root
        buildFile = temporaryFolder.newFile('build.gradle')

        buildFile << """
            plugins {
                id 'com.bmuschko.docker-remote-api'
            }
			
            repositories {
                mavenCentral()
            }
        """

        setupDockerServerUrl()
        setupDockerCertPath()
        setupDockerPrivateRegistryUrl()

        buildFile << """
            task dockerVersion(type: com.bmuschko.gradle.docker.tasks.DockerVersion)
        """

        when:
        BuildResult result = build('dockerVersion')

        then:
        result.output.contains('Retrieving Docker version.')
    }

    def cleanupSpec() {
        println("Cleanup after all tests!")
        // remove all used containers (containers list)
    }

    private void setupDockerServerUrl() {
        dockerServerUrl = TestConfiguration.dockerHost

        if (dockerServerUrl) {
            buildFile << """
                docker.url = '$dockerServerUrl'
            """
        }
    }

    private void setupDockerCertPath() {
        File dockerCertPath = TestConfiguration.dockerCertPath

        if (dockerCertPath) {
            buildFile << """
[...]

PR #442, opened 2017-08-18
Ensure all tests clean up after themselves #383

edesdan:
Hi Guys, I did a little bit of work for the #383 issue. So basically now after you run the functional tests all the created docker images and docker containers are automatically removed. I also added two new methods to start/remove the docker registry container that was causing 3 tests to fail. Let me know what you think.

### DIFF
--- a/gradle/functional-test.gradle
+++ b/gradle/functional-test.gradle
@@ -30,6 +30,7 @@ check.dependsOn functionalTest
 
 dependencies {
     functionalTestCompile 'commons-lang:commons-lang:2.6'
+    functionalTestCompile 'com.github.docker-java:docker-java:3.0.12'
 }
 
 gradlePlugin {
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -24,6 +24,7 @@ import com.github.dockerjava.core.DefaultDockerClientConfig
 import com.github.dockerjava.core.DockerClientBuilder
 import com.github.dockerjava.core.DockerClientConfig
 import com.github.dockerjava.core.command.PullImageResultCallback
+import groovy.util.logging.Slf4j
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.junit.Rule
@@ -31,6 +32,7 @@ import org.junit.rules.TemporaryFolder
 import spock.lang.Shared
 import spock.lang.Specification
 
+@Slf4j
 abstract class AbstractFunctionalTest extends Specification {
 
     @Rule
@@ -47,7 +49,7 @@ abstract class AbstractFunctionalTest extends Specification {
 
     @Shared
     DockerClientConfig config = DefaultDockerClientConfig.createDefaultConfigBuilder()
-        .withDockerHost("unix:///var/run/docker.sock")
+        .withDockerHost(TestConfiguration.dockerHost)
         .build()
 
     @Shared
@@ -102,29 +104,32 @@ abstract class AbstractFunctionalTest extends Specification {
 
         def dockerContainersAfterTests = dockerClient.listContainersCmd().withShowAll(true).exec() as Set
 
-        def removableDockerContainerIds = dockerContainersAfterTests.collect { it.id } - dockerContainersBeforeTests.collect { it.id }
+        def removableDockerContainerIds = dockerContainersAfterTests.collect {
+            it.id
+        } - dockerContainersBeforeTests.collect { it.id }
 
-        println("Removing ${removableDockerContainerIds.size()} containers: ${removableDockerContainerIds}")
+        log.info("Removing ${removableDockerContainerIds.size()} containers: ${removableDockerContainerIds}")
         removableDockerContainerIds.each {
             try {
                 dockerClient.removeContainerCmd(it)
                     .withForce(true)
                     .exec()
             } catch (NotFoundException ex) {
-                println("Exception while removing container: ${ex.message} ")
+                log.warn("Exception while removing container: ${ex.message} ")
             }
         }
-
-        println("Docker containers removed.")
+        log.info("Docker containers removed.")
     }
 
     private void removeDockerImages() {
 
         def dockerImagesCreatedAfterTests = dockerClient.listImagesCmd().exec() as Set
 
-        def removableDockerImageIds = dockerImagesCreatedAfterTests.collect { it.id }  - dockerImagesBeforeTests.collect { it.id }
+        def removableDockerImageIds = dockerImagesCreatedAfterTests.collect {
+            it.id
+        } - dockerImagesBeforeTests.collect { it.id }
 
-        println("Removing ${removableDockerImageIds.size()} images: ${removableDockerImageIds}")
+        log.info("Removing ${removableDockerImageIds.size()} images: ${removableDockerImageIds}")
 
         removableDockerImageIds.each {
             try {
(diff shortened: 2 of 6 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -34,6 +45,20 @@ abstract class AbstractFunctionalTest extends Specification {
 
     String dockerServerUrl
 
+    @Shared
+    DockerClientConfig config = DefaultDockerClientConfig.createDefaultConfigBuilder()
+        .withDockerHost("unix:///var/run/docker.sock")
```
orzeh: Now tests won't run on Windows. It would be better to use `TestConfiguration` here.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -63,6 +88,105 @@ abstract class AbstractFunctionalTest extends Specification {
...
+                    .exec()
+            } catch (NotFoundException ex) {
+                println("Exception while removing image: ${ex.message} ")
+            }
+        }
+
+        println("Docker images removed.")
+    }
+
+    protected void startDockerRegistryContainer() {
```
orzeh: I would change it to `startDockerRegistryIfNeeded`

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -63,6 +88,105 @@ abstract class AbstractFunctionalTest extends Specification {
...
+
+        def containers = dockerClient.listContainersCmd().exec()
+        println("Active dockerContainersBeforeTests: ${containers}")
```
orzeh: This prints lots of junk in console, maybe use only container ids?

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -63,6 +88,105 @@ abstract class AbstractFunctionalTest extends Specification {
...
+
+        boolean isRegistryAlreadyRunning
+
+        isRegistryAlreadyRunning = containers.find {
+            return it?.names?.contains("/registry")
```
orzeh: `return` can be omitted.

orzeh: in most cases this should work, however if I start registry container with custom name it will fail. But for now it's :ok_hand:

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
cdancy: This is why I suggested to use the AbstractDockerTask and grab the client from there. It removes the need to do this as well as define a docker-java version in yet another place.

edesdan: yeah, I agree it is not the best, I'll will review this part.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
edesdan: ehm not sure I'll think about it...

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
edesdan: sure, I will remove the prints

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
edesdan: Yeah you are right but for now I will go with that..

gradle/functional-test.gradle
```diff
@@ -30,6 +30,7 @@ check.dependsOn functionalTest
 
 dependencies {
     functionalTestCompile 'commons-lang:commons-lang:2.6'
+    functionalTestCompile 'com.github.docker-java:docker-java:3.0.12'
```
cdancy: As I noted below: if we use the AbstractDockerTask we don't need to rely upon bringing this onto the `functionalTestCompile` classpath.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -63,6 +88,105 @@ abstract class AbstractFunctionalTest extends Specification {
...
+                .withForce(true)
+                .exec()
+        } catch (NotFoundException ex) {
+            println("Exception while removing container: ${ex.message} ")
+        }
+    }
+
+
+
+
```
cdancy: Lots of extra newlines added. Lets remove these.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -115,7 +239,7 @@ abstract class AbstractFunctionalTest extends Specification {
         "gradle/${generateRandomUUID()}"
     }
 
-    protected String createUniqueContainerName() {
+    protected final String createUniqueContainerName() {
```
cdancy: Why this change?

cdancy:
My only complaint is that I think we may be over thinking this. From my POV I don't think we need to worry about which images/containers were around before the start of testing so long as we have a standard naming convention (which we do to some extent and I'm open to change it) and can simply iterate, once testing has completed, through all of the remaining images and containers and remove those that match our naming convention.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
cdancy: If we do end up keeping this piece of code I would change it to `logger.info` as I still think it's useful just not something we would care to see for every run.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
edesdan: I guess it got in the mix during my tests... I will remove it.

edesdan:
@cdancy I see your point but I think this is no over thinking. I think it is more safe doing it in this way and 'self' maintainable (basically you don't need to rely on any specific logic for naming containers and images). I tried going in the other direction first and I saw the drawbacks. Anyway that's my POV. I'll try to update the PR ASAP. Thanks.

cdancy:
@edesdan fair enough. Feel free to continue going down your road I would just make sure to use the abstract class so that we don't have to include another dependency.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
edesdan: @cdancy can you point me to the code where I can grab the client? I cannot find any AbstractDockerTask class. Thanks.

edesdan: @orzeh I see how I can use TestConfiguration class to get the docker config. I guess @cdancy suggestion is to avoid bringing the client dependency but I'm still not finding any reference where to see how to grab a docker-client.

edesdan:
I did another commit to address the minor comments. Regarding no adding a new dependency I need help from you guys, as I sad in the previous comments I don't know how to address the issue. If you just can past some code snippet I'm happy to do the change. Thanks.

cdancy:
@edesdan [have a look here](https://github.com/bmuschko/gradle-docker-plugin/blob/84939895a69e48c2f277d56868be855a67a9c994/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractDockerRemoteApiTaskFunctionalTest.groovy#L21). It would be much easier to implement a custom "cleanup task" of some sort and just purge them all this way IMO.

edesdan:
@cdancy What you mean by 'much easier'? I'm not getting your point. I had a look at the link you sent me but still I'm not sure what I should do. If it is not a big issue for you I will take this work I have done and merge the PR : it works good to me and I think I have addresses all minor issues raised after the first commit. Feel free to merge it and change it, or feel free to reject it. I don't think I can spend more time on this, sorry about that.

Anyway I think this should be a priority, it could stop people to collaborate to the project. Honestly when I first executed the tests I had a look at my docker images/containers it was a mess: waste of space and memory.

cdancy:
@edesdan I'll try to put something together this week but the main idea is:

If you use a custom `AbstractDockerRemoteApiTask` you don't have to add a functional test dependency on `docker-java` which is yet another thing we will have to maintain. Instead you get passed an instance of `dockerClient` to do whatever you want with:

```
            // using pseudo-code here to get the point across
            class CleanupTask extends AbstractDockerRemoteApiTask {
                @Override
                    def listOfGeneratedImage = dockerClient.getListOfImagesSomehow()
                    dockerClient.removeImages(listOfGeneratedImage)

                    def listOfGeneratedContainers = dockerClient.getListOfContainersSomehow()
                    dockerClient.removeContainers(listOfGeneratedContainers)
                }
            }
```

cdancy:
Closing for now as this PR has been open almost 2 years and I think there are more complete solutions we can take.

Issue #351, opened 2017-02-27
Conditional building of image


tdudgeon:
How is it best to configure a DockerBuildImage to only do the build when necessary?
For instance, let's assume we have a Java web application build built with the gradle war plugin.
If I call the war target the war file only gets rebuilt if it's necessary to do so.
But if I use the gradle-docker-plugin to build a docker image that incorporates that war file it rebuilds the docker image every time, even if the war has not been updated.
The scenario here could be a multi-project build when you are running integration tests using docker images from some of the sub-projects, and you only want to rebuild when something has changed in  that sub-project.
In simplest terms this means make the DockerBuildImage task run only when the output of the war task is newer than the docker image that was last built.

cdancy:
@tdudgeon this issue has come up numerous times, and at one point even tried to implement a solution, but upon further investigation it was deemed impossible to tell whether an image was different from some other image. Looking at things from our point of view, meaning not taking into account your war file, how would we be able to tell if this image being built would be newer than some previously existing image? We're looking at things from the docker POV and there really is no sane way to determine

You could probably implement something on your end (i.e. maybe using an `onlyIf` block on this particular task) if you're sure the image is different than some other already built image.

PR #443, opened 2017-08-18
:dockerBuildImage is suspiciously building too often

childnode:
By testing some cases regarding the UP-TO-DATE behaviour (#357 -> #433, #434), I stumbled over a case that is possibly a bug to be discussed.

To prove that, I began to write some test 8fab023 .. in the meanwhile there is a second commit with a suggested improvment patch for the `DockerJavaApplicationPlugin`

Details:
- [ ] it is expected, that a `DockerBuildImage` only runs twice if any input has changed. i.e. either the Dockerfile itself or any file affected.
  - (?) this has to be discussed, if this is either resolvable - since we allow to pass an existing Dockerfile - or it is "correct" because `docker build` not even respects that too. So it might be ok that DockerBuildImage rebuilds anytime?! The UP-TO-DATE assumption only comes from the traditional gradle build behaviour not doing things twice. 
- regarding the `DockerJavaApplicationPlugin`
  - [x] it is expected, that any incoming change from the `:distTar` (type `Tar` from gradle application plugin) affects `:dockerCopyDistResources` (type `Copy`) and therefor `:dockerDistTar` (type `Dockerfile`)
    - while the UP-TO-DATE logic of `:dockerDistTar` seems to be correct with the changes from #433 it now (again) respects if the filename of the distTar changes.
`:dockerCopyDistResources` works out of the box due to gradle `Copy` from instructions implies any changes to `:distTar` affects this task too.
  - it is expected, that no changes made to the Dockerfile (or input in general) doesn't trigger `:dockerBuildImage` -> see first point in general.

The possible improvement I see:
- [ ] The `DockerBuildImage` by DockerJavaApplicationPlugin `:dockerBuildImage` might get an [`onlyIf {}`](https://docs.gradle.org/current/dsl/org.gradle.api.Task.html?#org.gradle.api.Task:onlyIf(groovy.lang.Closure)) spec / directive to regard the status of the `DockerFile` task `:dockerDistTar`

=> see patch 570b3ef

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -15,11 +15,15 @@
  */
 package com.bmuschko.gradle.docker
 
+import org.gradle.api.GradleException
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 import org.junit.Rule
 import org.junit.rules.TemporaryFolder
 import spock.lang.Specification
+
+import java.nio.file.Files
+import java.nio.file.Paths
 
 abstract class AbstractFunctionalTest extends Specification {
     @Rule
@@ -51,6 +55,7 @@ abstract class AbstractFunctionalTest extends Specification {
         setupDockerServerUrl()
         setupDockerCertPath()
         setupDockerPrivateRegistryUrl()
+        setupDockerTemplateFile()
 
         buildFile << """
             task dockerVersion(type: com.bmuschko.gradle.docker.tasks.DockerVersion)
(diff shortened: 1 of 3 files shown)


orzeh:
@childnode It was discussed a while ago in several places (like in #351). TLDR: we have to always execute DockerBuildImage because we can't say if the resulting image will change.

childnode:
yeah .. that's clear for the `DockerBuildImage` Task type we can't write anything valid.
That's why I only suggested a patch for the DockerJavaApplicationPlugin defined task `:dockerBuildImage` ... with the following rules and assumptions made_:
```ockerBuildImage by implementing onlyIf {}

assumptions
- there are no other ADD instructions than the one set by the plugin
- there are no other COPY instructions
- there are no generic instructions from an Dockerfile template

if any of the assumptions didn't match, the image is rebuild in any case

else, these rules apply:
- image is upToDate if both dockerDistTar and dockerCopyDistResources are upToDate```

PR #449, opened 2017-08-30
add ipV4Address to Create Docker Container

ekoon:


cdancy:
Functional tests to prove this works as expected?

cdancy:
@ekoon any possibility to add functional tests for this?

cmaron:
In my own testing (using a variant of `DockerCreateContainer`) I couldn't get this to work without also having a docker network with a subnet that would contain the IP I tried to use. For example, if I wanted to assign the IP `172.10.10.1` without a network that had that address in it's subnet, I would get this error:

```
[Task worker for ':'] INFO org.apache.http.impl.execchain.RetryExec - I/O exception (com.fasterxml.jackson.core.JsonGenerationException) caught when processing request to {}->unix://localhost:80: Null key for a Map not allowed in JSON (use a converting NullKeySerializer?)
```

I tested with a custom version of DockerCreateNetwork that supported `subnet` and then things worked.

cdancy:
@cmaron that makes sense. So to get an actual test here we would need to amend `CreateDockerNetwork`. Do you have a branch with this code still? Care to send in a PR? I suspect that the work to do that would probably encapsulate this work and we kill 2 birds with 1 stone :)

cdancy:
Taking this in as I tested it and it does work. It does require a complicated test setup which we can circle back around to at some later date.

Issue #448, opened 2017-08-29
Use docker-java parse for binds


rwittrick:
In line with other methods using parse from docker-java, could the same be done for Bind, e.g in DockerThreadContextClassLoader createBind

```groovy
def createBind(String path, String volume) {
  Class bindClass = loadBindClass()
  Method bindParseMethod = bindClass.getMethod("parse", String.class)
  method.invoke(null, String.join(":", path, volume))
}
```

the tests in DockerThreadContextClassLoaderIntegrationTest look like there would be minimal changes, if any

An example of why this is useful:
`createBind('/proc','/hostfs/proc:ro')`
to bind host system drives as readonly.  The default access mode in docker-java is rw.

If that looks sensible, I'll submit a pull request

orzeh:
@rwittrick any PR are welcome!

PR #452, opened 2017-09-01
Update createBind to use docker-java parse method

rwittrick:
RE Issue #448

Update createBind  to use parse in docker-java

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
@@ -223,7 +223,7 @@ class DockerThreadContextClassLoaderIntegrationTest extends AbstractIntegrationT
         when:
         def instance = null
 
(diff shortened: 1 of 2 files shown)


src/integTest/groovy/com/bmuschko/gradle/docker/utils/DockerThreadContextClassLoaderIntegrationTest.groovy
```diff
@@ -223,7 +223,7 @@ class DockerThreadContextClassLoaderIntegrationTest extends AbstractIntegrationT
         when:
         def instance = null
 
-        def binds = ['/my/path': 'my/volume', '/other/path': '/other/volume']
+        def binds = ['/this/path': 'this/volume', '/that/path': '/that/volume', '/other/path': '/other/volume:ro']
```
orzeh: what is the difference between `'/this/path': 'this/volume', '/that/path': '/that/volume'`?

rwittrick: None really I guess, I just appended to the test, with an additional value to test with the extra suffix.  There was a two element array without suffix , I could consolidate the test back to two elements, one with, one without a suffix.

cdancy:
LGTM. Thanks @rwittrick !

Issue #456, opened 2017-09-04
Clarify how to apply the plugin


thokari:
Hi, I just tried to apply the plugin, and I am getting this error:

```
unable to resolve class com.bmuschko.gradle.docker.tasks.image.Dockerfile
   @ line 1, column 1.
     import com.bmuschko.gradle.docker.tasks.image.Dockerfile
     ^
```

I have a `build.gradle` with this content: 

```
buildscript {
    repositories {
        jcenter()
    }

    dependencies {
        classpath 'com.bmuschko:gradle-docker-plugin:3.1.0'
    }
}

apply plugin: com.bmuschko.gradle.docker.DockerRemoteApiPlugin
apply from: 'gradle/docker.gradle'
```

And a file `gradle/docker.gradle` with this content:

```
import com.bmuschko.gradle.docker.tasks.image.Dockerfile
import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage

task createDockerfile(type: Dockerfile) {
    destFile = project.file('build/mydockerfile/Dockerfile')
    from 'ubuntu:12.04'
    maintainer 'Benjamin Muschko "benjamin.muschko@gmail.com"'
}
```

The part in the documentation where it says "If you’re using this plugin from a build file which is not the main build file you must apply plugin like so:" is highly confusing, because it remains unclear in which file the statements should go respectively.

cdancy:
Closing as this has been resolved via #457

PR #457, opened 2017-09-04
Clarify docs

thokari:
This addresses #456.
Not sure about the sentence directly below, where it mentions `mavenCentral()`, when in the example above it says `jcenter()`. Maybe just remove that sentence?


cdancy:
I like it and liked that you provided an actual example. LGTM. Thanks!

PR #68, opened 2015-05-10
Separate application lib from rest of deployed resources.

johnrengelman:
Allows for increase caching when only the application code is changing.

This is what the `Dockerfile` ended up looking like:

```
FROM java:8
MAINTAINER john
EXPOSE 8080
ENTRYPOINT ["/example-books/bin/example-books"]
ADD example-books/ /example-books
ADD app-lib/example-books.jar /example-books/lib/example-books.jar
```

I also changed the name of the task that creates the `Dockerfile` from `dockerDistTar` to `dockerfile` since this task is not longer reliant on the `distTar` task. It actually using the `installDist` task now.

PR #459, opened 2017-09-08
Allow Docker to cache app libraries dir when DockerJavaApplication plugin is used

orzeh:
This PR changes how Docker image is built when `DockerJavaApplicationPlugin` is used.

In created `Dockerfile` application files created by [Distribution plugin](https://docs.gradle.org/current/userguide/distribution_plugin.html) are copied into the image before the application JAR (there are 2 ADD instructions now). This allow Docker to use cache of layer containing app libraries and shell scripts (if they didn't change) when building subsequent images. 

This work is just refreshed version of #68

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -501,7 +501,7 @@ The plugin provides a set of tasks for your project and preconfigures them with 
 [options="header"]
 |=======
 |Task name                 |Depends On                |Type                                                                                 |Description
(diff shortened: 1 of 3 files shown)


eshepelyuk:
Thanks, I'll try to test it as soon as possible.

cdancy:
That was easier to dust off then I thought. LGTM. Haven't given things a go myself and so will let others chime in if they have thoughts one way or another.

eshepelyuk:
It just works :) Great job !

eshepelyuk:
Hi, do you have any plans to release this soon ?

cdancy:
Merged now. Thanks for knocking this out @orzeh !

PR #480, opened 2017-10-06
Fix setting shm size when creating container

### LINKED ISSUE #477: Setting shmSize on DockerCreateContainer Clobbers Other Properties
Setting shmSize on DockerCreateContainer seems to overwrite any other values set on the hostConfig.  Simple example:

```
    targetImageId { buildMyAppImage.getImageId() }
    privileged = true
    shmSize = 2147483648
    portBindings = ['8080:8080']
}
```

When you create the container you'll see shmSize is set but not portBindings or privileged.  Issue seems to be the following:

```
        if(getShmSize() != null) { // 0 is valid input
            def hostConfig = threadContextClassLoader.createHostConfig([shmSize: getShmSize()])
            containerCommand.withHostConfig(hostConfig)
        }
```

The command has a hostConfig already configured and many properties set on it at this point.  Possible solution seems to be to just set the shmSize on the existing hostConfig.

orzeh:
Resolves #477


cdancy:
Don't know how this got by us :)

PR #485, opened 2017-10-22
Add `DockerWaitHealthyContainer` task

gabrielhuff:
## Work Summary

- Add `DockerWaitHealthyContainer` task
  - Create functional tests
    - Create test class
    - Create successful execution test
    - Create timeout test
    - Create quick stop test
    - Create not running test
    - Add cleanup logic (revertable)
  - Create task
    - Create task class
    - Define I/O
    - Create utility
    - Implement task logic
    - Move to `extras` subpackage (revertable)
- Correction 1
  - Update interface: expose polling behavior as API
    - Add reactive tests
      - Create periodic emission test
      - Create last successful emission test 
    - Expose polling cycle as input
    - Invoke `onNext` at each cycle passing the health state string
  - Update implementation
    - Check based on command instead of client
    - Document IO
    - Rewrite polling `while` loop
- Correction 2
  - Update README file

## Notes

- I'm destroying the images and containers created by the tests so that the machine running them doesn't get flooded. I'm not sure if there is a reason why this is not being done on the other tests (maybe to isolate the features being tested or make the code more readable), but this behavior can be removed by reverting commit deb37eb.
- The new task was created under the `com.bmuschko.gradle.docker.tasks.container.extras` package as suggested by @cdancy. It can be moved to the container package by undoing commit c97f9ac.
- I changed the way timeout works (comparing to the snippet from the issue) so that it looks similar to the `DockerWaitContainer` task. The following changes were made:
  -  The `timeout` input can now be `null` (no timeout), which now is the default value
  -  The `checkInterval` input was removed in order to make the API more simple.

## Open Tasks

- [x] **Update README** - Since the task belongs to a new package, I'm not sure about how to update the README file.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.bmuschko.gradle.docker.tasks.container.extras
+
+import com.bmuschko.gradle.docker.tasks.container.DockerExistingContainer
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
+
+class DockerWaitHealthyContainer extends DockerExistingContainer {
+
+    @Input
+    @Optional
+    Integer timeout
+
+    @Override
+    void runRemoteCommand(dockerClient) {
+        logger.quiet "Waiting for container with ID '${getContainerId()}' to be healthy."
+
+        Long deadline = timeout ? System.currentTimeMillis() + timeout * 1000 : null
+        def timeout = { deadline ? System.currentTimeMillis() > deadline : false }
+        while(!timeout()) {
+            if (check(dockerClient)) return
+            sleep(100)
+        }
+        if (!check(dockerClient)) throw new GradleException("Health check timeout expired")
+    }
+
+    private boolean check(dockerClient) {
+        def command = dockerClient.inspectContainerCmd(getContainerId())
+        def response = command.exec()
+        def state = response.state
+        if (!state.running) {
+            throw new GradleException("Container with ID '${getContainerId()}' is not running")
+        }
+        return state.health.status == "healthy"
+    }
+}
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
@@ -22,30 +22,47 @@ import org.gradle.api.tasks.Optional
 
 class DockerWaitHealthyContainer extends DockerExistingContainer {
 
+    /**
+     * Wait timeout in seconds.
+     */
     @Input
     @Optional
     Integer timeout
+
+    /**
+     * Interval between each check in milliseconds.
+     */
+    @Input
+    @Optional
+    Integer checkInterval
 
     @Override
     void runRemoteCommand(dockerClient) {
         logger.quiet "Waiting for container with ID '${getContainerId()}' to be healthy."
 
+        def command = dockerClient.inspectContainerCmd(getContainerId())
         Long deadline = timeout ? System.currentTimeMillis() + timeout * 1000 : null
-        def timeout = { deadline ? System.currentTimeMillis() > deadline : false }
-        while(!timeout()) {
-            if (check(dockerClient)) return
-            sleep(100)
+        long sleepInterval = checkInterval ?: 500
+
+        while(!check(deadline, command)) {
+            sleep(sleepInterval)
         }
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
```diff
@@ -0,0 +1,51 @@
...
+import com.bmuschko.gradle.docker.tasks.container.DockerExistingContainer
+import org.gradle.api.GradleException
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Optional
+
+class DockerWaitHealthyContainer extends DockerExistingContainer {
+
+    @Input
+    @Optional
+    Integer timeout
```
orzeh: Please add doc about time unit

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
```diff
@@ -0,0 +1,51 @@
...
+        def timeout = { deadline ? System.currentTimeMillis() > deadline : false }
+        while(!timeout()) {
+            if (check(dockerClient)) return
+            sleep(100)
+        }
+        if (!check(dockerClient)) throw new GradleException("Health check timeout expired")
+    }
+
+    private boolean check(dockerClient) {
+        def command = dockerClient.inspectContainerCmd(getContainerId())
```
orzeh: Maybe we should reuse command instance instead of recreating it every 100 milliseconds?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    void runRemoteCommand(dockerClient) {
+        logger.quiet "Waiting for container with ID '${getContainerId()}' to be healthy."
+
+        Long deadline = timeout ? System.currentTimeMillis() + timeout * 1000 : null
+        def timeout = { deadline ? System.currentTimeMillis() > deadline : false }
+        while(!timeout()) {
+            if (check(dockerClient)) return
+            sleep(100)
```
orzeh: I would allow user to define it (via `checkInterval`) and set the default value for at least 500 milliseconds.

cdancy:
Let's take into account the use of `onNext` callback and allow the user to get ..
 Something... With each iteration of this wait look. Perhaps pass in the state of the container?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if (!check(dockerClient)) throw new GradleException("Health check timeout expired")
```
cdancy: Feels like we can optimize this somehow. We shouldn't have to make another follow up call once outside of the loop.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private boolean check(dockerClient) {
```
cdancy: Lets rename this to `isContainerHealthy`.

cdancy:
I second @orzeh thoughts: very good start. Thanks for the detailed write up, clean code, and well written functional tests. I added some additional comments as well to address. Thanks!

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
gabrielhuff: Ops! Completely forgot about this one

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
gabrielhuff: The reason why I removed `checkInterval` was because it felt like it was an implementation detail. Currently we are polling the container health status, but maybe the docker API allows us to be notified via callbacks on the future. Anyway, fixed on 161ec4b

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
gabrielhuff: Good idea. Fixed on 34f10e2

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
gabrielhuff: I rewrote this on 7fbfea8

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerWaitHealthyContainer.groovy
gabrielhuff: Since the logic changed on 7fbfea8, I'm sticking with `check` since it's calling `onNext` and verifying the timeout all at once

gabrielhuff:
Thanks again for the feedback. I updated PR description with the new changes (under the 'Correction 1' task).

I followed @cdancy 's interesting suggestion about the `onNext` callback. Also fixed some implementation related stuff.

cdancy:
LGTM as well. @gabrielhuff as far as editing our README I would create another section entitled `Extras`, just below `Networks`, and add a single rowed table with a link to this task and a brief description. Once that is done than we can merge.

cdancy:
LGTM. As soon as this PR goes through I will merge.

gabrielhuff:
@cdancy @orzeh Thanks for the opportunity guys. Keep up the good work, this plugin is amazing!

Issue #473, opened 2017-10-04
Could not find or load main class


karlmuscat:
Hi.

Having issues with creating a basic Docker image using Spring Boot 2.0.0.M3 and gradle-docker-plugin 3.2.0. Everything builds great but upon running the container, it seems like the main class is not being found. The container exits immediately with an error: `'Error: Could not find or load main class ...'`

Attached a sample build.gradle for reference.
[sample.txt](https://github.com/bmuschko/gradle-docker-plugin/files/1354423/sample.txt)

Everything works as it should with Spring 1.5.7.RELEASE and plugin 3.2.0 (or 3.0.11). Upgrading to Spring Boot 2.0.0.M4 is not possible as it generates other issues already reported here [https://github.com/bmuschko/gradle-docker-plugin/issues/467](https://github.com/bmuschko/gradle-docker-plugin/issues/467)

Any ideas?

cdancy:
I don't use Spring so can't say exactly what the issue could be but can you get things going without using this plugin and so successfully? Just trying to understand where the bottleneck is. It seems the image is created successfully but starting the container fails?

cdancy:
@orzeh any thoughts here? I believe you use spring and may be able to provide more insight.

orzeh:
@karlmuscat I can reproduce your issue. It appears because in the docker image there is no JAR with application sources. That being said I don't understand why you use `application` plugin for SpringBoot app? IMO it doesn't make sense and it's the source of problems. Even if the JAR would eventually be in the docker image it would be Spring Boot executable JAR with all dependencies inside it (those dependencies are already in lib folder in the image) and main class located in `BOOT-INF/classes` so the error remains.

karlmuscat:
@orzeh I might be missing something here, but I'm using the `application` plugin because it's documented [here](https://github.com/bmuschko/gradle-docker-plugin#usage-examples-1) and without it, the `docker-java-application` tasks (specifically the `dockerBuildImage`) won't be available. (Warning: gradle noob writing)

With both the `application` and `org.springframework.boot` plugins enabled, 2 distributions are generated. This is expected as per Spring's documentation. The `application` one is broken because of what you explained above. The Spring boot one works fine. The problem is that the docker task is picking up the `application` one instead of the Spring boot .tar. 

From the `gradle-docker-plugin` documentation, I can't see a way how to specify the distribution to add, unless I use the `docker-remote-api`, which I want to avoid. What are my options?

[...]

Issue #467, opened 2017-09-21
gradle and spring boot plugin and errors with docker java


vbasem:
thanks alot for the great plugin. 
we have sadly encountered a strange issue with the docker java plugin ever since we upgraded to spring boot M4.

```
Execution failed for task ':mrm-external_ota:buildImage'.
> Could not resolve all dependencies for configuration ':mrm-external_ota:dockerJava'.
   > Could not determine artifacts for javax.ws.rs:javax.ws.rs-api:2.1
      > Could not get resource 'https://[internalrep]/repository/maven-central/javax/ws/rs/javax.ws.rs-api/2.1/javax.ws.rs-api-2.1.$%7Bpackaging.type%7D'.
         > Could not HEAD 'https://[internalrep]/repository/maven-central/javax/ws/rs/javax.ws.rs-api/2.1/javax.ws.rs-api-2.1.$%7Bpackaging.type%7D'. Received status code 500 from server: Server Error
```

while I have no problem resolving and building the project, it fails while fetching the deps during docker image builds and as I see it ```$%7Bpackaging.type%7D``` is not replaced correctly which should be the package type variable.

If I switch back to spring boot M3 plugin, this will run as well, so not sure whats going on!

Any ideas ?

gnilron:
I have similar issues with Springboot 2.0M4

Seems to be related to jersey and a new version with Springboot 2.0M4, not sure why that causes it. 

I get this stacktrace with M4 but M3 works fine
`
[dockerjava-jaxrs-async-0] ERROR com.github.dockerjava.core.async.ResultCallbackTemplate - Error during callback
java.lang.IllegalStateException: InjectionManagerFactory not found.
	at org.glassfish.jersey.internal.inject.Injections.lambda$lookupInjectionManagerFactory$0(Injections.java:98)
`

orzeh:
@vbasem  @gnilron this seems to be two different issues. Could you both provide minimal build script which reproduces them?

orzeh:
I've just created small Spring Boot M.4 application and was able to build a Docker image without any issues.

cdancy:
This all sounds like typical classpath issues due to older/newer libs on the path. I opened an ISSUE with the docker-java guys to get a shaded version of their plug-in but that seems to have stalled.

@orzeh we could build and publish our own. I really don't want to maintain that but maybe there is a solution here which we could work with. IDK.

vbasem:
[...]

PR #501, opened 2017-11-28
Add support to specify the name of the install and jar tasks to use

The-Alchemist:
The Spring Boot project uses the following goals instead of `distTar`/`distZip`:

```
bootDistTar - Bundles the project as a distribution.
bootDistZip - Bundles the project as a distribution.
installBootDist - Installs the project as a distribution as-is.
```

This is because of [Spring Boot's executable JAR](https://docs.spring.io/spring-boot/docs/current/reference/html/deployment-install.html) format, similar to "fat JARs".

However, the `DockerJavaApplicationPlugin` class is hardcoded to use the "default" tasks:

```
            Sync installTask = project.tasks.getByName(DistributionPlugin.TASK_INSTALL_NAME)
            Jar jarTask = project.tasks.getByName(JavaPlugin.JAR_TASK_NAME)
```

This PR extends the `DockerJavaApplication` class to allow setting the `jar` and `install` tasks by name, defaulting to the original behavior.

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
@@ -122,6 +122,21 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
         task.tag == testTagName
     }
 
+    def "Can configure the jar and install tasks"() {
+        given:
+        applyDockerJavaApplicationPluginAndApplicationPlugin(project)
+        when:
+        project.docker.javaApplication {
+            jarTaskName = 'bootJar'
+            installTaskName = 'installBootDist'
+        }
+        then:
+        project.tasks.findByName(DockerJavaApplicationPlugin.DOCKERFILE_TASK_NAME)
+        project.tasks.findByName(DockerJavaApplicationPlugin.BUILD_IMAGE_TASK_NAME)
+        project.tasks.findByName(DockerJavaApplicationPlugin.PUSH_IMAGE_TASK_NAME)
+    }
+
+
     private void applyDockerJavaApplicationPluginWithoutApplicationPlugin(Project project) {
         project.apply(plugin: DockerJavaApplicationPlugin)
     }
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -17,6 +17,8 @@ package com.bmuschko.gradle.docker
 
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile.CompositeExecInstruction
+import org.gradle.api.distribution.plugins.DistributionPlugin
+import org.gradle.api.plugins.JavaPlugin
 
 class DockerJavaApplication {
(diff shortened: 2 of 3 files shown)


cdancy:
I'm fine with this if it helps others downstream. @orzeh any thoughts/opinions one way or another?

src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
```diff
@@ -122,6 +122,21 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
...
+        applyDockerJavaApplicationPluginAndApplicationPlugin(project)
+        when:
+        project.docker.javaApplication {
+            jarTaskName = 'bootJar'
+            installTaskName = 'installBootDist'
+        }
+        then:
+        project.tasks.findByName(DockerJavaApplicationPlugin.DOCKERFILE_TASK_NAME)
+        project.tasks.findByName(DockerJavaApplicationPlugin.BUILD_IMAGE_TASK_NAME)
+        project.tasks.findByName(DockerJavaApplicationPlugin.PUSH_IMAGE_TASK_NAME)
```
orzeh: What this test is supposed to check?

orzeh:
@The-Alchemist I think this would be a valuable addition to the `DockerJavaApplicationPlugin`, however I would like you to add more tests (see my comments).

The-Alchemist:
@orzeh : Sorry, you're absolutely right.  I wasn't sure what the proper way to test this is, so I just left it as a placeholder.

Would you consider it sufficient to make sure the custom task name is propagated to the `DockerJavaApplicationPlugin`, akin to a unit test?  (What's the best way to do that?)

Or are you expecting a full Spring Boot integration test?

orzeh:
@The-Alchemist it would be nice to have a functional test with Spring Boot app. You can take an [existing test](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy#L32) as an example. Please let me know if you need some help with this.

The-Alchemist:
@orzeh : I added a test which fails, but I'm not sure if it's a false negative.  I'm trying to detect that it runs the `bootJar` task.

P.S. I couldn't put the test inside `DockerJavaApplicationPluginFunctionalTest` because my test requires using a custom plugin, which I couldn't put in `DockerJavaApplicationPluginFunctionalTest` or Gradle could error saying that you can't have `plugins` sections so late in the file.

orzeh:
@The-Alchemist I've updated your PR. Now the test is working. Please review.

The-Alchemist:
@orzeh : Turns out my changes weren't actually affecting the plugin.  I think it had to do with the fact that the custom configuration wasn't being set until _after_ the plugin was applied.

New commit changes that, although I get a strange error (at least locally).

orzeh:
@The-Alchemist you encountered #467

orzeh:
@The-Alchemist I've revisited your work and have had a thought that using `DockerJavaApplicationPlugin` with SpringBoot doesn't make any sense (see #473). I mean, you are enforced to apply `application` plugin only to get `DockerJavaApplicationPlugin` working. Maybe we can refactor in a way that it would be possible to apply only SB plugin? You work seems to be a step in that direction.

@The-Alchemist @cdancy  WDYT?

cdancy:
@orzeh on vacation at the moment which is why I haven't been as responsive. I look at you as the SME around this area so whatever you think is best I trust your judgement.

PR #504, opened 2017-12-15
Support for zero exposed ports in java application

aax:
PR #254 introduced support for multiple exposed ports in java application. However, as far as I can tell, it was still not possible to have no ports exposed at all. This PR should fix this. The default of one exposed port 8080 is unchanged, but it is now possible to set ports = [] and override this default.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -24,11 +24,11 @@ class DockerJavaApplication {
     String maintainer = System.getProperty('user.name')
     @Deprecated
     Integer port = 8080
-    Set<Integer> ports = []
+    Set<Integer> ports
     String tag
 
     Integer[] getPorts() {
-        return ports.size() > 0 ? ports : [port]
+        return ports != null ? ports : [port]
     }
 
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -24,11 +24,11 @@ class DockerJavaApplication {
     String maintainer = System.getProperty('user.name')
     @Deprecated
     Integer port = 8080
-    Set<Integer> ports = []
+    Set<Integer> ports
     String tag
 
     Integer[] getPorts() {
-        return ports.size() > 0 ? ports : [port]
+        return ports != null ? ports : [port]
```
cdancy: Elvis operator?

```
Integer[] getPorts() {
    ports ?: [port]
}
```

cdancy:
One minor comment to address but LGTM as well.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
aax: Elvis doesn't help here, because the empty list is falsy.

cdancy:
@aax thanks for the PR!

Issue #461, opened 2017-09-11
DockerJavaApplicationPlugin support multiple tags


sdavids:
Related to:
https://github.com/bmuschko/gradle-docker-plugin/pull/380

Please make it available in the DockerJavaApplicationPlugin too, thanks.

orzeh:
@sdavids can you provide PR for this?

cdancy:
@sdavids any interest in submitting a PR?

cdancy:
@sdavids closing due to inactivity. Feel free to re-open if you'd like to submit a PR or talk further.

PR #505, opened 2017-12-15
Support multiple tags in DockerJavaApplicationPlugin

aax:
Issue #461 asked for support for multiple tags in DockerJavaApplicationPlugin. This PR gives a way to achieve the same goal not by directly supporting multiple tags in the configuration of javaApplication. But it allows to override the tags configuration of the DockerBuildImage task, which was not possible before.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
@@ -18,6 +18,7 @@ package com.bmuschko.gradle.docker
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
 import com.bmuschko.gradle.docker.tasks.image.DockerPushImage
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile
+import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 import org.gradle.api.distribution.plugins.DistributionPlugin
@@ -104,7 +105,7 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
             description = 'Builds the Docker image for the Java application.'
             dependsOn createDockerfileTask
             conventionMapping.inputDir = { createDockerfileTask.destFile.parentFile }
-            conventionMapping.tag = { determineImageTag(project, dockerJavaApplication) }
+            conventionMapping.tags = { [determineImageTag(project, dockerJavaApplication)] as Set }
         } as DockerBuildImage
     }
 
@@ -122,7 +123,15 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
         project.task(PUSH_IMAGE_TASK_NAME, type: DockerPushImage) {
             description = 'Pushes created Docker image to the repository.'
             dependsOn dockerBuildImageTask
-            conventionMapping.imageName = { dockerBuildImageTask.getTag() }
+            conventionMapping.imageName = {
+                if (dockerBuildImageTask.getTags().size() > 1) {
+                    throw new GradleException(String.format("Pushing multiple tags is not supported, " +
+                        "please override dockerPushImage.imageName property to specify which tag to push. " +
+                        "Expected: exactly 1 tag. Actual: %s tags (%s)",
+                        dockerBuildImageTask.getTags().size(), dockerBuildImageTask.getTags().join(', ')));
+                }
+                dockerBuildImageTask.getTags()[0]
+            }
         }
     }
 }
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
@@ -299,6 +299,60 @@ EXPOSE 8080
         noExceptionThrown()
     }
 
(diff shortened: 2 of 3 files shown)


orzeh:
@aax thanks for the PR! It would be great to have this feature. That being said your implementation is hard to follow at least for me. I image we just add `tags` property to `javaApplication` and tweak a little existing code. @cdancy WDYT?

aax:
@orzeh thanks for the feedback. I decided against the `tags` property in `javaApplication` because I'd like the flexibility to build tags but not push them. So it seemed easier and more logical to me to put this into the respective tasks (`dockerBuildImage`/`dockerPushImage`). 

Plus, `DockerPushImage` doesn't support multiple tags right now. This would have to be dealt with separately, too.

orzeh:
@aax `DockerJavaApplicationPlugin` is very opinionated. More complex scenarios, like the one you mentioned should be handled with low level tasks.

cdancy:
By taking in this PR we'd essentially be breaking existing code if I'm not mistaken? Is there room for having both approaches for the moment with the existing being in a deprecated state until some future release to allow developers time to migrate?

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -104,7 +105,7 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
             description = 'Builds the Docker image for the Java application.'
             dependsOn createDockerfileTask
             conventionMapping.inputDir = { createDockerfileTask.destFile.parentFile }
-            conventionMapping.tag = { determineImageTag(project, dockerJavaApplication) }
+            conventionMapping.tags = { [determineImageTag(project, dockerJavaApplication)] as Set }
```
aax: Using the `tags` property of DockerBuildImage instead of (deprecated) property `tag` allows overriding the default config of `javaApplication`. In DockerBuildImage, `tag` has preference over `tags`, which made it effectless to set `tags` before.

I don't consider this change to be breaking, because the behavior is only changed if someone had the aforementioned effectless property `tags` set.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -122,7 +123,15 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
         project.task(PUSH_IMAGE_TASK_NAME, type: DockerPushImage) {
             description = 'Pushes created Docker image to the repository.'
             dependsOn dockerBuildImageTask
-            conventionMapping.imageName = { dockerBuildImageTask.getTag() }
+            conventionMapping.imageName = {
+                if (dockerBuildImageTask.getTags().size() > 1) {
```
aax: This check forces you to override settings for dockerPushImage iff you specified multiple tags in dockerBuildImage.
This is somewhat ugly, but it's just here because DockerPushImage doesn't support multiple tags to be pushed. Otherwise, the default behavior could be to push all built tags.

The change is not breaking because it behaves exactly the same in the default case of just 1 tag.

aax:
@orzeh I'm happy to provide a solution within javaApplication, but I'm unsure how to deal with the fact, that DockerPushImage does not support this. If you could sketch how you see this working, it would be a great help. Apart from that, I consider my approach as "handling it with low level tasks", because I let javaApplication untouched, but give the user the opportunity to go a level down and achieve the goal with the dockerBuildImage task.

@cdancy I added some comments to the changes in DockerJavaApplicationPlugin.groovy explaining why I think the changes are not breaking. If I'm wrong, please drop me a hint.

cdancy:
@aax fair enough. Feels like then we could simplify things if we added the necessary functionality to `DockerPushImage`?

orzeh:
@aax sorry for my late response.

> Apart from that, I consider my approach as "handling it with low level tasks", because I let javaApplication untouched, but give the user the opportunity to go a level down and achieve the goal with the dockerBuildImage task.

We have some misunderstanding here. What I was trying to say is that if you use `DockerJavaApplicationPlugin` you shouldn't use any _low level tasks_. If you have to, it's probably better to give up using it altogether.

Coming back to your PR, IMO it would be better to add `tags` property to `DockerJavaApplication` (and deprecate `tag`) and create dedicated `DockerPushImage` task for each value in `tags`. @aax @cdancy WDYT?

cdancy:
@orzeh feels like we should just solve this problem within DockerPullImage and let multiple tags be used which in turn would make this PR much easier to code.

orzeh:
@cdancy you mean `DockerPushImage`? But docker doesn't support multiple tags in push image operation.

cdancy:
@orzeh yeah ... I wonder if it would make more sense to use `DockerTagImage` here and support multiple tags there (assuming it can be done) so that we don't have to create X number of `DockerPushImage` task assuming I understand things correctly.

bmuschko:
I suggest that we [clean up all deprecated methods](https://github.com/bmuschko/gradle-docker-plugin/issues/675) first. That will 1) force us to use the non-deprecated methods and 2) we'll have to add a `tags` property for some other tasks as well to make the API look consistent. At the moment it's a bit awkward that some task take one tag whereas others take a list of tags. After that we can make the decision whether is makes sense to expose multiple tags for an extension of the convention plugins.

I am going to close this PR for now. It has gotten quite outdated. Thanks for your work! @aax Would you like to help out with removing the deprecations? We can remove them one by one with individual pull requests if needed.

PR #506, opened 2017-12-15
Add JSON output option to DockerInfo.

jonpspri:
As part of a multi-architecture build, I need to be able access variables from the DockerInfo of the Docker with which I am interacting.  This adds a parameter (dockerInfoFile) to the DockerInfo task which, if set, will trigger a write of JSON representing the DockerInfo received from the Docker instance.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerInfoFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerInfoFunctionalTest.groovy
@@ -18,25 +18,7 @@ class DockerInfoFunctionalTest extends AbstractFunctionalTest {
         result.output.contains('Retrieving Docker info.')
     }
 
-    def "Calls onNext when provided"() {
-        buildFile << """
-            import com.bmuschko.gradle.docker.tasks.DockerInfo
-
-            task dockerInfo(type: DockerInfo) {
-                onNext { logger.quiet "In onNext as expected" }
-            }
-        """
-
-        when:
-        BuildResult result = build('dockerInfo')
-
-        then:
-        (result.output.contains('Retrieving Docker info.')
-        && result.output.contains('In onNext as expected')
-        && !result.output.contains('Containers'))
-    }
-
-    def "Passes dockerinfo to onNext"() {
+    def "Passes dockerinfo to onNext if provided, w/o logger output"() {
         buildFile << """
             import com.bmuschko.gradle.docker.tasks.DockerInfo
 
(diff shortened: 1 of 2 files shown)


orzeh:
@jonpspri thanks for the PR! I think it would be better to follow the convention we use in other tasks with respect to handling results - let's use [`onNext`](https://github.com/bmuschko/gradle-docker-plugin#onnext) handler and pass there object received from Docker daemon. It will be more general solution and one can save the result as JSON or do whatever is needed.

jonpspri:
Fair 'nuff.  Let me take a look at how the onNext handling works.  Stand by... :)

src/main/groovy/com/bmuschko/gradle/docker/tasks/DockerInfo.groovy
```diff
@@ -37,5 +41,9 @@ class DockerInfo extends AbstractDockerRemoteApiTask {
         logger.quiet "NGoroutines          : $info.nGoroutines"
         logger.quiet "Swap Limit           : $info.swapLimit"
         logger.quiet "Execution Driver     : $info.executionDriver"
+
+        if (onNext) {
+            onNext.call(info)
+        }
```
orzeh: Can you add `else` here and move logger statements into it?

jonpspri: Oh... so display Or onNext but not both.  Hmm... I kind if liked still having the logger, but that might be a matter of personal taste.  I suppose one can always manually log from within the onNext.

jonpspri:
Also... will start working on Functional test.  Thanks for the example pointer.

src/main/groovy/com/bmuschko/gradle/docker/tasks/DockerInfo.groovy
jonpspri: Done.

jonpspri:
Done.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerInfoFunctionalTest.groovy
```diff
@@ -17,4 +17,44 @@ class DockerInfoFunctionalTest extends AbstractFunctionalTest {
         then:
         result.output.contains('Retrieving Docker info.')
     }
+
+    def "Calls onNext when provided"() {
```
cdancy: Can we can combine the expectations of this test with `Passes dockerinfo to onNext`?

cdancy:
One minor comment to address as I think we can combine the functional tests written into 1. Other than that it LGTM.

jonpspri:
Yeah, I was thinking about pulling them together after I wrote the two of them and saw how similar they were.  There are different schools of test cases -- "Write lots because the differences between them can sometimes add fidelity to diagnosing a problem" and "Write fewer and still cover cases because it's less to maintain."  I usually adhere to the former school, but given how heavy these integration cases are (after watching my Mac churn for an hour), I'm happy to move toward the latter.  Stand by for commit...

jonpspri:
et voila.

cdancy:
@jonpspri thanks for the PR!

Issue #518, opened 2018-01-24
DockerWaitHealthyContainer throws exception


bmuschko:
My task for verifying the health check fails with the exception below.

```
task waitOnHealthyContainer(type: DockerWaitHealthyContainer) {
    group = 'Docker'
    mustRunAfter startContainer
    targetContainerId { createContainer.getContainerId() }
}
```

Could it be that the Docker Java version got out-of-date or is not compatible with the latest version of Docker? I am using Version 17.12.0-ce-mac49 (21995).

```
> Task :waitOnHealthyContainer FAILED
Waiting for container with ID '4adb593414eecfb5aae710fc4b95832a7a21d80ccb3a8f7428a81ef94a345772' to be healthy.


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':waitOnHealthyContainer'.
> Cannot get property 'status' on null object
```

The stack trace is here:

```
Caused by: java.lang.NullPointerException: Cannot get property 'status' on null object
        at com.bmuschko.gradle.docker.tasks.container.extras.DockerWaitHealthyContainer.check(DockerWaitHealthyContainer.groovy:62)
        at com.bmuschko.gradle.docker.tasks.container.extras.DockerWaitHealthyContainer.runRemoteCommand(DockerWaitHealthyContainer.groovy:47)
        at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$_runReactiveStream_closure1.doCall(AbstractDockerRemoteApiTask.groovy:57)
        at com.bmuschko.gradle.docker.utils.DockerThreadContextClassLoader.withClasspath(DockerThreadContextClassLoader.groovy:55)
        at com.bmuschko.gradle.docker.utils.ThreadContextClassLoader$withClasspath.call(Unknown Source)
        at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask.runInDockerClassPath(AbstractDockerRemoteApiTask.groovy:62)
        at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$runInDockerClassPath.callCurrent(Unknown Source)
        at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask$runInDockerClassPath.callCurrent(Unknown Source)
        at com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask.runReactiveStream(AbstractDockerRemoteApiTask.groovy:56)
[...]

PR #520, opened 2018-01-25
If NPE thrown trying to access state.health.status than catch NPE and…

cdancy:
… attempt to access state.status

Addresses #518

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWaitHealthyContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWaitHealthyContainerFunctionalTest.groovy
@@ -69,6 +69,45 @@ class DockerWaitHealthyContainerFunctionalTest extends AbstractFunctionalTest {
         build('waitContainer')
     }
 
+    def "Wait for a generic container to be healthy"() {
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.image.*
+            import com.bmuschko.gradle.docker.tasks.container.*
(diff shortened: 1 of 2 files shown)


cdancy:
@orzeh can you please take a look at how I refactored the logic a bit here to see if makes sense to you and if there is anything I'm missing? The idea is that if the `Health` object is null, and thus not present, then we fall back to just using the generic status of the container.

cdancy:
@bmuschko agreed however we already have loads of tests around this area, one that looks exactly like your snippet above, and I'm still scratching my head as to why when the tests run this `Health` object is present and yet when I do a `docker inspect <container-name>` on my system I don't see it. Have to do a bit more digging here as to what is going on. Maybe the `Health` object goes away after some point once the container is live and running? IDK.

orzeh:
@cdancy I think `Health` object is only returned when [`HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) instruction is used. IMO the task `DockerWaitHealthyContainer` is supposed to be used with conjunction with it. If we don't receive health object it means something is not configured well and we should add log statement or/and break build. This probably requires dedicated functional test as it will document the expected behavior. WDYT?

cdancy:
@orzeh agreed on the logging. Maybe then if `Health` is missing we just return true here or do you think it makes sense to use the implementation I put in place. I'm on the fence here and could go either way.

orzeh:
@cdancy I would use your new implementation with logging.

cdancy:
Sounds good. Thanks @orzeh !

PR #529, opened 2018-02-02
Add skipMaintainer to DockerJavaApplication

### LINKED ISSUE #445: MAINTAINER deprecated in Docker Engine v1.13.0
https://docs.docker.com/engine/deprecated/#maintainer-in-dockerfile

https://github.com/bmuschko/gradle-docker-plugin#user-content-extension-properties-1

"maintainer - Default value: Value of user.home system property"

Please provide a way not to set MAINTAINER or set an appropriate LABEL instead.

orzeh:
This change will fix #445.

The new property `skipMaintainer` allows to create `Dockerfile` without `MAINTAINER` instruction in it. This preserves backward compatibility. In 4.x release it should be removed and default behavior must be changed to use `LABEL`.


cdancy:
An acceptable solution all things considered. LGTM.

PR #531, opened 2018-02-03
Add DockerLoadImage and DockerSaveImage tasks

### LINKED ISSUE #395: add DockerSaveImage and DockerLoadImage
It would be very helpful to have tasks in this plugin that implement the equivalent of 
(after docker pull image) docker save image | gzip > name.tar.gz
and 
docker load image < name.tar.gz

orzeh:
Resolves #395

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -686,6 +686,82 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
         build('workflow')
     }
 
+    def "Can build an image and save it to file"() {
+        File imageDir = temporaryFolder.newFolder('images', 'minimal')
+        File dockerFile = createDockerfile(imageDir)
+        String imageName = createUniqueImageId()
+        File savedImage = temporaryFolder.newFile()
+
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+            import com.bmuschko.gradle.docker.tasks.image.DockerSaveImage
+
+            task buildImage(type: DockerBuildImage) {
+                inputDir = file("${dockerFile.parentFile.path}")
+                tag = "${imageName}"
+            }
+
+            task saveImage(type: DockerSaveImage) {
+                dependsOn buildImage
+                repository = "${imageName}"
+                destFile = new File("${savedImage.absolutePath}")
+            }
+
+            task workflow {
+                dependsOn saveImage
+            }
+        """
+
+        expect:
+        build('workflow')
+        savedImage.exists()
+        savedImage.newInputStream().available()
+    }
+
+    def "Can build an image and save it to file and load"() {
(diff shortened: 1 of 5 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -686,6 +686,82 @@ class DockerWorkflowFunctionalTest extends AbstractFunctionalTest {
...
+
+            task workflow {
+                dependsOn saveImage
+            }
+        """
+
+        expect:
+        build('workflow')
+        savedImage.exists()
+        savedImage.newInputStream().available()
```
cdancy: The `exists()` will always return true since we created the file and the `available()` check _should_ always return true here as there will always be something to read ... I think?

What about removing this test altogether and just focus on the one you wrote below whilst adding additional tasks to stand up a container based on the image we just loaded and check if it's started, has ports open, etc?

cdancy:
Good work and thanks for this. We've had this on our backlog for sometime now but have never gotten around to it. Minor comment to address with respect to the test as I think just 1 will do here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerLoadImage.groovy
```diff
@@ -0,0 +1,15 @@
+package com.bmuschko.gradle.docker.tasks.image
+
+import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
+import org.gradle.api.tasks.Input
+
+class DockerLoadImage extends AbstractDockerRemoteApiTask {
+
+    @Input
+    InputStream imageStream
```
cdancy: Think I'm leaning against this. I _think_ it's more natural, at least in the gradle world, to just pass in a `File` instance for things of this nature. However I can see the usefulness of passing in an `InputStream` as that can be very flexible. Hmmm ... what about turning this into a `def` which could than allow for the user to pass in either a `File` or `InputStream` instance and we simply check at runtime what we have and proceed accordingly?

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
orzeh: Yeah, `exists()` is redundant. `anailable()` is used by `docker-java` folks to test their code so I would leave it.

I like the idea of a single test, the workflow might be: build -> save -> remove -> load -> create container.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerLoadImage.groovy
orzeh: In Groovy you can easily create `InputStream` from `File`: 
````groovy
new File().newInputStream()
````
so I wouldn't change that.

cdancy: Fair enough

cdancy:
Thanks for the PR @orzeh!

Issue #532, opened 2018-02-04
Renaming Container / add DockerRenameContainer


brauchle:
Would be great to add the ability of renaming containers (would need it for a deployment process, otherwise I would have to remove the old container before I can create a new one with the same name)

The `docker-java` library seems to be able to do it.

Link to the api [https://docs.docker.com/engine/api/v1.35/#operation/ContainerRename](https://docs.docker.com/engine/api/v1.35/#operation/ContainerRename)

```
POST /containers/{id or name}/rename?name=new_name
Rename the container id to a new_name

orzeh:
@brauchle would you like to send PR?

cdancy:
@orzeh planned on doing a new release either today or tomorrow. If OP does,not chime in I can knock this out as you've done more than your share the past couple days ;)

orzeh:
@cdancy sounds good 👍

brauchle:
Unforunately i am not that deep into this project.
Would be great if one of you could do it 😃

cdancy:
Task `DockerRenameContainer` has been added and version `3.2.4` has been released.

brauchle:
Just tested it, works like a charm!
Thanks guys!

PR #533, opened 2018-02-05
Added task DockerRenameContainer

cdancy:
Addresses #532


orzeh:
One minor comment - we should add new task to README.

cdancy:
Good catch. Adding now

PR #548, opened 2018-02-24
Isolated class loader

cdancy:
Initial shot at trying to use a custom class-loader when dealing with `docker-java` libs.


cdancy:
Closing/Rejecting this PR, in favor of #550 , as it's too much of a pain to get working properly.

PR #549, opened 2018-02-25
Honor DOCKER_CERT_PATH env var if present

### LINKED ISSUE #327: Add default behavior for docker.certPath if default behavior for docker.url is used
I like the behavior added by #310 where docker.url will default to System.env.DOCKER_HOST when DOCKER_HOST is set, or unix:///var/run/docker.sock when DOCKER_HOST is not set.

When docker.url and docker.certPath are not specified, I would like default behavior for docker.certPath.  My current workaround is:

```groovy
docker {
    certPath = url.startsWith('tcp://') ? new File("$System.env.DOCKER_CERT_PATH") : null
}
```

aol-nnov:
fixes #327

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
@@ -33,6 +33,7 @@ class DockerExtension {
     public DockerExtension(Project project) {
         this.project = project
         this.url = getDefaultDockerUrl()
+        this.certPath = getDefaultDockerCert()
     }
 
     void registryCredentials(Closure closure) {
@@ -64,6 +65,17 @@ class DockerExtension {
         }
         project.logger.info("Default docker.url set to $dockerUrl")
         dockerUrl
+    }
+
+    public File getDefaultDockerCert() {
+        String dockerCertPath = System.getenv("DOCKER_CERT_PATH")
+        if(dockerCertPath) {
+            File certFile = new File(dockerCertPath)
+            if(certFile.exists()) {
+                return certFile
+            }
+        }
+        return null
     }
 
     @CompileStatic


src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
```diff
@@ -33,6 +33,7 @@ class DockerExtension {
     public DockerExtension(Project project) {
         this.project = project
         this.url = getDefaultDockerUrl()
+        this.certPath = this.url.startsWith('tcp://') ? new File(System.getenv("DOCKER_CERT_PATH")) : null
```
cdancy: If the user has this env-var set, for whatever reason, lets just set it here and do something like:
```
def certPathEnvVar = System.getenv("DOCKER_CERT_PATH")
if (certPathEnvVar) {
    this.certPath = project.file(certPathEnvVar)
    if (!this.certPath.exists()) {
        throw new GradleException("Could not find DOCKER_CERT_PATH @ ${certPathEnvVar}")
    }
}
```

aol-nnov: @cdancy do we nececerily need an exception in the constructor?
please see my followup.

missing disclamer: I'm not a groovy dev at all.

src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
```diff
@@ -66,6 +67,17 @@ class DockerExtension {
...
+        String dockerCertPath = System.getenv("DOCKER_CERT_PATH")
+        if(dockerCertPath) {
+            File certFile = new File(certPath);
+            if(certFile.exists() && this.url.startsWith('tcp://')) {
+                return certFile
+            }
+        }
+        return null;
+    }
+
```
cdancy: This looks better. Why do we only want to use this if the `url` scheme is `tcp`?

aol-nnov: > Why do we only want to use this if the url scheme is tcp?

it's a snippet from the original issue which was proposed as a workaround.
Also, I'm not sure, if we need a cert for unix socket or windows pipe.

cdancy: I'd remove that check as there is nothing stating it can't be used in those places. Other than that things LGTM.

cdancy: And for original comment that has since disappeared: I believe if we don't thrown an exception here, as part of this check, docker-java will eventually make a follow up check and thrown an exception there. My thinking was to fail as early as possible.

aol-nnov: >  docker-java will eventually make a follow up check and thrown an exception there

but since I return `null` if cert file does not exits, upstream will be happy and we'll proceed as if there were no cert at all, got docker barking, scratch our head and back to square #1 checking our cert.

It looks to me as premature optimization, which is not always good, you know. :)

cdancy: hahahaha yes yes I know ;) I'm not sure of the internals of `docker-java`, or how the check is handled, and if things are null will the check be skipped downstream? I seem to remember they will do an additional check, if no value/null is set, but its been a while since I looked at their codebase.

cdancy:
Thanks for the long awaited fix @aol-nnov ;)

PR #550, opened 2018-02-25
Use docker-java-shaded in place of docker-java

cdancy:
As our users keep getting bit by transitive deps with docker-java this is an attempt at creating an uber/fat/shadow jar to use in its place. As `docker-java` doesn't provide an artifact like this I went ahead and created one for us or really for anyone that may need it.

Still a work in progress just wanted to get the general PR out there to work on. Changed a couple repo definitions as I was working on things locally and wanted to access mavenLocal. Also changed things to point at jcenter, instead of mavencentral, as the former is a proxy to the latter plus more packages.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,8 @@ apply from: "$rootDir/gradle/publishing.gradle"
 apply from: "$rootDir/gradle/documentation.gradle"
 
 repositories {
-    mavenCentral()
+    mavenLocal()
+    jcenter()
 }
 
 dependencies {
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -57,7 +57,8 @@ abstract class AbstractFunctionalTest extends Specification {
             }
 
             repositories {
-                mavenCentral()
+                mavenLocal()
+                jcenter()
             }
         """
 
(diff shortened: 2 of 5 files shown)


cdancy:
Currently failing due to the latest version is not yet synched to jcenter.

cdancy:
Contacted JFrog support on the above issue and it turns out there was a linkage problem on their end. Travis-CI is now moving along as expected.

bwolf:
It seems to me that `buildscript { repositories { jcenter() } }` is not enough to resolve the `docker-java` artifact. I had to add `jcenter()` to by `subprojects { repositories { ... } }` (originally I had there just mavenCentral). I am using a root build script plus several subprojects with their respective build scripts. The root project uses the subprojects clojure to define common behaviour (e.g. build docker images).

cdancy:
@bwolf that is odd. Do you have a reproducible snippet you can show us?

bwolf:
Lets see if I can extract the relevant parts to provide a quickstart.

bwolf:
Ok, I extracted the relevant bits and put them in a repo: https://github.com/bwolf/gradle-docker-issue. Maybe I'm doing something wrong. Let's see…

cdancy:
@bwolf I was able to reproduce and indeed this is _something_ like a bug though I'm not sure. It has to do with how we figure out where we load our [internal dependencies from at runtime](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy#L46). Because you define the `mavenCentral()` repo in your subproject, which is where you are also applying this plugin, we see that some project repo(s) are defined and so attempt to use those. The logic I think needs to be cleaned up since we moved to the shaded version of this plugin which is currently only hosted on jcenter.

The good news about all of this is that you don't actually have to define mavenCentral at all and instead can use jcenter instead as it acts as a proxy to the former. This will in turn solve your immediate problem and gave you 1 avenue to go through to get your deps.

In the meantime I'll open a bug to refactor the logic here as it's a bit wonky given how things are setup now.

Issue #565, opened 2018-04-02
Add option to DockerSaveImage task for image compression


scottresnik:
### Expected Behavior
Add a property to the DockerSaveImage task, e.g. "compressImage", that when set to true would compress the image to be saved using GZip.  The default setting would be false to be consistent with existing behavior.

### Current Behavior
There is no such setting on the DockerSaveImage task.  Currently images are saved as a tar.

### Context
Currently I zip up images using Gradle for transport to a remote docker container.  Compression drops the size of my image enough to make it worth my while.

### Steps to Reproduce (for bugs)
Not really applicable.

### Your Environment
Not a bug.

cdancy:
@scottresnik hmmm ... we build upon the `docker-java` library which in turn is just a wrapper around OOTB docker commands. I don't think docker supports compression by default in their commands at least last time I checked they did not and the spirit of this library is to be/create gradle wrappers around those same commands.

How are you doing things now? Do you have a code snippet you can share or care to send in a PR so that we can see what you're talking about? Indeed tar files are not compressed and are just a packaged version of files and so can be quite big and having something like this might make sense.

scottresnik:
Right now, I'm using the gradle zip task.  The tar task allows gzip compression, but it ended up creating a gzipped tar of a tar which is a bit of a mess.

I have the code done already, I can create a PR.

cdancy:
Closing as this functionality has now been merged to master.

PR #566, opened 2018-04-02
ISSUE-565: Add option to compress, using gzip, saved docker images.

scottresnik:
This represents the changes required to compress the tar based image returned from docker-java via gzip (#565).

I did not see any existing tests for the DockerSaveImage task.  I would be happy to add one if someone could point out a good example to base a test off of.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
@@ -1,18 +1,24 @@
 package com.bmuschko.gradle.docker.tasks.image
 
 import com.bmuschko.gradle.docker.AbstractFunctionalTest
+import com.bmuschko.gradle.docker.utils.IOUtils
 import org.gradle.testkit.runner.BuildResult
+import spock.lang.Stepwise
 
+import java.util.zip.GZIPInputStream
+
+@Stepwise
 class DockerSaveImageFunctionalTest extends AbstractFunctionalTest {
 
     final String ImageId = createUniqueImageId()
     final String controlSavedImage = "build/docker/${ImageId}-docker-image-control.tar"
+    final String imageFile = "build/docker/${ImageId}-docker-image.tar"
+    final String compressedImageFile = "build/docker/${ImageId}-compressed-docker-image.tar.gz"
 
     def setup() {
         buildFile << """
             import com.bmuschko.gradle.docker.tasks.image.Dockerfile
             import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
-            import com.bmuschko.gradle.docker.tasks.image.DockerSaveImage
 
             task dockerfile(type: Dockerfile) {
                 from '$TEST_IMAGE_WITH_TAG'
@@ -24,16 +30,10 @@ class DockerSaveImageFunctionalTest extends AbstractFunctionalTest {
                 tag = "${ImageId}"
                 labels = ["setup":"${ImageId}"]
             }
-            task saveImageControl(type: DockerSaveImage) {
-                dependsOn buildImage
-                tag = "${ImageId}"
-                repository = "${ImageId}"
-                destFile = file("${controlSavedImage}")
-            }            
         """
 
         when:
-        BuildResult result = build('saveImageControl')
+        BuildResult result = build('buildImage')
 
         then:
         result.output.contains("Created image with ID")
@@ -47,7 +47,7 @@ class DockerSaveImageFunctionalTest extends AbstractFunctionalTest {
                 dependsOn buildImage
                 tag = "${ImageId}"
                 repository = "${ImageId}"
-                destFile = file("build/docker/${ImageId}-docker-image.tar")
+                destFile = file("${imageFile}")
             }
         """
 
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/utils/IOUtils.groovy
```diff
@@ -23,4 +23,14 @@ final class IOUtils {
             // ignore
         }
     }
+
+    static void closeQuietly(OutputStream output) {
```
cdancy: I would just change the original definition to take in a `Closeable instance` instead of creating another method.

cdancy:
We'd also need a functionalTest for this. Just search through our existing functionalTests to see where we use this task and add something there. Feel free to reach out if you have any questions.

orzeh:
@scottresnik I like the idea. As @cdancy wrote, just add functional test :+1:

scottresnik:
I just pushed a very simple functional test.  It calls the task with and without compression.  Then makes sure there were no exceptions.  Is it worth going deeper to verify that the compressed image is smaller than the non compressed image?  i.e. is the added property doing anything?

cdancy:
@scottresnik yeah I think that makes sense just to ensure things are actually being compressed.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+                destFile = file("build/docker/${ImageId}-docker-image.tar")
+            }
+        """
+
+        when:
+        build('saveImage')
+
+        then:
+        noExceptionThrown()
+        fileSizesMatch(new File(projectDir, "build/docker/${ImageId}-docker-image.tar"), new File(projectDir, controlSavedImage))
```
orzeh: why do you compare size of the same file? Or am I missing something?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+    void fileSizesMatch(File file1, File file2) {
+        long size1 = file1.size()
+        long size2 = file2.size()
+        if( size1 < size2 ) {
+            assert (size2 % size1) <= 100
+        }
+        else {
+            assert (size1 % size2) <= 100
+        }
```
orzeh: What is the purpose of this `if`?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
scottresnik: It's not the same file, it's the result of the same task being called twice.

What is the expectation of unpacking the tar of a docker image?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
scottresnik: If mod is called with a smaller value for the left operand, then that operand is returned.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
orzeh: OK, now I see how it works.

scottresnik: It's simple and crude.  Regarding your comment above:
>In functional test I would rather test if archive can be unpacked and read.

Do you have some insight about how I could verify the saved image based on something from docker?  Or is the expectation that it is already handled by the upstream docker-java project?

orzeh:
@scottresnik I write here instead of a comment.

I would write this test as follows:
* to test if save without compression works I would just check if destination file exists and has > 0 bytes
* to test if the compression works I would check if I can unpack gzipped archive.

Additional tests are done on the other side of the fence - in `docker-java`.

@cdancy any thoughts?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
cdancy: Feels like we can get away with something simple like: 
```
assert file('compressed-file.tar.gz').length() < file('compressed-file.tar.gz').length()
```

cdancy: I would do the same above just in reverse to keep things stupid-simple.

cdancy:
Couple further comments inline about the file size check which I think we can simplify some.

> to test if save without compression works I would just check if destination file exists and has > 0 bytes

@orzeh agreed but I like what he did with the "control file" (which I asked for) to test that it is bigger/smaller than some existing entity.

> to test if the compression works I would check if I can unpack gzipped archive.

I _think_ with the above checks we should be fine and don't have to unpack things here. I think if we did that we'd be testing java's ability to do things correctly and if there were a problem with the implementation it would pop at test time.

cdancy:
@orzeh Ok I see what you're saying now. Indeed I have run into "zip corruption" issues myself for seemingly random reasons which is why I hesitate to check for it here and trust that java is doing its job (and in the odd case its not leave it up to the dev to deal with). Lets go with what we have here as I think its clean enough and gets the job done.

cdancy:
@scottresnik merged to master. Thanks for the PR!

PR #572, opened 2018-04-19
Add javax.activation dep for users who are working with jdk9+

cdancy:


orzeh:
LGTM :+1: 

BTW maybe we should setup Travis to test also on newer JDKs?

cdancy:
@orzeh that's not a bad idea. No idea how to do that at the moment however. I believe the `docker-java` guys do something similar and maybe we can copy/paste from there.

PR #574, opened 2018-04-20
Add support for specifying a user/group within DockerExecContainer

### LINKED ISSUE #573: Allow specifying a user within DockerExecContainer tasks to execute commands
### Expected Behavior
Within a DockerExecContainer task an end-user should be able to specify the user the command should execute as if the docker image doesn't properly setup the USER attributes and the the initial entrypoint needs to be executed as the root user. 

### Current Behavior
This can not be performed, unless I missed something, as the DockerExecContainer task is missing the input to pass it to the underlining docker-java library

### Context
We have some generic docker containers that have entrypoints that setup environments for easier end-user access. These entrypoints need to be run as root, so the Dockerfile USER will not work and passing the user argument to a non-root user within the DockerCreateContainer task will cause the entrypoint to fail.

This directly would map to the docker exec -u  argument.

### Initial Work 
A patch for this can be seen here: https://github.com/stevendpclark/gradle-docker-plugin/commit/a28d8ef8d4a9b9b513671d306ff982a26a13f505 

If there are no objections I could open up a PR request to merge in the change.

stevendpclark:
* Allow a task to specify what user/group combination to execute
   the command as within the Docker container.
 * Added a simple integration test that validates expected user/group
   through the use of id commands to display the current user/group.

closes #573

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerExecContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerExecContainerFunctionalTest.groovy
@@ -110,7 +110,7 @@ class DockerExecContainerFunctionalTest extends AbstractFunctionalTest {
             ex.message.contains('is not running')
     }
 
-    def "Execute command as a non-root user running container"() {
+    def "Execute command as a non-root user within a running container"() {
         buildFile << """
             import com.bmuschko.gradle.docker.tasks.image.DockerPullImage
             import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
@@ -139,7 +139,7 @@ class DockerExecContainerFunctionalTest extends AbstractFunctionalTest {
                 dependsOn startContainer
                 targetContainerId { startContainer.getContainerId() }
                 cmd = ['sh', '-c', 'id -u && id -g']
-                user = "10000:10001"
+                user = '10000:10001'
             }
 
             task logContainer(type: DockerLogsContainer) {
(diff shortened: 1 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/DockerExecContainerFunctionalTest.groovy
```diff
@@ -109,4 +109,60 @@ class DockerExecContainerFunctionalTest extends AbstractFunctionalTest {
...
+            task startContainer(type: DockerStartContainer) {
+                dependsOn createContainer
+                targetContainerId { createContainer.getContainerId() }
+            }
+
+            task execContainer(type: DockerExecContainer) {
+                dependsOn startContainer
+                targetContainerId { startContainer.getContainerId() }
+                cmd = ['sh', '-c', 'id -u && id -g']
+                user = "10000:10001"
```
cdancy: Single quotes instead of doubles here and below?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.groovy
```diff
@@ -32,6 +32,10 @@ class DockerExecContainer extends DockerExistingContainer {
     @Optional
     Boolean attachStderr = true
 
+    @Input
+    @Optional
+    String user
```
orzeh: Javadoc with information about format would be valuable I guess.

cdancy: Was just about to say the same :)

stevendpclark: Thanks for the review feedback, would you want  `@since 3.2.8` within the javadoc as well?

orzeh: sure!

PR #579, opened 2018-05-02
declare entrypoint in shell form

iikirilov:
Added `entryPointShellForm` to be able to specify and entrypoint of an image in shell form rather than using a generic `instruction`

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
@@ -19,6 +19,7 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -19,6 +19,7 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.OutputFile
 import org.gradle.api.tasks.TaskAction
+import org.gradle.internal.impldep.org.apache.commons.collections.Closure
```
bmuschko: This implementation uses an internal class from the Gradle API. It should be avoided if possible. Maybe that was just an incorrect import and was supposed to refer to `groovy.lang.Closure`.

bmuschko:
The syntax is not vastly different from the existing API. Instead of a List you use a String. From my perspective it doesn't justify extending the API. It will just make it harder for users to pick a method. 

I'd like to close this PR. Thanks for your work! If you want to have the functionality then it's very easy to extend from the `Dockerfile` class and use it in your own project or plugin.

PR #585, opened 2018-05-22
Bug fix to account for docker-java changing name of cpuset to cpusetcpus

cdancy:
Minor fix for param rename as well as add option to existing test to ensure this is at least tested in the future.


cdancy:
Thanks @orzeh ! Merging now.

PR #589, opened 2018-05-29
Add new task `DockerClient`

cdancy:
Task whose sole purpose is to hand back the `docker-java` client to the defined `onNext` closure. If it's not defined we print out a ridiculous message.


orzeh:
LGTM :+1:

cdancy:
Thanks for the review @orzeh !

PR #595, opened 2018-06-06
Explicitly call toString() on values in maps passed to Docker API

### LINKED ISSUE #404: Many properties fail if GString is used instead of String
Example: 
```

ext { 
  imageName =  getProp("imageName", "${group}/" + getProp("image", project.name))
   tagVersion = getProp("tagVersion", "${->version }" )

// ... 
}


task buildImage( type: DockerBuildImage ) {
    // ...
    tags = [ project.imageName , project.tagVersion ]
}

```
Fails with 

 java.lang.ClassCastException: org.codehaus.groovy.runtime.GStringImpl cannot be cast to java.lang.String org.codehaus.groovy.runtime.GStringImpl cannot be cast to java.lang.String
Created null. image: null 

Workaround:

  tags = [ ''+ project.imageName , '' + project.tagVersion ]

orzeh:
Fix #404

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -374,7 +374,7 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         }
 
         if(getLabels()) {
-            containerCommand.withLabels(getLabels())
+            containerCommand.withLabels(getLabels().collectEntries { [it.key, it.value.toString()] })
         }
 
         if(getMacAddress()) {
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -119,7 +119,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         } else if (getTags()) {
             def tagListString = getTags().collect {"'${it}'"}.join(", ")
             logger.quiet "Using tags ${tagListString} for image."
-            buildImageCmd.withTags(getTags())
+            buildImageCmd.withTags(getTags().collect { it.toString() }.toSet() )
         }
 
         if (getNoCache()) {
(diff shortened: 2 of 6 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -374,7 +374,7 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         }
 
         if(getLabels()) {
-            containerCommand.withLabels(getLabels())
+            containerCommand.withLabels(getLabels().collectEntries { [it.key, it.value.toString()] })
```
cdancy: Should we do a `toString()` on the key as well?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -139,7 +139,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         }
 
         if (getLabels()) {
-            buildImageCmd.withLabels(getLabels())
+            buildImageCmd.withLabels(getLabels().collectEntries { [it.key, it.value.toString()] })
```
cdancy: Maybe a `toString()` on the key here as well?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerListImages.groovy
```diff
@@ -53,7 +53,7 @@ class DockerListImages extends AbstractDockerRemoteApiTask {
         }
 
         if (getLabels()) {
-            listImagesCmd.withLabelFilter(getLabels())
+            listImagesCmd.withLabelFilter(getLabels().collectEntries { [it.key, it.value.toString()] })
```
cdancy: And here if it makes sense?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
orzeh: I'll check this.

orzeh:
@cdancy I've checked keys as GStrings  and there were no errors. I'll leave it as is. You can merge if there are no other comments/issues to address.

cdancy:
@orzeh sounds good. Merging now.

PR #597, opened 2018-06-08
Remove deprecated ResponseHandler.

cdancy:


cdancy:
This looks like a pain in the butt to get right...

bmuschko:
Is anything holding this up? What's the replacement for this functionality?

cdancy:
@bmuschko just fixing all of the broken issues :) Haven't had time to circle back on it.

cdancy:
@bmuschko the replacement is the use of the `onNext` closure which is adopted by various tasks throughout the project where it makes sense.

bmuschko:
@cdancy What exactly are the broken issues? Can you give me links or pointers to code?

cdancy:
@bmuschko checkout the travis ci-jobs or just ran the branch locally. It needs some work. I thought it would be an easy win, and only removed what I thought was needed, but after running things I recognized more needed to be done.

https://travis-ci.org/bmuschko/gradle-docker-plugin/builds/389872555?utm_source=github_status&utm_medium=notification

bmuschko:
@cdancy @orzeh I fixed up the tests. Please have another look.

BTW: The design of `AbstractReactiveStreamsTask` seems to be a bit off. `AbstractDockerRemoteApiTask` extends `AbstractReactiveStreamsTask` which is the base class for pretty much every task in this code base. However, `onNext` is only handled by some of the tasks. I think that's confusing to users. They might provide a closure to `onNext` and nothing will happen with it. We should pull out `onNext` from that class and only introduce it where it's actually being handled. Furthermore, instead of creating a large inheritance hierarchy for tasks (specific task -> `AbstractDockerRemoteApiTask` -> `AbstractReactiveStreamsTask` -> DefaultTask`) we should favor delegation instead.

bmuschko:
I am not quite sure what's up with that Bintray plugin. Maybe there's something wrong with the Git history and it can't handle it properly. We may have to create a new branch from `master` to fix it. I will have a look at it later.

bmuschko:
@cdancy Can you also bring me up to speed on why the `ResponseHandler` approach didn't deem sufficient and was replaced by the callback methods?

cdancy:
The reactive-streams are an industry standard at this point and so adopting that, over a custom solution, seemed the better route. Furthermore the closure model fits better with gradle tasks. 

The implementation, and it not working for every task so to speak, is fair but on purpose. I've been thinking of adopting the model I put together with the kubernetes plugin wherein each task returns something for onNext to process but haven't had the time to do so.

cdancy:
We've also has issues with the bintray plugin and newer versions of gradle here which has stopped us from adopting anything recent. Other priorities have come up and so we haven't had the time to dig into why.

cdancy:
> It should be fairly easy to change the implementation to only allow onNext if a task actually lets you iterate over a response value. That'll make it easier for user to understand what they can do without running into a situation that requires them to debug the code. Who really reads documentation? ;-)

There is nuance here in that some things happen at runtime. I agree this can be cleaned up and the impl could be handled better.

PR #603, opened 2018-06-16
Allow for pushing multiple tags.

cdancy:


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -36,6 +36,13 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
     @Optional
     String tag
```
orzeh: Should we deprecate this?

orzeh:
@cdancy PR looks good, however tests are failing on Travis. Did you check it on your local?

cdancy:
@orzeh did not try locally just threw this together and hoped for the best :) I'll try and take a look at things over the next couple days but have been focused on the `gradle-docker-application-plugin` I noted previously and the `gradle-docker-database-plugin` which builds on the former to easily create dockerized database application stacks. But I'll stop with the self promotion now :)

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
cdancy: I think so. I've been throwing around an idea of bumping the plugin to `4.0.0` and with that include a refactoring of a handful of tasks to allow for multiple things to be worked on (i.e. multiple execs, multiple pushes, etc..).

archvill:
Can we align the meaning of `imageName` and `tags` props for both DockerBuildImage and DockerPushImage? Why latter require _both_ props defined separately? I think it will become even more confusing after replacing `tag` with `tags`.

cdancy:
@archvill the idea is that we would deprecate the use of `tag` and use ONLY `tags`. Both tasks you note would have a `tags` option. Is there something else you were thinking? Feel free to send in a PR to discuss and potentially have merged.

archvill:
@cdancy ,
```
    dependsOn createDockerfile
    inputDir = createDockerfile.destFile.parentFile
    tags = ['${imageName}:111',
            '${imageName}:222',
            '${imageName}:333']
}
task pushImage(type: DockerPushImage) {
    dependsOn buildImage2
    imageName = "${imageName}"
    tags = ['111', '222', '333']
}
```
My question is, can we remove `imageName` in `DockerPushImage` and specify image repository inside each tag in `tags` separately? Or can we add `imageName` to `DockerBuildImage` and replace repository+tag in `tags` here with only image tags?

cdancy:
Closing this for now. We can always circle back on supporting this at a later date if it makes sense.

PR #608, opened 2018-06-26
Expose network mode in the DockerBuildImage task

### LINKED ISSUE #555: How to specify build options when building the docker image
We need to set the --network [options](https://docs.docker.com/engine/reference/commandline/build/#options) for some of our images. Is there any way we can set those when creating the image via DockerBuildImage task?
e.g something equivalent to
docker build -t test/image:latest --network=host --build-arg VERSION=A123 
Thanks!

tlusk:
This simply exposes the docker-java property so it can be used in the Gradle task.

This cannot be merged in until the corresponding pull request in docker-java has been merged and a new release has been made.

See: https://github.com/docker-java/docker-java/pull/1061

closes #555

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -110,6 +110,16 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         result.output.contains("Using cache")
     }
 
+    def "can build image using --network host"() {
+        buildFile << buildImageWithHostNetwork()
+
+        when:
+        build('buildWithHostNetwork')
+
+        then:
+        noExceptionThrown()
+    }
+
     private String buildImageWithShmSize() {
         """
             import com.bmuschko.gradle.docker.tasks.image.Dockerfile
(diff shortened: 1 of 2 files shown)


cdancy:
@tlusk thanks for the PR! If and when this gets merged with those guys, and they spin a new release, feel free to ping us on this end and we will go ahead and merge this PR.

tlusk:
@cdancy The upstream changes for this have been released in docker-java 3.1.0 RC4

bmuschko:
@tlusk I don't think we'll want to pull a release candidate version of Docker Java and release with it. We are going to wait until the final is out.

cdancy:
@bmuschko while I agree with the sentiment we're using an RC version now. Those guys haven't done a proper release since last September and who knows when the next one will be done. The maintainers have said that master, and thus the RC releases, are stable and not failing and testing with them does suggest this is true.

It's for this reason and handful of others I've floated around the idea of using spotify's docker-client lib which at the time was still getting off the ground but has now surpassed feature-wise the docker-java library. It would take some work to switch but might be worth it as it seems to be better maintained and more mature.

bmuschko:
> It's for this reason and handful of others I've floated around the idea of using spotify's docker-client lib which at the time was still getting off the ground but has now surpassed feature-wise the docker-java library. It would take some work to switch but might be worth it as it seems to be better maintained and more mature.

I am all in for it! I was actually looking into switching over years ago. It'd be worth putting together a prototype. I might look into it but want to fix a couple of other issues beforehand.

cdancy:
@bmuschko that would be great and would probably be a fair amount of work so we can divide things up when the time comes.

cdancy:
They also have an uber/shadow/fat jar that they publish and would remove us (myself mainly) from having to host our own shaded version of `docker-java` (to get around gradles classpath cloberring) because those guys don't have the time to code things.

tanelso2:
Is there any chance of this getting merged soon? This would be very useful for our use case

tlusk:
@cdancy I noticed you bumped up the `docker-java` version to `rc5` so this MR should be good to merge in now.

I rebased and added a basic functional test case for this flag.

cdancy:
@tlusk taking it in now. Thanks for the PR!

PR #616, opened 2018-07-07
Shade cglib and its dependencies.

cdancy:


src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
```diff
@@ -60,10 +60,11 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
 
         Configuration config = project.configurations[DOCKER_JAVA_CONFIGURATION_NAME]
         config.defaultDependencies { dependencies ->
-            dependencies.add(project.dependencies.create("com.aries:docker-java-shaded:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION"))
+            dependencies.add(project.dependencies.create("com.aries:docker-java-shaded:$DockerRemoteApiPlugin.DOCKER_JAVA_DEFAULT_VERSION:cglib@jar") {
```
cdancy: Created a special version of the `docker-java-shaded`, via an additional artifact/classifier, which includes the cglib and all its dependencies.

cdancy: I also used `@jar` and `transitive = false` which both amount to the same thing: just grab the single jar and that's it. Both are not needed but I slipped them in anyway to get the point across :)

PR #623, opened 2018-07-22
Correct task input and output annotations

bmuschko:
The task `validateProperties` (provided by the Java plugin dev plugin) verifies that custom tasks properly assign input and output property annotations. It reported a bunch of warnings. This pull request fixes them. Please review (probably best by looking at the commit messages) and let me know if I am missing something.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExistingContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExistingContainer.groovy
@@ -18,7 +18,6 @@ package com.bmuschko.gradle.docker.tasks.container
 import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
 import groovy.transform.CompileStatic
 import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.Internal
 
 import java.util.concurrent.Callable
 
@@ -26,7 +25,7 @@ abstract class DockerExistingContainer extends AbstractDockerRemoteApiTask {
     /**
      * Container ID used to perform operation. The container for the provided ID has to be created first.
      */
-    @Internal
+    @Input
     String containerId
 
     void targetContainerId(Closure containerId) {
@@ -36,10 +35,5 @@ abstract class DockerExistingContainer extends AbstractDockerRemoteApiTask {
     @CompileStatic
     void targetContainerId(Callable<String> containerId) {
         targetContainerId { containerId.call() }
-    }
-
-    @Input
-    String getContainerId() {
-        containerId
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerInspectExecContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerInspectExecContainer.groovy
@@ -20,11 +20,6 @@ class DockerInspectExecContainer extends AbstractDockerRemoteApiTask {
         conventionMapping.execId = execId
     }
 
-    @Input
-    String getExecId() {
-        execId
-    }
-
     @Override
     void runRemoteCommand(Object dockerClient) {
         logger.quiet "Inspecting exec with ID '${getExecId()}'."


cdancy:
Feels weird to put internal on the on* closures as they are for public consumption. It will shut gradle up but I wonder how developers will interpret this.

bmuschko:
> Feels weird to put internal on the on* closures as they are for public consumption. It will shut gradle up but I wonder how developers will interpret this.

`@Internal` only means that they are not taken under consideration for incremental build functionality. I think the naming is unfortunate as it carries a specific meaning to people reading the code. Using `@Optional` doesn't make sense as it has not effect for incremental build functionality. It's also not meant for "just" documentation purposes. The combination of `@Input` and `@Optional` would make more sense but it's not really an input for the task. It's more a post-processing operation.

cdancy:
We could just pull these in via a trait and gradle wouldn't complain. I've done this elsewhere with success as @Internal simply gives off the wrong impression and we already have a hard enough time getting folks to understand the gradle model.

bmuschko:
> we already have a hard enough time getting folks to understand the gradle model.

Can you provide more context? Which folks are we talking about here? Users of this plugin?

The naming of `@Internal` really is a Gradle core issue. People do not understand what it is meant to convey. Input/output annotations are an implementation detail anyway. Users shouldn't have to deal with them. It either works properly or it doesn't. If users have the wrong understanding of what `@Internal` means then we should point users to the Gradle core documentation (which clearly states the meaning).

Using `@Optional` doesn't really help the cause. It's only meant to be used in conjunction with other input/output annotations. Adding `@Optional` just for documentation purposes misses the mark from my perspective.

Using traits strongly binds you to Groovy as the implementation language for the plugin. The same with closures. Instead of closures we should use `Action` to be able to support the Kotlin DSL as well. The Groovy DSL does some magic under the covers to turn a closure into an action.

orzeh:
> The naming of @Internal really is a Gradle core issue. People do not understand what it is meant to convey. Input/output annotations are an implementation detail anyway. Users shouldn't have to deal with them. It either works properly or it doesn't. If users have the wrong understanding of what @Internal means then we should point users to the Gradle core documentation (which clearly states the meaning).

Then maybe we should put fully qualified annotation name so developers/users can see that those are Gradle specific things and reach its documentation if needed?

> Using traits strongly binds you to Groovy as the implementation language for the plugin. 

I agree we shouldn't use traits here.

> The same with closures. Instead of closures we should use Action to be able to support the Kotlin DSL as well. The Groovy DSL does some magic under the covers to turn a closure into an action.

You mean Gradle DSL?

cdancy:
> Can you provide more context? Which folks are we talking about here? Users of this plugin?

Developers here where I work which are across the globe. We have everyone from beginners to experts who are heads down focused on their area and just want something stupid-simple to use and understand.

> Using traits strongly binds you to Groovy as the implementation language for the plugin.

To be fair the plugin is written entirely with groovy with a handful of places we have some hooks for `Kotlin` specific things. Re-writing entirely in Kotlin would be a bad idea as it would alienate users who have not yet migrated. A new plugin with Kotlin only in mind would be something but I don't think we can honestly do that here and then to hold ourselves back from using features of the language this plugin is built on feels wrong. I'm fine with what's in place now as I don't feel overly strongly about it but having worked with loads of developers, from all different backgrounds, this type of thing does come up as developers simply don't have the time to dig into what a given thing is supposed to mean and just take it at face value .... which is what they should be doing as the code itself should be self-documenting.

cdancy:
> Input/output annotations are an implementation detail. Users shouldn't have to know about them if they are just using the API. Linking to the Gradle Javadocs from the plugin Javadocs should do the trick if someone is more interested in the details.

So this is where I disagree. While this may be true, to the end-user, it gives another impression which is the problem I'm referring to. There may not be a better way I'm just noting it causes more confusion than it helps and by not having the @Internal all that happened was we got lots of deprecation warnings while building the project. Irritating for us yes but better documentation wise for the end-user who is going to be looking at tasks to see what its inputs are.

bmuschko:
> There may not be a better way I'm just noting it causes more confusion than it helps and by not having the `@Internal` all that happened was we got lots of deprecation warnings while building the project.

Thanks for giving me your perspective on it. My hunch is that the naming of "Internal" just confuses users. Let's say the annotation was called `@NotConsideredForUpToDateChecking` (or something else) then people probably wouldn't care. Is that a correct assumption? 

I am going to bring this up in the Gradle community Slack channel though it's very unlikely that the annotation will be replaced anything soon. Just wondering...did one of the plugin users bring up the use of `@Internal` in the past?

I understand your point of developers wanting to have comprehensive APIs. However, if you really want to understand Gradle's input/output checking then you will also need to have an understanding of the annotations. Anything that doesn't use "Input" or "Output" in its naming would have the same issue e.g. `Nested`, `Console` or `Destroys` (see the [full list of annotations](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:task_input_output_annotations)). I don't think it's out of this world to ask users to understand the API they are trying to use and its implications. Though I have to admit that the number of annotations have grown over the years and their use may not be fully clear to beginners of Gradle. IMHO this falls more into the bucket of Gradle core.

> Irritating for us yes but better documentation wise for the end-user who is going to be looking at tasks to see what its inputs are.

The callback closures could or could not be considered an input to the task. I see them more as post-processing functionality (maybe similar to adding another `doLast` action that works off the result of the previous action).

PR #627, opened 2018-07-23
Provide links to external APIs so that Groovydocs can create connection

bmuschko:
Leads to a more comprehensive plugin API from the perspective of an end-user.

Also see discussion here: https://github.com/bmuschko/gradle-docker-plugin/pull/623


cdancy:
LGTM

Issue #628, opened 2018-07-23
Add custom flag input to DockerBuildImage, or add explicit support for --cache-from


exFalso:
Currently there's no way to pass in `--cache-from` to `DockerBuildImage`.

### Expected Behavior

We should be able to specify the flag.

### Current Behavior

Currently we cannot specify this flag.

### Context

I'm trying to use the build cache using `--cache-from` to speed up CI builds.

bmuschko:
`DockerBuildImage.cacheFrom` does exist now.

PR #629, opened 2018-07-23
Add support for --cache-from

exFalso:
https://github.com/bmuschko/gradle-docker-plugin/issues/628

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -76,6 +76,10 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
     @Optional
     Map<String, String> buildArgs = [:]
 
+    @Input
+    @Optional
+    Set<String> cacheFrom = []
+
     /**
      * Size of <code>/dev/shm</code> in bytes.
      * The size must be greater than 0.
@@ -154,6 +158,10 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
         buildArgs.each { arg, value ->
             buildImageCmd = buildImageCmd.withBuildArg(arg, value)
+        }
+
+        if (!cacheFrom.isEmpty()) {
+            buildImageCmd = buildImageCmd.withCacheFrom(cacheFrom)
         }
 
         def callback = onNext ? threadContextClassLoader.createBuildImageResultCallback(this.onNext)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -156,6 +160,10 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             buildImageCmd = buildImageCmd.withBuildArg(arg, value)
         }
 
+        if (!cacheFrom.isEmpty()) {
```
cdancy: You should be able to get away with `if (cacheFrom) { ...` as groovy's truthiness check should take into account empty empty collections if I'm not mistaken.

cdancy:
What happens if the cache does not exist or is empty? Maybe add a test, or append to an existing, to confirm one way or another?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: For the time that we still use convention mapping we should rather use the getter methods instead of the direct access to the property. For example see this: https://github.com/exFalso/gradle-docker-plugin/blob/f648a028dccd83a909571fbc983e55f3eb694e69/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy#L153-L154

bmuschko:
@exFalso Are you still interested in addressing the review comments?

bmuschko:
I haven't seen any activity on this pull request. I am going to close this for now. Let us know if you'd like to address the review items in which case we can reopen the pull request.

exFalso:
@bmuschko I've been away, just arrived from holidays. Yes, I'd like to address the comments sometime in the next couple of days

daninthewoods:
Would be awesome to have this feature, thank you both.

exFalso:
@bmuschko @cdancy Addressed comments.

I added three tests, one testing the flag without anything in the cache, one testing a build without the flag but the layer in the cache (in which case the layer *isn't* used), and one that tests the flag with the layer (in which case the layer *is* used)

It was a very good idea to add tests, it turned out dockerjava has a bug with this flag! It seems to double-unmarshall the query parameter. I added an ugly workaround for this, see DockerBuildImage.groovy

The changes are pushed my master, but they don't seem to appear here, I'm assuming it's because the PR is closed. The diff is here https://github.com/bmuschko/gradle-docker-plugin/compare/master...exFalso:master

exFalso:
Should I resubmit this PR?

cdancy:
@exFalso yeah resubmit PR

PR #635, opened 2018-07-28
Fully deprecate MAINTAINER instruction and replace with LABEL

bmuschko:
Uses quotes for values in Map-based instruction to allow for spaces.


bmuschko:
Some of the unit tests are currently failing. Bits of Dockerfile handling should be unified and simplified. I'll look into it.

bmuschko:
Removed the bits that weren't necessary for the deprecations.

cdancy:
Looks good from my end. It's nice to have you around knocking out these long standing issues. You should do this more often ;)

PR #563, opened 2018-03-24
Document using container names for management

mwhipple:
The current GroovyDoc for `DockerExistingContainer` does not mention the Docker provided support for using a container name instead of ID and an example allowing for managing  of a Docker container across Gradle invocations has been added.

Provides solution for case outlined in #562

Issue #599, opened 2018-06-15
Support autoRemove in DockerCreateContainer


mwhipple:
This was just merged to `docker-java` so mostly just added here to get it on the radar (and I may grab it if I have time).
https://github.com/docker-java/docker-java/pull/1050

### Expected Behavior

Analogous to using `docker run --rm` this provides automatic removal of a container after shutdown,
which is convenient for running local dev containers.

### Current Behavior

Currently this can be accomplished by an additional remove task triggered a stop, so this would just
remove a slight amount of noise.

### Context

The new functionality is not mapped onto the fluent CreateContainerCmd interface which is leaned on by the existing task (apparently due to the associated churn). A potential question would be whether to preserve the flatter model currently used in the plugin and have the task handle the delegation to the HostConfig, or whether to add the nested structure to more closely mirror docker-java. FWIW I'd vote moving the code in that direction and using delegation to flatten the exposed interface.

cdancy:
Was just thinking about this the other day and indeed we should get this in as I think it will be useful. 

@mwhipple you _may_ be interested in a recent plugin I put together, which builds upon this plugin, to allow developers/companies to create something like a `site` plugin for their dockerized apps or really any arbitrary dockerized application.

https://github.com/project-aries/gradle-docker-application-plugin

mwhipple:
I should be able to throw a PR together for this today (let me know if anyone else already started it).

bmuschko:
@mwhipple Go for it. Please also add a functional test. Thanks!

cdancy:
This will require a bump in the `docker-java-shaded` library. @mwhipple once I release a new version, which should be shortly, you can add that change to your PR.

bmuschko:
@cdancy Thanks for that pointer. I completely forgot about it. Can't wait to get onto the Spotify library. ;-)

cdancy:
@bmuschko at least our users are not hitting random classpath/cglib issues anymore so there's that :)

cdancy:
@mwhipple new version has been released. Thanks for the PR on that library as well ;)

PR #637, opened 2018-07-30
Add background interaction example to user guide

mwhipple:
Migration of #563 to user guide

### DIFF
--- a/src/docs/asciidoc/15-examples.adoc
+++ b/src/docs/asciidoc/15-examples.adoc
@@ -30,6 +30,52 @@ task buildImage(type: DockerBuildImage) {
 }
 ----
 
+==== Interacting With a Container Running in the Background
+
+This code demonstrates managing a container running in the background with this plugin
+which can facilitate locally running a Dockerized application. This can be done by
+creating the container with an explicit `containerName` which is then used as
+the `targetContainerId` for relevant tasks.
+
+.build.gradle
+[source,groovy]
+----
+apply plugin: 'com.bmuschko.docker-remote-api'
+
+import com.bmuschko.gradle.docker.tasks.container.*
+
+ext {
+  CONTAINER_NAME = 'myapp'
+} 
+
+task buildMyAppImage(type: DockerBuildImage) {
+    inputDir = file('docker/myapp')
+    tag = 'test/myapp:latest'
+}
+
+task dockerCreateContainer(type: DockerCreateContainer) {
+    dependsOn buildMyAppImage
+    targetImageId { buildMyAppImage.imageId }
+    containerName = CONTAINER_NAME
+}
+
+task dockerStartContainer(type: DockerStartContainer) {
+    dependsOn dockerCreateContainer
+    targetContainerId { dockerCreateContainer.getContainerName() }
+}
+
+task dockerStopContainer(type: DockerStopContainer) {
+    targetContainerId { dockerCreateContainer.getContainerName() }
+}
+
+task dockerRemoveContainer(type: DockerRemoveContainer) {
+    targetContainerId { dockerCreateContainer.getContainerName() }
+}
+
+// Remove the container to allow reuse of the name.
+dockerStopContainer.finalizedBy dockerRemoveContainer
+----
+
 ==== Executing Functional Tests Against a Running Container
 
 The following example code demonstrates how to build a Docker image from a Dockerfile, starts up a container for this image and exercises functional tests against the running container.
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExistingContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExistingContainer.groovy
@@ -23,7 +23,10 @@ import java.util.concurrent.Callable
 
 abstract class DockerExistingContainer extends AbstractDockerRemoteApiTask {
     /**
-     * Container ID used to perform operation. The container for the provided ID has to be created first.
+     * Container identifier upon which the operation will be performed.
+     *
+     * The ID or name of a container are supported as identifiers.
+     * The container for the provided identifier has to be created first.
      */
     @Input
     String containerId


src/docs/asciidoc/15-examples.adoc
```diff
@@ -30,6 +30,52 @@ task buildImage(type: DockerBuildImage) {
...
+
+.build.gradle
+[source,groovy]
+----
+apply plugin: 'com.bmuschko.docker-remote-api'
+
+import com.bmuschko.gradle.docker.tasks.container.*
+
+ext {
+  CONTAINER_NAME = 'myapp'
```
bmuschko: Extra properties are not really environment variables. Could you please use camel case notation here and use the 4 spaces indentation?

src/docs/asciidoc/15-examples.adoc
```diff
@@ -30,6 +30,52 @@ task buildImage(type: DockerBuildImage) {
 }
 ----
 
+==== Interacting With a Container Running in the Background
```
bmuschko: The title of this section doesn't seem to indicate an actual interaction with the container. From what I can see in the code example it rather shows a full lifecycle of a container: create, start, stop, remove. I think a different title may make more sense.

src/docs/asciidoc/15-examples.adoc
```diff
@@ -30,6 +30,52 @@ task buildImage(type: DockerBuildImage) {
...
+
+task buildMyAppImage(type: DockerBuildImage) {
+    inputDir = file('docker/myapp')
+    tag = 'test/myapp:latest'
+}
+
+task dockerCreateContainer(type: DockerCreateContainer) {
+    dependsOn buildMyAppImage
+    targetImageId { buildMyAppImage.imageId }
+    containerName = CONTAINER_NAME
```
bmuschko: Can you explain why using the container name is used (maybe even preferred) over the container ID in this example? Sure, you can use either or but why would a user want to go for the container name instead?

src/docs/asciidoc/15-examples.adoc
```diff
@@ -30,6 +30,52 @@ task buildImage(type: DockerBuildImage) {
...
+task dockerStopContainer(type: DockerStopContainer) {
+    targetContainerId { dockerCreateContainer.getContainerName() }
+}
+
+task dockerRemoveContainer(type: DockerRemoveContainer) {
+    targetContainerId { dockerCreateContainer.getContainerName() }
+}
+
+// Remove the container to allow reuse of the name.
+dockerStopContainer.finalizedBy dockerRemoveContainer
```
bmuschko: Can you explain the use case you are modeling here? Should the build _always_ stop and remove the container? Or would this be a use case where you start a container by running a distinct build and then may sometime later stop and remove the container but manually interacting with the container? It's not super clear from the description. We should make it more clear.

src/docs/asciidoc/15-examples.adoc
mwhipple: Sounds reasonable. How about `Creation and Management of a Background Container for Local Development`

src/docs/asciidoc/15-examples.adoc
mwhipple: sure, did you want that as a comment here or included in the docs?

The reason is that the containerId will be unique per created container and therefore would need to be tracked when interacting with a container that is started in the background (i.e. the containerId would normally be available over the course of a single invocation of gradle and therefore not usable in a subsequent invocation to do something like stop a running container). Additionally in this use case the containers are ephemeral and therefore the identifier is likely reusable. So: the idea of eagerly feeding in a defined identifier rather than a generated unique identifier allows for a simple reliable handle for the resulting container.

src/docs/asciidoc/15-examples.adoc
mwhipple: I'll look at tightening up the description of the use case. This bit is a workaround for `autoRemove` (https://github.com/bmuschko/gradle-docker-plugin/issues/599) which is all part of a way to manage ephemeral development containers.

bmuschko: Thanks, I think letting users understand what problem were are trying to solve with the code example is very important. If this is more of a workaround then we should fix the underlying issue instead (if possible).

src/docs/asciidoc/15-examples.adoc
bmuschko: Thanks, let's start with describing the use case first e.g. with a concrete example (not the code).

src/docs/asciidoc/15-examples.adoc
mwhipple: if I have time this week I'll look at grabbing that issue.

bmuschko: @mwhipple Can we close this PR given that it is a workaround for not having `autoRemove`?

mwhipple:
This PR is still valid: the important bit is the user provided name/identifier. The autoRemove addition just avoids the need for the finalizedBy workaround. I'll try to improve the docs this week (though I unfortunately document like a developer)

cdancy:
@mwhipple is this PR still valid in your opinion now that we've got a fix in place and merged to master?

mwhipple:
Here's my take and I'd look for feedback on how/if to integrate the info:

`autoRemove` is only a detail, the key piece in this PR is that the container can be interacted with using a provided name rather than the generated id. Rather than a recipe a note to that effect is likely sufficient. Here's a small brain dump that outlines some use cases that I now think fizzle down to that idea:

The reuse of the same identifier can be helpful in general when developing an app and only caring about the latest build of that app, but maybe looking to repeatedly do things to the running container that corresponds to that latest build while developing (such as starting the container and attaching to that container); this is also useful when managing containerized dependencies since it provides a simple handle to use to start, query, and stop containers while also protecting against duplicated running instances.

The intent of this PR was the start, query, stop lifecycle stuff for the app under development where for instance you could run something like `./gradlew dockerStartContainer` -> do some local testing -> `./gradlew dockerStopContainer` and have the application running in the background while the local testing was being done (and the generated id would be lost between those two gradle invocations). However, I question whether this is a practice that is worth explicitly documenting since it would likely be simpler and more consistent with the application plugin to just run the container in the foreground.

Another motivation for the general pattern could be to use this plugin to manage other containers such as those used for dependencies (so for instance have an invocation which results in steps being executed such as dockerStartDBContainer -> startApp -> dockerStopDBContainer), but that seems enough outside of the core purpose of this plugin that documentation may just amount to noise.

All of these uses may be enough of a stretch that they don't warrant any kind of first class documentation but the pointer to using the container name may help anyone looking to do something similar with the rope they need.

bmuschko:
+1 I don't think it's a common use case to go into interactive mode with the code you have in a Gradle build script. In those case I'd probably just use the Docker CLI directly.

I think the important takeaway is that you can assign a "name" to a container and use that to refer to it. At the moment the naming of the method `targetContainerId` is not appropriate as it also works with the name. From my perspective it's better to improve on that front. I propose that we introduce a replacement method and deprecated the existing one + add documentation for it in the form of Javadocs.

I think we can close this PR if you agree.

mwhipple:
Sounds good to me, should I open an issue to make sure that doesn't get lost?

bmuschko:
Yes, please. Thanks!

PR #639, opened 2018-08-02
Support autoRemove for containers

### LINKED ISSUE #599: Support autoRemove in DockerCreateContainer

mwhipple:
Why is this feature beneficial?

Helps avoids accumulation of containers and facilitates name reuse.

Closes #599


mwhipple:
Based on the PR template I'll add another test for when autoRemove isn't specified

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -177,6 +177,13 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
     @Optional
     Long shmSize
 
+    /* 
+     * Automatically remove the container when the container's process exits.
+     * This has no effect if RestartPolicy is set.
+     */
+    @Input
+    Boolean autoRemove = false
```
mwhipple: I talked myself into the wrong direction I think: I'll change the default to null and add the condition below

mwhipple:
This should have corresponding doc updates, I'll work on that later today

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -398,6 +406,9 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         if(getShmSize() != null) { // 0 is valid input
             containerCommand.hostConfig.withShmSize(getShmSize())
         }
+        if (getAutoRemove() != null) { // explicit false is valid
```
cdancy: You're probably just following convention but you don't need to do the explicit null check as groovy will handle things. A simple `if (getAutoRemove()) { ... }` should suffice.

mwhipple: I was avoiding groovy truthiness in case someone explicitly sets false and the default behavior of the upstream libraries change (not that that seems likely)

cdancy: Ok fair enough.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -177,6 +177,14 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
     @Optional
     Long shmSize
 
+    /* 
+     * Automatically remove the container when the container's process exits.
+     * This has no effect if RestartPolicy is set.
+     */
```
bmuschko: Would should indicate the version that introduces the new property with `@since`.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -133,4 +134,94 @@ class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
         result.output.contains("seven=eight")
         result.output.contains("nine=ten")
     }
+
+    def "Set autoRemove"() {
```
bmuschko: Let's be more explicitly about the expected behavior in the title. Same for the title test case below.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -177,6 +177,14 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
     @Optional
     Long shmSize
 
+    /* 
+     * Automatically remove the container when the container's process exits.
+     * This has no effect if RestartPolicy is set.
```
bmuschko: Can you please link to the `restartPolicy` here via `{@link #restartPolicy}?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -16,6 +16,7 @@
 package com.bmuschko.gradle.docker.tasks.container
 
 import com.bmuschko.gradle.docker.AbstractFunctionalTest
+import org.gradle.api.tasks.TaskExecutionException
 import org.gradle.testkit.runner.BuildResult
 
 class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
```
bmuschko: @cdancy In general would should externalize some of the fixture logic. This gets copied and over and over again.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -133,4 +134,94 @@ class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
...
+                dependsOn stopContainer
+                targetContainerId { createContainer.getContainerId() }
+            }
+            
+            task workflow {
+                dependsOn inspectContainer
+            }
+        """
+
+        expect:
```
bmuschko: I'd prefer `when`/`then` here.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -133,4 +134,94 @@ class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
...
+        BuildResult result = buildAndFail('workflow')
+        result.output.contains('No such container')
+    }
+
+    def "Unset autoRemove"() {
```
bmuschko: I don't think we are unsetting the value here. It simply runs with the default value `null`. Let's be more explicitly about the test case title e.g. "Does not automatically remove container if auto-remove wasn't enabled".

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -16,6 +16,7 @@
 package com.bmuschko.gradle.docker.tasks.container
 
 import com.bmuschko.gradle.docker.AbstractFunctionalTest
+import org.gradle.api.tasks.TaskExecutionException
```
bmuschko: This is an unused import.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
mwhipple: This was trying to follow convention :) I'll update them with a more descriptive spec form.

bmuschko: The titles of existing test cases are not necessarily very expressive and should be changed as well. We'll clean up a bunch of existing code over the next couple of weeks.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
bmuschko: I committed a change that simplifies the logic in this test class.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
bmuschko: Titles have been changed on `master`.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
cdancy: I wonder if we could do the same but in a more global manner so that all tests clean up after themselves. That way no one will have to worry about tearing down a container it will simply just happen as part of setting up the test.

mwhipple: Is that #383 ?

mwhipple:
I should be able to get this cleaned up today (pesky day job getting in the way)

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
bmuschko: I guess we could but it would require specific naming standards when creating images and containers so we can identify what to prune. We should also keep the base images. Those can be kept in between builds so we don't have to re-download them every time.

I could image it to work the following way:

1. When creating images in a functional test assign the label `gradle-docker-plugin`.
2. When running containers in a functional test assign the label `gradle-docker-plugin`.
3. Issue a `docker container prune --filter="label=gradle-docker-plugin"` and `docker image prune --filter="label=gradle-docker-plugin"` command in a `finalizedBy` task for `functionalTest`. I don't think we need to run this through the Docker Java API as it is more of a build management functionality.

mwhipple:
Have another commit in flight to align with the new functest stuff, also removing the disabled case since it seems very redundant without the boilerplate

mwhipple:
all set for re-review

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -75,6 +75,24 @@ class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
...
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId { pullImage.getImageId() }
+                autoRemove = true
+            }
+        """
+        buildFile << containerUsage(containerCreationTask)
+
+        when:
+        BuildResult result = buildAndFail('logContainer')
```
bmuschko: This test case shouldn't actually fail. What I'd like to see here is a test case that doesn't use and wire the `removeContainer` task that is already set up in the method `containerUsage`. It will mean splitting up the fixture a little more. The test case you have here might be another case but I don't think it's necessary to cover it explicitly. I'd hope for Docker Java to handle the situation properly.

mwhipple: Are there any tasks that can query for container presence without expecting them to exist? I couldn't quickly find any: I can wire it up more directly with docker-java but it will delay it a bit.

cdancy: You could use the DockerClient task which hands you back the raw client and from there you could query things. Though I do think a DockerRemoveContainer, with an onError closure defined to catch the non existent container, seems like a sane path forward here.

bmuschko: I also think `DockerInspectContainer` is probably the best option here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -177,6 +177,16 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
     @Optional
     Long shmSize
 
+    /* 
+     * Automatically remove the container when the container's process exits.
+     *
+     * This has no effect if {@link #restartPolicy} is set.
+     * @since 3.6.0
```
bmuschko: At the moment the next version is probably going to be `3.5.1` though I think we should bump the [existing changes](https://github.com/bmuschko/gradle-docker-plugin/milestone/70?closed=1) to 3.6.0. @cdancy Do you agree?

cdancy: Agreed.

bmuschko: @cdancy I changed the milestone to 3.6.0.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
bmuschko: @mwhipple Do you think you'll get to touching up the PR soon? Would love to release a new version.

mwhipple:
I'll work on getting this cleaned up shortly.

mwhipple:
Split up the existing test stuff to tie into it, handled the error, and added the other scenario back since the wiring was different enough. Let me know if I broke some coding conventions (codenarc looks fine). I'll try to get a doc update together later today.

bmuschko:
Nice, thanks! Will review soon.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -398,6 +408,9 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         if(getShmSize() != null) { // 0 is valid input
             containerCommand.hostConfig.withShmSize(getShmSize())
         }
+        if (getAutoRemove() != null) { // explicit false is valid
+            containerCommand.hostConfig.withAutoRemove(autoRemove)
```
bmuschko: This should also use `getAutoRemove()`.

bmuschko:
Looks like some tests are failing. Could you please make CI pass for your changes? Does it pass if you run `./gradlew fT --tests com.bmuschko.gradle.docker.tasks.container.DockerCreateContainerFunctionalTest` locally?

mwhipple:
It passes locally, I'll check out the travis results

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -75,13 +79,77 @@ class DockerCreateContainerFunctionalTest extends AbstractFunctionalTest {
...
+                autoRemove = true
+            }
+        """
+
+        String containerInspect = """
+            task inspectStoppedContainer(type: DockerInspectContainer) {
+                dependsOn stopContainer
+                targetContainerId { startContainer.getContainerId() }
+
+                onError = { err -> println 'RESULT: ' + err }
```
mwhipple: looks like this should likely be `onError {` rather than the direct property assignment and that may also allow use of  a GString (which was giving issues)

mwhipple:
the failure is something I get locally with a hang in `com.bmuschko.gradle.docker.tasks.DockerReactiveMethodsFunctionalTest` but I thought it was just my computer. Maybe it's a result of the bump in the docker-java, I'll poke around.

bmuschko:
You may want to run all functional tests without any of your changes except the new version of Docker Java just to see if that's the issue.

mwhipple:
The specific line that's causing the hang appears to be: https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerReactiveMethodsFunctionalTest.groovy#L352

mwhipple:
replacing that line to a log message and replacing `buildAndFail` with just `build` does the trick...is there a specific reason for the thrown exception there?

bmuschko:
> replacing that line to a log message and replacing buildAndFail with just build does the trick...is there a specific reason for the thrown exception there?

@cdancy @orzeh Can you comment on this?

cdancy:
Looks like we expect the container start to fail due to the `1` exit code being returned. If this is failing on your branch I suspect it must be failing on master as well?

mwhipple:
seems to be fine on master based on CI builds. I can poke around a bit more in a couple hours (and try the suggestion of isolating the docker-java bump).

cdancy:
@bmuschko we can do that. We can always easily do a patch release when this goes in. Did you want to kick things now? Judging by various comments I've seen it would appear you are eager to get things out of the door :)

bmuschko:
@cdancy Would you mind doing it? I'd have to read the release docs first and need to run in 2 mins. I can do the next release.

cdancy:
@bmuschko not a problem at all and in fact I like doing them so that I know exactly what is going out.

mwhipple:
Confirmed that the test passes on a clean master branch but hangs after bumping the docker-java plugin

mwhipple:
When adding a timeout to the wait task it passes...but after fudging the assertion to see the output I see:
```
     > Task :createContainer
|      Created container with ID '4bb740d684a3e6fd57307e31be096c55bcd2a7d33f3a67d599a5d45be32210cd'.
|       
|      > Task :startContainer
|      Starting container with ID '4bb740d684a3e6fd57307e31be096c55bcd2a7d33f3a67d599a5d45be32210cd'.
|      Error during callback
|      org.gradle.api.GradleException: Container failed!
...
```
so it looks like the onNext is ending up getting attached to the startContainer task.

mwhipple:
The behavior appears to be consistent in both versions: on rc3 the runContainers (wait task) just fails quickly with the error attached to startContainer, on rc4 the wait task blocks indefinitely unless there's a timeout.

mwhipple:
It appears that the hanging is caused by the `completed` countdown latch never being zeroed within the WaitContainerResultCallback, ultimately blocking on this line: https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/core/async/ResultCallbackTemplate.java#L89

That will be decremented  by the `startContainer` events but the order of evaluation must have switched so in `rc3` the required calls to `#close` happen prior to awaitStatusCode so the latch is open while in `rc4` the await happens first which blocks the thread (which is presumably the thread that would later result in the close calls). I'll keep working on isolating the specific change that causes this.

bmuschko:
@mwhipple Did you make anymore progress on identifying the root cause?

mwhipple:
Not yet, my spare time got pulled to something else. I should be able to circle back to this this week.

bmuschko:
@mwhipple Great, looking forward to getting this issue fixed. Even without your changes we seem to be stuck with the current Docker Java version without identifying the root cause.

santhoshk2311:
@mwhipple Gradle docker plugin latest release publish to jcenter/bintray is blocked by this.
Any idea when this issue will get fixed?. Thanks

cdancy:
@santhoshk2311 I have a workaround I'm putting together today. If everyone signs off I'll merge that and release and otherwise we will have to revert some things as this has proven to be a tricky issue.

cdancy:
@bmuschko what do you think about pulling this in for the new release? Anything outstanding we still want to do here? Now that we have a fix in place I can massage this branch and fix conflicts if only to get us another feature to add to the new release.

bmuschko:
@cdancy Yes, I think it's a good idea to merge this as well. It was also waiting on the fix with the hanging test and it's a good flag to have.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
bmuschko: @cdancy This should be 3.6.2.

cdancy: On it

bmuschko:
@cdancy Looks like this new test cases is broken on CI.

cdancy:
@bmuschko we just can't win. It works locally and on **this** branch here but failed on master? I'll give things another look/run and see if I can spot anything.

bmuschko:
@cdancy Only fails if you run the full test class.

```
./gradlew clean fT --tests com.bmuschko.gradle.docker.tasks.container.DockerCreateContainerFunctionalTest
```

The fix is the following:

```
if (getAutoRemove()) {
    containerCommand.hostConfig.withAutoRemove(getAutoRemove())
}
```

Before it was only checking for not-null for not for `true`.

cdancy:
@bmuschko is the `fT` something new? Why wouldn't the full tests run automatically anyway and if they are not running what is?

santhoshk2311:
@cdancy Looks like the CI issue is fixed. Any idea when to expect the new release published to Jcenter/bintray?

cdancy:
@santhoshk2311 just kicked release now. Assuming all goes well it should be available within the next 15 minutes or so.

cdancy:
Version `3.6.2` is now available.

mwhipple:
Awesome, thanks! Sorry for not getting the chance to dig deeper into the semaphore problem: mostly the combination of a teething baby and probably also the hope that the problem would be avoided rather than solved 😄

PR #640, opened 2018-08-04
Introduce explicit declaration that task supports onNext callback

bmuschko:
Before it was unclear if the logic in onNext is actually evaluated by a task without having a look at the code. Task now express this automatically if they implement the trait ResultCallback.

Furthermore, simplifies the inheritance hierarchy by removing the unnecessary parent class. All tasks of this plugin had the same handling anyway.

The overall behavior for end users stays the same.


bmuschko:
@cdancy There are two aspects to this change.

1. Pulling the logic of `AbstractReactiveStreamsTask` down into `AbstractDockerRemoteApiTask`. There was no point of having the class as all tasks in this plugin extend from `AbstractDockerRemoteApiTask`. The logic uses the Closures `onError` and `onComplete` if they have been provided. Result: Simplifies the class hierarchy and less indirection.
2. The handling of the callback `onNext` is special. Some tasks use it, some don't. It's not clear to end users which ones actually handle the callback if a Closure has been provided. You can only know if you look at the code. From a user's perspective that's extremely confusing. The new trait clearly indicates which of the task handle the Closure. For all other tasks that do not implement the trait, a user cannot even provide a Closure. From my perspective that is much clearer than a default implementation that does nothing. What would a default implementation even do e.g. for `DockerStopContainer` if no callback exists in Docker Java?

cdancy:
If it's not clear than I would say we need to rework it, or have all tasks support it if only in a limited fashion (which I'm in favor of), or fix the documentation but it seems logically wrong to separate things out whilst also changing the naming convention.

For a default implementation we could simply return a boolean, which is harmless in its own right, or just invoke it passing in nothing. I'd prefer that over having only some do things while others don't if only to make it work across the board.

bmuschko:
> it seems logically wrong to separate things out whilst also changing the naming convention.

What's the change in naming convention?

> For a default implementation we could simply return a boolean, which is harmless in its own right, or just invoke it passing in nothing. I'd prefer that over having only some do things while others don't if only to make it work across the board.

From my perspective this is simple OO-design. If an implementation doesn't support specific functionality then there's no point in faking it just to fulfill the contract. Example: You have an `Animal` interface and a `FlyingAbility` interface. Some animals simply don't fly and therefore don't implement the `FlyingAbility` interface. I see it them same way here. We have tasks handling Docker operations. Not all of them handle data returned from the Docker daemon and therefore don't need to implement `doNext`. It serves as documentation and clearly tells the user when to expect data that can be handled.

In the end you are maintaining the code base so I'll let you do the final call. From a user's perspective it's confusing at the moment. For now I'd propose to just remove `AbstractReactiveStreamsTask`. There's no point in having this class. As a build script developer I'd ask myself which one to extend (`AbstractReactiveStreamsTask` vs. `AbstractDockerRemoteApiTask`) if I wanted to add functionality or which one to react to if I wanted to have unified logic (`tasks.withType()`).

bmuschko:
@cdancy Let me know what the decision will be on this.

bmuschko:
@cdancy @orzeh Ping! Can we come to a decision on this matter one way or another?

bmuschko:
@cdancy With "change" you mean just the removal of `AbstractReactiveStreamsTask` right?

cdancy:
@bmuschko not exactly but its fine. My "gripes" were more philosophical and I don't hold to these particularly ones strongly enough to stop the PR from moving forward.

And my apologies for not getting back to this sooner. I actually had this PR opened in a browser tab on my home machine for the longest time to remind me to sit and review it more thoroughly I just haven't had the time recently and it kept slipping my mind. I'm fine with things as they are now.

bmuschko:
I'll do this separately after merging the https://github.com/bmuschko/gradle-docker-plugin/pull/659.

PR #644, opened 2018-08-10
Change DockerClient to be cached by configuration

### LINKED ISSUE #643: Unable to configure different Docker instances per task
```groovy
    url = 'unix:///var/run/docker.sock'
    certPath = null
}
task pushImage(type: DockerPushImage) {
    url = 'unix:///var/run/docker.sock'
    certPath = null
}
    url = 'docker-ucp.corp.com'
    certPath = '/etc/docker/corp.com'
}
```

### Expected Behavior
Image is built and pushed using the local Docker daemon. Container is created in the remote Docker UCP instance.

### Current Behavior
Image is built and pushed using the local Docker daemon. Container is created in the local Docker daemon.

### Context
We can't build and push using a remote Docker EE instance because the build and push calls can be run on different nodes. We'd like to build and push in the local Docker daemon and run on the remote.

See above.

### Your Environment
Docker EE install.
Build machine has local Docker daemon and access to Docker EE.

### Issue
In `DockerRemoteApiPlugin`, a single `DockerThreadContextClassLoader` instance is used for all tasks. `DockerThreadContextClassLoader` caches `DockerClient` without considering differing configuration.

### Recommendation
I recommend memoizing the `DockerClient` based on the configuration.

I'm willing to submit a PR.

double16:
One DockerClient was created per DockerThreadContextClassLoader. If Docker tasks have differing configurations, only the first would apply. This change caches the DockerClient based on the configuration to correctly support tasks with differing configurations.

closes #643


cdancy:
@double16 nice and clean I like it. Looks like the test failures are maybe un-related to this change?

double16:
```
> The whole test suite should be executed. The test com.bmuschko.gradle.docker.DockerSpringBootApplicationPluginFunctionalTest.Can create image for a Spring Boot application and push it to DockerHub [#plugin.identifier plugin] was skipped. Please check the execution condition!
```

I can't see how these changes would cause this. It'd be interesting to see if a build on master would fail this way.

double16:
@cdancy Is there anything I need/can do to move this forward?

cdancy:
@double16 let me give things a go here at my desk. Stay tuned...

@bmuschko @orzeh any thoughts one way or another?

cdancy:
Master passes locally as well as this branch. Let me try re-kicking things and see what happens.

bmuschko:
I'll try to look at it later today or tomorrow.

RE the CI error. It seems as if the branch build doesn't pick up all the environment variables necessary needed for pushing to Docker Hub. It'll take a look.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+class DockerMultipleClientFunctionalTest extends AbstractGroovyDslFunctionalTest {
+    def "DockerClient respects task configuration"() {
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.DockerClient
+
+            task dockerClient1(type: DockerClient) {
+                onNext { client ->
+                    if (client != null) {
+                        logger.quiet "config: " + client.dockerClientConfig.toString()
+                    } else {
```
bmuschko: I think we can just remove the `else` blocks and assert the "config" message. The "config" message already verifies that the client is not null.

cdancy:
FYI: version `3.6.1` has been released with this change.

PR #645, opened 2018-08-15
When parsing deprecated envs property ensure we limit our response to…

Psylofreun:
… 2 to preserve backward compatibility and user expectations.

Please do not create a Pull Request without first creating an ISSUE.
Any change needs to be discussed before proceeding.
Failure to do so may result in the rejection of the Pull Request.

Explain the **details** for making this change: What existing problem does the Pull Request solve? Why is this feature beneficial?

**On adding new feautes/endpoints**

No more than 1 endpoint should be coded within a Pull Request. 
This alone requires enough code to review and adding more than 1 WILL result in your Pull Request either being ignored or rejected outright.

**On adding Mock and Integ Tests**

At _least_ 2 mock tests and 2 integ tests are required prior to merging. 
Each pair should should test what the success and failure of added change looks like.
More complicated additions will of course require more tests.

Current CI does NOT exercise `integ` tests and so each Pull Request will have to be run manually by one of the maintainers to confirm it works as expected: please be patient.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -290,7 +290,7 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
 
         // marshall deprecated old list onto new map
         getEnv()?.each { envVar ->
-            def keyValuePair = envVar.split('=')
+            def keyValuePair = envVar.split('=', 2)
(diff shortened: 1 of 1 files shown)


bmuschko:
Could you please explain what backward compatibility we are talking about here and why this change is necessary? Could you please also point me to the test that verifies the proper behavior? Please add a new test if there's no test for it yet.

bmuschko:
It's unclear why you'd like this change. Unfortunately, we didn't receive a response. I am going to close this pull request for now. Let us know if you'd like to continue the discussion.

PR #646, opened 2018-08-15
Support for --cache-from

exFalso:
Reopen of https://github.com/bmuschko/gradle-docker-plugin/pull/629, issue https://github.com/bmuschko/gradle-docker-plugin/pull/628

dockerjava seems to doubly-parse this argument, the workaround is to doubly-serialize.

Added three tests, testing the following:

| --cache-from flag present? | layer to be built in cache? | result                          |
|----------------------------|-----------------------------|---------------------------------|
| true                       | false                       | build success, cache isn't used |
| false                      | true                        | build success, cache isn't used |
| true                       | true                        | build success, cache is used    |

It's a bit tricky to populate the cache without "linking" it to local builds, the way the tests do it is they build a layer, push it to the registry, then delete the layer locally, and finally pull the layer. This populates the cache but removes the build "link".

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -160,8 +160,10 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             buildImageCmd = buildImageCmd.withBuildArg(arg, value)
         }
 
-        if (!cacheFrom.isEmpty()) {
-            buildImageCmd = buildImageCmd.withCacheFrom(cacheFrom)
+        if (getCacheFrom()) {
+            // Workaround a bug in dockerjava that double-unmarshalls this argument
+            def doubleMarshalledCacheFrom = ["[\"${getCacheFrom().join('","')}\"]".toString()].toSet()
+            buildImageCmd = buildImageCmd.withCacheFrom(doubleMarshalledCacheFrom)
         }
 
         def callback = onNext ? threadContextClassLoader.createBuildImageResultCallback(this.onNext)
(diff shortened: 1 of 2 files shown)


exFalso:
CI seems to be failing because of an unexpected skip of a `DockerSpringBootApplicationPluginFunctionalTest` test. The precondition checks for docker hub credentials availability. I don't think this PR could affect this. Any ideas?

cdancy:
@bmuschko was taking a look at this elsewhere for #644 . I'll run things locally and if all goes as expected I'll merge this.

cdancy:
All tests pass locally. I'll wait a bit and see if these guys want to chime in one way or another but LGTM.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -156,6 +160,12 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             buildImageCmd = buildImageCmd.withBuildArg(arg, value)
         }
 
+        if (getCacheFrom()) {
+            // Workaround a bug in dockerjava that double-unmarshalls this argument
```
bmuschko: Did you report this bug to Docker Java? I'd like to avoid adding workarounds in this plugin.

cdancy: Definitely open a big with those guys but I don't personally think this should stop things from moving forward. And given that they do releases once a blue moon I don't think we should hold things up on our end.

On a side note: we've been using spotifys docker-client here internally for writing a Jenkins shared library and it's really easy and great IMO. Its giving me more hope that switching things on our end wouldn't be so bad.

exFalso: I submitted a fix https://github.com/docker-java/docker-java/pull/1086

exFalso:
Is there anything else I can do for this PR?

cdancy:
@exFalso no I think you're good. The change is on their impl, which we are tied to, and so don't need to worry about much else. You've already opened a bug so all we can do is wait. I'll take this in and look to kick a release.

cdancy:
FYI: version `3.6.1` has been released with this change.

exFalso:
@cdancy Note on this, the fix is now merged upstream in docker-java, so when the next version is released and the dependency bumped to `3.1.0` it should break these tests, and the workaround can be deleted

PR #652, opened 2018-08-31
ISSUE-651: implemented support for --pid option in DockerCreateContai…

### LINKED ISSUE #651: support --pid in DockerCreateContainer
There is no support for the --pid flag (docker create).

### Expected Behavior
It could look something like:
task createExpertTestContainer(type: DockerCreateContainer) {
    targetImageId { "imagename:latest" }
    portBindings = ["8082:8080"]
    containerName = "containerName"
    **pid = "host"**
    tty = true
}

### Current Behavior
There is no property for pid mode in the DockerCreateContainer-Task and so in DockerCreateContainer.setContainerCommandConfig the method withPidmode() from CreateContainerCmd (docker-java) is never called.

### Context
I wanted to create a container with --pid=host

marcoebbinghaus:
adds support for --pid option in docker create / DockerCreateContainer-Task

closes #651

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -457,6 +457,45 @@ class DockerWorkflowFunctionalTest extends AbstractGroovyDslFunctionalTest {
         result.output.contains("RestartPolicy : on-failure:999")
     }
 
+    def "Can build an image, create a container and set pid"() {
+        File imageDir = temporaryFolder.newFolder('images', 'minimal')
+        File dockerFile = createDockerfile(imageDir)
+
+        String uniqueContainerName = "trololo"
+
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+            import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+            import com.bmuschko.gradle.docker.tasks.container.DockerInspectContainer
+
+            task buildImage(type: DockerBuildImage) {
+                inputDir = file("${dockerFile.parentFile.path}")
+                tag = "${createUniqueImageId()}"
+            }
+
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn buildImage
+                targetImageId { buildImage.getImageId() }
+                containerName = "${uniqueContainerName}"
+                pid = "host"
+                cmd = ['/bin/sh']
+            }
+
+            task inspectContainer(type: DockerInspectContainer) {
+                dependsOn createContainer
+                targetContainerId { createContainer.getContainerId() }
+            }
+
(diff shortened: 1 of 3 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -231,6 +235,10 @@ class DockerCreateContainer extends AbstractDockerRemoteApiTask {
         restartPolicy = "${name}:${maximumRetryCount}"
     }
 
+    void pid(String name) {
+        pid = "${name}"
+    }
```
cdancy: We generally don't use methods unless there is a good reason. What's the thinking here?

bmuschko: Let's remove this method. Groovy automatically generates a setter method.

marcoebbinghaus: Oh good to know. Yes I will remove that method then. Should I push it as a 2nd commit or should I commit locally, squash both commits and force push it into the pr/651 branch so that it remains one commit/push on pr/651? 

And I don't get why the Travis CI build failed.
It says "Caused by: org.gradle.api.GradleException: The whole test suite should be executed. The test com.bmuschko.gradle.docker.DockerSpringBootApplicationPluginFunctionalTest.Can create image for a Spring Boot application and push it to DockerHub [#plugin.identifier plugin] was skipped. Please check the execution condition!" .. do I have any influence on that when pushing?

cdancy: Ignore the Travis CI failure for now as it's a known issue. I would send in another commit with the fix.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
```diff
@@ -457,6 +457,45 @@ class DockerWorkflowFunctionalTest extends AbstractGroovyDslFunctionalTest {
         result.output.contains("RestartPolicy : on-failure:999")
     }
 
+    def "Can build an image, create a container and set pid"() {
```
bmuschko: Can we please move this test into `DockerCreateContainerFunctionalTest.groovy`? You will find that there's plenty of setup code already you can reused. I want to remove this class in the long run and distribute non-duplicate test cases to the relevant test classes.

marcoebbinghaus: I moved the test into DockerCreateContainerFunctionalTest.groovy

marcoebbinghaus:
Thank you for merging @cdancy ! It's okay if it will be in 3.6.2, I don't need a patch release.

PR #655, opened 2018-09-11
Have DockerRegistryCredentials implement PasswordCredentials

### LINKED ISSUE #653: DockerRegistryCredentials should implement org.gradle.api.credentials.PasswordCredentials
### Current Behavior
Currently `DockerRegistryCredentials` doesn't implement the `PasswordCredentials` interface.

### Context
I have a general purpose login configuration extension I use across my build that I access like this:
```kotlin
val configureCompanyCredentials: (org.gradle.api.credentials.PasswordCredentials) -> Unit by rootProject.extra
```

Not implementing the interface, in order to work with this code, I need to create my own adapter class:
```kotlin
class DockerRegistryCredentialsAdapter(
    private val registryCredentials: DockerRegistryCredentials
) : org.gradle.api.credentials.PasswordCredentials {
    override fun setUsername(userName: String?) {
        registryCredentials.username = username
    }

    override fun getUsername(): String = error("Not needed")

    override fun getPassword(): String = error("Not needed")

    override fun setPassword(password: String?) {
        registryCredentials.password = password
    }
}
```

It would be nice to not have to do this.

### Your Environment
```
id("com.bmuschko.docker-remote-api") version "3.6.1"

Gradle 4.9
```

JLLeitschuh:
Closes #653

Signed-off-by: Jonathan Leitschuh <Jonathan.Leitschuh@plexxi.com>

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRegistryCredentials.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRegistryCredentials.groovy
@@ -15,10 +15,13 @@
  */
 package com.bmuschko.gradle.docker
 
+import groovy.transform.CompileStatic
+import org.gradle.api.credentials.PasswordCredentials
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 
-class DockerRegistryCredentials {
+@CompileStatic
+class DockerRegistryCredentials implements PasswordCredentials {
     public static final String DEFAULT_URL = 'https://index.docker.io/v1/'
 
     /**


src/main/groovy/com/bmuschko/gradle/docker/DockerRegistryCredentials.groovy
```diff
@@ -15,10 +15,13 @@
...
 package com.bmuschko.gradle.docker
 
+import groovy.transform.CompileStatic
+import org.gradle.api.credentials.PasswordCredentials
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 
-class DockerRegistryCredentials {
+@CompileStatic
+class DockerRegistryCredentials implements PasswordCredentials {
```
bmuschko: Thanks for providing the pull request. I think it would make sense to add a test case that shows the problem it solves. Is it correct to assume that it is required for using the plugin from the Kotlin DSL? I recently broke out functional test fixtures for Kotlin-based builds so adding a test case shouldn't be too hard.

JLLeitschuh:
@bmuschko You can see the use case in the linked issue: #653

What kind of test do you want? Just an `instanceof` test? That doesn't seem all that valuable.

bmuschko:
@JLLeitschuh Nah, had a deeper look at your change. LGTM as-is. Thanks for your pull request. I am good with the change. `PasswordCredentials` is part of Gradle's public API so I don't expect a breakage on our end in the near future.

JLLeitschuh:
@bmuschko Thanks!! 😄

bmuschko:
We might have to revert this change. The interface would force us to use a `String` as return type for the interface methods. 

```
String getPassword()
String getUsername()
void setPassword(String password)
void setUsername(String userName)
```
I am currently converting the plugin API to the `Provider` API and that would potentially prevent us from missing out on some of the lazy evaluation as we have to fulfill the contract imposed by the interface. This is how it will look like:

```
Property<String> getPassword()
Property<String> getUsername()
void setPassword(Property<String> password)
void setUsername(Property<String> userName)
```

I will let you know as soon as I convert the relevant pieces.

JLLeitschuh:
@big-guy or @oehme do you have any suggestions here?

JLLeitschuh:
I'm just aiming to make the API more flexible by providing built in adapters from your types to the more universally used interfaces that the Gradle API exposes already.
I have a lot of code that expects the `PasswordCredentials` interface and it would be nice to not have to write those adapters inside of your build but instead have them offered by the plugin.

Maybe it would be better for Gradle to offer a `ProviderBasedPasswordCredentials` interface.

bmuschko:
@JLLeitschuh Totally get it. It would make your life much easier. I am going to look at this but in a separate PR once we merge the provider change. Maybe the effort isn't that big. Would you mind opening an issue for Gradle core requesting a change to the existing interface?

JLLeitschuh:
Sure.

JLLeitschuh:
@bmuschko
https://github.com/gradle/gradle/issues/6943

bmuschko:
@JLLeitschuh Thanks for opening the issue. I removed your changes for now. It would break consumers immediately as soon as we merge the `Provider` API. We'll likely add an adapter implementation as proposed with the next release.

JLLeitschuh:
Thanks!!

Issue #654, opened 2018-09-05
Convert task inputs to new Gradle Provider API


JLLeitschuh:
https://docs.gradle.org/current/userguide/lazy_configuration.html

### Expected Behavior
All fields on all tasks need to support being configured lazily.

### Current Behavior
Currently, only some of the fields are settable lazily. For example, `DockerTagImage::repository` is missing a lazy configuration API.

### Context
The current way of setting fields that come from other fields is a very convoluted mix of `afterEvaluate` inside of task configurations.
For example:
```kotlin
val tagForCompany =
    tasks.register("dockerTagForCompanyUpload", DockerTagImage::class.java) {
        afterEvaluate {
            repository = "$companyDockerRepositoryUrl/${buildImage.tag.stripVersion()}"
            imageId = buildImage.tag
            tag = version.toString()
        }
        dependsOn(buildImage)
    }

tasks.register("dockerPushCompanyImage", DockerPushImage::class.java) {
    registryCredentials = docker.registryCredentials
    afterEvaluate {
        imageName = "$companyDockerRepositoryUrl/${tagForPlexxi.get().imageId.stripVersion()}"
    }
    tag = version.toString()

    dependsOn(tagForCompany)
}
```

### Your Environment
Gradle 4.9.
Plugin version 3.6.1

bmuschko:
Thanks for bringing this up. Closing as a duplicate of https://github.com/bmuschko/gradle-docker-plugin/issues/561. I am currently working on a first step toward it.

PR #659, opened 2018-09-22
Use Provider concept throughout to support lazy evaluation via public API

bmuschko:
- Introduces Provider API in convention plugins, their extensions and the custom task type `Dockerfile`.
- Removes methods taking `Closure` parameters and replace with `Provider`. It will require some changes in end user build scripts.
- Remove deprecations in extensions and `Dockerfile` custom task type.
- Next up: Change all custom task types to use `Provider` API.

**This is a breaking change and will require a major version release of the plugin. The minimum Gradle version required to use the plugin will be 4.3. Please check with your consuming projects.**

@JLLeitschuh Would you mind also having a look as you requested the feature?


cdancy:
@bmuschko ping when you've got things in working order and I'll review then. I take it from the `WIP` that you're still hacking away on it.

cdancy:
Interesting that your branch is not failing with the "onNext" issue that is popping for other PR's...

bmuschko:
@cdancy Yeah, it still requires some work. The changes are fast and have an impact on how you use the plugin to a certain extend. I am going to be done by the end of the week. Feel free to already get an overview on the impact on the end user. I want to make sure that your user base can easily migrate to a new plugin version. I might also add something to the user guide to explain the migration path.

bmuschko:
I believe the issue only comes in when you upgrade the version of Docker Java which I didn't do.

cdancy:
We use this plugin fairly extensively in our build pipeline and for local development purposes. I've got plugins myself built around this plugin. I can give things a go and see what breaks and provide feedback.

bmuschko:
@cdancy Sounds good. Let me do some more work in the next 2 days to get all tasks converted to the `Provider` API. Otherwise it might be somewhat inconsistent to use.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -17,33 +17,29 @@ package com.bmuschko.gradle.docker
 
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile
 import org.gradle.api.Action
+import org.gradle.api.Project
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Property
+import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Nested
 
 class DockerJavaApplication {
```
JLLeitschuh: Can you annotate this with `@CompileStatic` now?

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -17,33 +17,29 @@ package com.bmuschko.gradle.docker
...
-
-    @Deprecated
-    Integer port = 8080
-    Set<Integer> ports
-    String tag
-
-    Integer[] getPorts() {
-        return ports != null ? ports : [port]
+    final Property<String> baseImage
+    final Property<String> maintainer
```
JLLeitschuh: This says above that `maintainer` will be removed. Isn't this the perfect time to remove it?

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -59,42 +55,48 @@ class DockerJavaApplication {
...
+        void defaultCommand(Provider<List<String>> commandProvider) {
+            addInstruction(new Dockerfile.DefaultCommandInstruction(commandProvider))
         }
 
         void entryPoint(String... entryPoint) {
-            instructions << new Dockerfile.EntryPointInstruction(entryPoint)
+            addInstruction(new Dockerfile.EntryPointInstruction(entryPoint))
+        }
+
+        void entryPoint(Provider<List<String>> entryPointProvider) {
```
JLLeitschuh: Don't you want this to be a `ListProvider<String>` instead?

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -98,42 +117,59 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
             description = "Copies the distribution resources to a temporary directory for image creation."
             dependsOn installTask
             from { installTask.destinationDir.parentFile }
-            into { createDockerfileTask.destFile.parentFile }
+            into { createDockerfileTask.destFile.get().asFile.parentFile }
```
JLLeitschuh: Does `into` support taking a provide?
If so, you can change this to:
```groovy
into = createDockerfileTask.destFile.map { it.asFile.parentFile }
```

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -98,42 +117,59 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
...
             description = 'Builds the Docker image for the Java application.'
             dependsOn createDockerfileTask
-            conventionMapping.inputDir = { createDockerfileTask.destFile.parentFile }
-            conventionMapping.tag = { determineImageTag(project, dockerJavaApplication) }
+            inputDir = project.provider(new Callable<Directory>() {
+                @Override
+                Directory call() throws Exception {
+                    DirectoryProperty inputDirProperty = project.layout.directoryProperty()
+                    inputDirProperty.set(createDockerfileTask.destFile.get().asFile.parentFile)
+                    inputDirProperty.get()
```
JLLeitschuh: Uhhhhh.... What? Creating a provider just to return it's value? Are you sure this is what you want to be doing here??

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -98,42 +117,59 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
...
+            }
+        })
     }
 
     private void createPushImageTask(Project project, DockerBuildImage dockerBuildImageTask) {
         project.task(PUSH_IMAGE_TASK_NAME, type: DockerPushImage) {
             description = 'Pushes created Docker image to the repository.'
             dependsOn dockerBuildImageTask
-            conventionMapping.imageName = { dockerBuildImageTask.getTag() }
+            imageName = dockerBuildImageTask.getTag()
```
JLLeitschuh: 👍

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
```diff
@@ -1,17 +1,24 @@
...
+    final ListProperty<Integer> ports
+    final Property<String> tag
 
-    Integer[] getPorts() {
-        return ports != null ? ports as Integer[] : [8080] as Integer[]
+    DockerSpringBootApplication(Project project) {
+        baseImage = project.objects.property(String)
+        baseImage.set('openjdk:jre-alpine')
+        ports = project.objects.listProperty(Integer)
+        ports.add(8080)
```
JLLeitschuh: Do you want to make this a constant somewhere? You have it duplicated in at least two places.

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy
```diff
@@ -60,7 +64,7 @@ class DockerSpringBootApplicationPlugin implements Plugin<Project> {
                 sync.description = "Copies the distribution resources to a temporary directory for image creation."
                 sync.dependsOn project.tasks.getByName(BasePlugin.ASSEMBLE_TASK_NAME)
                 sync.from { archiveTask.archivePath }
-                sync.into { createDockerfileTask.destFile.parentFile }
+                sync.into { createDockerfileTask.destFile.get().asFile.parentFile }
```
JLLeitschuh: Same comment as above about possibly using `map`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -628,56 +936,53 @@ class Dockerfile extends DefaultTask {
     }
 
     static abstract class FileInstruction implements Instruction {
-        private final Object src
-        private final Object dest
-        private final Object flags
+        private final String src
+        private final String dest
+        private final String flags
+        private final Provider<File> provider
```
JLLeitschuh: Is this really a `File` or a `Directory`.
You probably want this to be a `RegularFileProperty` or a `DirectoryProperty`:

- https://docs.gradle.org/current/javadoc/org/gradle/api/file/RegularFileProperty.html
- https://docs.gradle.org/current/javadoc/org/gradle/api/file/DirectoryProperty.html

JLLeitschuh:
@bmuschko This is awesome.
You really want to use `DirectoryProperty` and `RegularFileProperty ` if possible because doing so automatically creates `dependsOn` couplings between the the task that creates these Properties and the task that consumes them. (There's much better support for this coming in Gradle 5.0).

On the whole this looks really good. I have to applaud you for the significant amount of work this must have taken to do! 👏 👏

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
cdancy: Agreed.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
bmuschko: I was planning to make everything in this code base statically compilable...at least as much as possible. However, I want to tackle this as part of a different PR. This PR is already enormous.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -16,411 +16,746 @@
...
+import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.*
 
+import javax.annotation.Nullable
+
 @CacheableTask
 class Dockerfile extends DefaultTask {
-    private final List<Instruction> instructions = new ArrayList<Instruction>()
+    private final ListProperty<Instruction> instructions
```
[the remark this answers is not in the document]
bmuschko: The input annotation is already on `getInstructions()`. We have to use `@Nested` here as it is a list of POJOs with input fields.

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
bmuschko: Good idea. Same is true for the base image. Let's also do that in a separate PR.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
bmuschko: If you look at the API of `ListProperty` you will see that the class extends from `Provider<List<T>>`. Passing in a `ListProperty` should work here. I don't necessarily want to expose a `Property` as the provided value should be immutable. Therefore, a `Provider` is likely the better choice.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
bmuschko: The `into` methods in `AbstractCopyTask` do not expose a `Provider` variant yet. Otherwise I'd agree with the change.

orzeh:
@bmuschko thanks for a really good job!
My only concern is that Provider API is an incubating Gradle feature so I believe you are pretty sure thaet the API won't change before it goes GA?

bmuschko:
@JLLeitschuh Thanks, it already took me multiple days to adopt the API and I am still not done. I am hoping that it will make it easier for consumers to use the plugin in practice.

cdancy:
> though I don't see any other solution to the problem on the horizon.
@bmuschko is there a problem per se or just updating the plugin to the latest and greatest features of gradle?

orzeh:
> @orzeh There's a certain risk involved in adopting this API as it might break our plugin consumers if something changes radically. I asked [when this API will go final](https://gradle-community.slack.com/archives/CA7UM03V3/p1537717616000100) yesterday but haven't gotten a response by the Gradle core team. I think it's the way forward as it allows consumers to provide lazily evaluated values from the build script. If we, as the maintainers of this plugin, think it's too risky to adopt the API then we can hold back though I don't see any other solution to the problem on the horizon.

Fair enough. IMO it's a step in right direction :+1:

bmuschko:
@cdancy The problem is the use of convention mapping to delay the evaluation of a property value. While it works, most users do not know the concept, it's an internal and completely undocumented API. As soon as you need to lazily evaluate something it becomes a massive hack (see the current implementation of the `DockerJavaApplicationPlugin` or the `DockerSpringBootApplicationPlugin`. Another example is described here: https://github.com/bmuschko/gradle-docker-plugin/issues/654#issue-357410970

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
JLLeitschuh: That makes sense to me.

JLLeitschuh:
I know that somewhere in your task hierarchy I think you were extending an internal gradle task in order to gain access to the `convention` mapping. Has this base class dependency been removed?

bmuschko:
> I know that somewhere in your task hierarchy I think you were extending an internal gradle task in order to gain access to the convention mapping. Has this base class dependency been removed?

@JLLeitschuh Thanks for the reminder. I just changed from `ConventionTask` to `DefaultTask`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileToContainer.groovy
```diff
@@ -41,7 +42,7 @@ class DockerCopyFileToContainer extends DockerExistingContainer {
      */
     @Input
     @Optional
-    Closure<File> tarFile
+    final RegularFileProperty tarFile = newInputFile()
```
JLLeitschuh: You should probably change the annotations to something like:
```
@InputFile
@PathSensitive(PathSensitivity.RELATIVE)
```

bmuschko: Yes, absolutely in the long run. At the moment the task is not marked with `@Cacheable` so the path sensitivity wouldn't make much of a difference. I did this for `Dockerfile` which already is cacheable.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy
```diff
@@ -16,411 +16,746 @@
...
         }
     }
 
     void instructionsFromTemplate(String templatePath) {
         instructionsFromTemplate(project.file(templatePath))
     }
 
-    void instructionsFromTemplate(Closure templatePath) {
-        instructionsFromTemplate(project.file(templatePath()))
-    }
```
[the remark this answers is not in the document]
bmuschko: Added a method that takes a `Provider<RegularFile>` though at the moment it is somewhat pointless as the contained value is evaluated right away. We can improve on that later.

bmuschko:
@cdancy @orzeh @JLLeitschuh Alright, after touching a majority of the files in this plugin I think I am done with the migration to the Provider API. Now is the time to give it another review and try it out in your own projects.

@JLLeitschuh The `implements PasswordCredentials` didn't make it as the return type has to be a provider. I could have introduced the following but then the API of this plugin wouldn't be uniform with the rest.

```
String getUsername()
Property<String> getUsernameProperty()
...
```

src/main/groovy/com/bmuschko/gradle/docker/DockerRegistryCredentials.groovy
```diff
@@ -16,38 +16,47 @@
...
 import groovy.transform.CompileStatic
-import org.gradle.api.credentials.PasswordCredentials
+import org.gradle.api.Project
+import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Optional
 
 @CompileStatic
-class DockerRegistryCredentials implements PasswordCredentials {
+class DockerRegistryCredentials {
```
JLLeitschuh: Do you want to do what I suggested here and offer a `asPasswordCredentials` method on this type?
https://github.com/bmuschko/gradle-docker-plugin/pull/655#issuecomment-424041169

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -98,42 +117,57 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
...
 
-    private String determineImageTag(Project project, DockerJavaApplication dockerJavaApplication) {
-        if (dockerJavaApplication.tag) {
-            return dockerJavaApplication.tag
-        }
+    private Provider<String> determineImageTag(Project project, DockerJavaApplication dockerJavaApplication) {
+        project.provider(new Callable<String>() {
+            @Override
+            String call() throws Exception {
+                if (dockerJavaApplication.tag.getOrNull()) {
```
thc202: Any specific reason why it's being used `getOrNull` over `isPresent`?

bmuschko: Should have been `isPresent()` mostly. I was planning to take another sweep through the change. Thanks for reminding me.

bmuschko:
@cdancy Any chance you can try this out with your projects soon? Any change on `master` causes a merge conflict in my branch so I'd like to integrate it.

bmuschko:
@cdancy Ping. I'd love to merge this soon. Could you please give it a shot?

cdancy:
@bmuschko forgot to get back on this but yes you have my blessing :) Things worked here and in another plugin I maintain that uses this in the backend.

JLLeitschuh:
🎆 Woot! 🎆

bmuschko:
@JLLeitschuh I still need to do some polishing but happy we have this change in `master` now. I am hoping to get this out by the end of next week if possible so that user can try it out.

cdancy:
I think a point release with this in place to let folks kick the tires would be a good idea.

bmuschko:
@cdancy This change is really a breaking change so I don't think a point release will do. For example feeding in values for some methods has changed. Most importantly, consumers will have to use Gradle > 4.3+. Anything older won't work anymore. I am advocating for releasing a major release version.

cdancy:
Hmmm ... But there is nothing new for the end user to take advantage of as this is more an internal thing with some hopefully positive side-effects for end users. Id be game for a new major release just feels excessive given that we are not, at least in my opinion, delivering something totally new folks can take advantage of.

bmuschko:
@cdancy I guess it depends on your use case. As a user you can now provide values in a lazy fashion by providing a `Provider`. There's no more hacks like `afterEvaluate`, convention mapping or lazy Groovy String evaluations. Good example is this one: https://github.com/bmuschko/gradle-docker-plugin/issues/654 which indeed is horrible code with potential side effects. Maybe those cases are not prominent in the projects you are dealing with.

The thing that's really breaking is the fact that we now require a much newer minimum Gradle version. I should also update that in the user guide. I believe we make a reference there.

bmuschko:
@cdancy Oh, and we actually made it possible for Kotlin DSL users to use the plugin (without hacks). This is to be 100% proven though. I am planning to add examples in the user guide and wanted to try out some thing before we release this version.

cdancy:
Fair enough. Major version it is then :)

PR #660, opened 2018-09-24
Bump docker-java to rc5

cdancy:


bmuschko:
@cdancy Did you have a chance to investigate the "hanging" functional test? At the moment any release activity hinges on getting this fixes. I'd like to release soon. We'll either have to get to the bottom of it or revert the change (and anything depending on it).

cdancy:
I haven't had a chance to sit with it due to deadlines here internally and not having my normal hours to work on other projects. I'll definitely have a look this weekend to see if we can make heads or tails of it.

bmuschko:
@cdancy I had a brief look at the issue and couldn't quite figure it out. I believe the `completed` CountDownLatch in `ResultCallbackTemplate` never gets the `countDown()` event. I tried catching the exception from the Closure and calling `onError(e)` would do it but no dice.

I suggest we revert this back if you don't have the time to investigate right now. It probably makes the most sense to look into migrating to the Spotify library.

cdancy:
@bmuschko this is a weird one. I had a look over the weekend but couldn't make head or tails of things. Re-ran with `forwardOutput` and indeed the exception is there but seemingly gets sent to the ether. Trying a few more things as it _feels_ there is something right in front of my eyes. IDK. I'll give this a few more goes but if nothing comes of it indeed I see we revert, fix what breaks, and kick another release and come back to this later.

bmuschko:
@cdancy Cool, the only thing I potentially wanted to do for this release is to move the release notes into AsciiDoc. It will make it easier to point users to current release and the changes via social media. I will put together a PR in the next hour.

bmuschko:
@cdancy Feel free to release anytime. I am good the pushing out what we have for 3.6.2.

bmuschko:
@cdancy Oh, boy. Seems like the test is still failing intermittently and broke the release. Could it be that the there's something holding a state?

cdancy:
@bmuschko yeah this gets more and more weird. I _don't think_ any state is being held but I'm not sure. Maybe we're not cleaning up enough on travis? IDK. I can't get it to pop here though. I'm not sure if we can archive those logs on travis or not to see what's going on.

bmuschko:
@cdancy I don't see this behavior on my local machine either. Could it be that the default Docker version on Travis CI doesn't properly auto-remove containers it yet? I might look into that later today.

bmuschko:
@cdancy I think I fixed it by upgrading to the latest Docker version on Travis.

cdancy:
We'll take it!

PR #662, opened 2018-10-02
Apply fix to all methods which use cglib enhancers.

cdancy:
Fix applied more generally. All tests pass locally. Let me know what you guys think.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Tue Oct 02 09:47:29 EDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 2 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Tue Oct 02 09:47:29 EDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
cdancy: Intellij snuck this in but I'm in favor of keeping it.

bmuschko: The only difference is that you can navigate to the Gradle sources if needed. Keep in mind though that the archive is much bigger than the `bin` ZIP. In turn builds will take longer. I usually go with the `bin` archive for performance reasons.

cdancy: Yeah it is a good size bigger now that I'm looking at it. Ok let me back this out.

PR #663, opened 2018-10-02
Move release notes to user guide

bmuschko:
Makes it easier to link to release notes for a specific release.


bmuschko:
A change log will never go away. It's extremely important. Now we can directly retrieve a link to any release. Makes it much easier to hand it out to others.

PR #678, opened 2018-10-12
Use appropriate types for Groovy/Kotlin DSL interoperability for reactive streams functionality

bmuschko:
Follow up to https://github.com/bmuschko/gradle-docker-plugin/pull/677

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -18,6 +18,7 @@ package com.bmuschko.gradle.docker.tasks
 import com.bmuschko.gradle.docker.utils.ThreadContextClassLoader
 import groovy.transform.CompileStatic
 import groovy.transform.TypeCheckingMode
+import org.gradle.api.Action
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.provider.Property
@@ -51,23 +52,9 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     @Internal
     ThreadContextClassLoader threadContextClassLoader
 
-    /**
-     * Closure to handle the possibly throw exception
-     */
-    @Internal
-    Closure onError
-
-    /**
-     * Closure to handle results
-     */
-    @Internal
-    Closure onNext
-
-    /**
-     * Closure to handle task completion
-     */
-    @Internal
-    Closure onComplete
+    private Action<? super Throwable> errorHandler
+    protected Action<? super Object> nextHandler
+    private Runnable completeHandler
 
     @TaskAction
     void start() {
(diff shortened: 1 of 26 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -51,23 +52,9 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
...
-    Closure onNext
-
-    /**
-     * Closure to handle task completion
-     */
-    @Internal
-    Closure onComplete
+    private Action<? super Throwable> errorHandler
+    protected Action<? super Object> nextHandler
+    private Runnable completeHandler
```
cdancy: Why change the names?

bmuschko: Because Groovy auto-generates getters/setter under the cover which we don't want. We want end users to only provide a handler through the exposed method and not some `setOnNext(...)`. There's also no point in having a getter for end users. Overall a cleaner API surface.

cdancy:
This wasn't bad at all. LGTM.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: Also ... why a Runnable over Action for the `onComplete`?

bmuschko: The `complete` callback doesn't receive a parameter. At least at the moment. The only sensible generic type for an `Action` would be a `Void`. I also checked with the Gradle Kotlin DSL team and that's what they recommended.

cdancy: Gotcha. Because if you were to do a `Action<? super Void>` you _may_ be required to actually pass in a type/null to make it not complain?

bmuschko: I don't think Groovy would complain really because of it being so dynamic. I guess if you would use a parameter the value would be `null`. Not sure how the Kotlin DSL would behave. Fact is...there's simply nothing we pass in so there's no reason to define a type of it. For more info see https://gradle-community.slack.com/archives/CAD95CR62/p1539355045000100.

cdancy: Also ... why the mix between private and protected here?

bmuschko: Because the field `nextHandler` is used in subclasses whereas the other two fields are not.

bmuschko: Oh, forgot to mention. Introducing these methods also doesn't require us to add `@Internal` anymore as there's no more getter method.

PR #680, opened 2018-10-12
Remove CodeNarc verification

bmuschko:
In its current configuration only one package is checked and it doesn't fail if there is an issue. The rest of the code base is not checked.

I see no benefit in having this code expect if we wanted to enable it for the whole code base and fix the issues.


cdancy:
Yeah yank it out.

Issue #666, opened 2018-10-04
It should be clearer that existing container tasks support container names and not just generated ids.


mwhipple:
### Expected Behavior

The existing container tasks should rely on a property with a name that conveys that either a (generated) container id **or** a specified container name (such as provided during container creation) can be used. Documentation should be updated appropriately.

### Current Behavior

The current name of and documentation for `targetContainerId` obscure the fact that a container name could also be used, and that support does not seem easily discoverable (I picked it out of the Docker API docs).

### Context

Using explicit container names can ease several aspects of managing containers during development.
See #637

cdancy:
Whatever we decide to use so long as things are backwards compatible, at least for 1 major release, I'm OK with entertaining ideas. We (here internally) as well as my own projects/plugins that depend on this plugin use this option extensively.

bmuschko:
@cdancy Yes, absolutely. There will have to be a deprecation cycle for sure.

mwhipple:
FWIW, I think the documentation updates could be sufficient (I probably should update the issue title to be more problem rather than solution oriented). The Docker API docs use `id` (potentially for the sake of REST convention) and then just have `ID or name of the container` in the parameter description.
> Naming things is hard

bmuschko:
@mwhipple Added some documentation here: https://github.com/bmuschko/gradle-docker-plugin/pull/698

PR #698, opened 2018-11-06
Clarify that "ID" means identifier or name

bmuschko:
For more information, see https://github.com/bmuschko/gradle-docker-plugin/issues/666

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -15,22 +15,15 @@
  */
 package com.bmuschko.gradle.docker.tasks.container
 
-import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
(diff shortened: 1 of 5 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -15,22 +15,15 @@
...
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 
-import java.util.concurrent.Callable
-
-class DockerCreateContainer extends AbstractDockerRemoteApiTask {
-    @Internal
-    final Property<String> imageId = project.objects.property(String)
-
+class DockerCreateContainer extends DockerExistingImage {
```
orzeh: Shouldn't it be:
```suggestion
class DockerCreateContainer extends DockerExistingContainer {
```

bmuschko: No, you are creating a new container so cannot depend on an existing container ID. Here you need an image ID.

PR #709, opened 2018-11-29
Update to Gradle 5.0

bmuschko:
[Property fields do not provide any more default values.](https://github.com/gradle/gradle/issues/7583) Therefore, appropriate default values need to be set.

Gradle 5 now adds a [`timeout` field by default for tasks](https://docs.gradle.org/5.0/userguide/more_about_tasks.html#sec:task_timeouts). Task types provided by this plugin needed to rename the existing field to avoid an overlap. @cdancy Potentially, we can now get rid of these fields and use the built-in `timeout` instead or even get rid of some of the tasks?


cdancy:
> Potentially, we can now get rid of these fields and use the built-in timeout instead or even get rid of some of the tasks? For now I had to rename the field.

That's fine but this is a breaking change and will definitely break things here if I attempt to upgrade and so I think at least a minor version upgrade is warranted here. Other than LGTM.

bmuschko:
@cdancy Agreed. That's why I assigned version 4.1.0 to it.

PR #718, opened 2018-12-13
Issue#716 Fix DockerCommitImage task and functional test

rroques:
Fixes Issue#716

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
@@ -2,6 +2,7 @@ package com.bmuschko.gradle.docker.tasks.image
 
 import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
 import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.UnexpectedBuildFailure
 
 class DockerCommitImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
@@ -11,6 +12,8 @@ class DockerCommitImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
             task commitImage(type: DockerCommitImage) {
                 dependsOn startContainer
                 targetContainerId createContainer.getContainerId()
+                author = "john doe"
+                message = "My image created"
                 tag = "myimage:latest"
             }
         """
@@ -23,6 +26,46 @@ class DockerCommitImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         then:
         result.output.contains("Commiting image for container")
     }
+
+    def "cannot commit image without container"() {
+        given:
+        String commitImage = """
+            task commitImage(type: DockerCommitImage) {
+                dependsOn startContainer
+                finalizedBy removeContainer
+            }
+        """
+
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
+
+        then:
+        def e = thrown(UnexpectedBuildFailure)
+        e.message.contains('No value has been specified for property \'containerId\'.')
+    }
+
+    def "cannot commit image with invalid container"() {
+        given:
+        String commitImage = """
+            task commitImage(type: DockerCommitImage) {
+                dependsOn startContainer
+                finalizedBy removeContainer
+                targetContainerId "idonotexist"
+            }
+        """
+
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
+
+        then:
+        def e = thrown(UnexpectedBuildFailure)
+        e.message.contains('{"message":"No such container: idonotexist"}')
+    }
+
 
     static String containerStart(containerCommitImageExecutionTask) {
         """
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
@@ -1,4 +1,54 @@
 package com.bmuschko.gradle.docker.tasks.image
 
-class DockerCommitImageFunctionalTest {
+import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
+import org.gradle.testkit.runner.BuildResult
+
+class DockerCommitImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
+
+    def "can commit image"() {
+        given:
+        String commitImage = """
+            task commitImage(type: DockerCommitImage) {
+                dependsOn startContainer
+                targetContainerId createContainer.getContainerId()
+                tag = "myimage:latest"
+            }
+        """
+
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
+
+        then:
+        result.output.contains("Commiting image for container")
+    }
+
+    static String containerStart(containerCommitImageExecutionTask) {
+        """
+            import com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer
+            import com.bmuschko.gradle.docker.tasks.container.DockerStartContainer
+            import com.bmuschko.gradle.docker.tasks.image.DockerCommitImage
+            import com.bmuschko.gradle.docker.tasks.image.DockerPullImage
+
+            task pullImage(type: DockerPullImage) {
+                repository = '$TEST_IMAGE'
+                tag = '$TEST_IMAGE_TAG'
+            }
+
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImageId()
+                autoRemove = true
+                entrypoint = ['tail', '-f', '/dev/null']
+            }
+
+            task startContainer(type: DockerStartContainer) {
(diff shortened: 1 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+                autoRemove = true
+                entrypoint = ['tail', '-f', '/dev/null']
+            }
+
+            task startContainer(type: DockerStartContainer) {
+                dependsOn createContainer
+                targetContainerId createContainer.getContainerId()
+            }
+            
+            ${containerCommitImageExecutionTask}
```
bmuschko: This test case is missing a task that stops the container. If we don't stop the container then we'll end up with stranded containers whenever the functional test is run. Could you please add it as a `finalizedBy` task? You will find some examples in other test classes.

cdancy:
Didn't even see there was a PR out for this and I put one together :) Lets go with this one as it provides a more complete solution.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
```diff
@@ -67,8 +67,8 @@ class DockerCommitImage extends DockerExistingContainer {
 
     @Override
     void runRemoteCommand(dockerClient) {
-        logger.quiet "Commiting image for container '${getContainerId()}'."
-        def commitCmd = dockerClient.commitCmd(getContainerId())
+        logger.quiet "Commiting image for container '${getContainerId().get()}'."
+        def commitCmd = dockerClient.commitCmd(getContainerId().get())
```
cdancy: Can we go with what I did for #720 here? I think it would be a good idea to start a new standard in that we need to check that the containerId was set and fail accordingly.

bmuschko: The `get()` method will actually throw an exception if no value was set.

cdancy: `getOrNull()`?

cdancy: I only say to use that as it _may_ not be exactly obvious when buried within the gradle stacktrace at least in this case we can control the message.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
rroques: Thanks for the feedback. I added image and container clean up in the test. Let me know how it looks.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
rroques: I updated the code with the implementation from #720

bmuschko: @cdancy The method `get()` will throw an exception if no value was set. The method `getOrNull()` will not throw an exception but will return null if no value is net. I'd try to create a test and see how the error message looks like. Otherwise, we are likely just duplicating code.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+            
+            task removeContainer(type: DockerRemoveContainer) {
+                removeVolumes = true
+                force = true
+                targetContainerId startContainer.getContainerId()
+            }
+
+            ${containerCommitImageExecutionTask}
+
+            task removeImage(type: DockerRemoveImage) {
```
bmuschko: You won't need to remove the image as it is being reused for other functional tests. We'll only want to download that once.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
+import org.gradle.testkit.runner.BuildResult
+
+class DockerCommitImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
+
+    def "can commit image"() {
+        given:
+        String commitImage = """
+            task commitImage(type: DockerCommitImage) {
+                dependsOn startContainer
```
bmuschko: This should say `finalizedBy removeContainer`.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
bmuschko: Oh, I see. Yeah, let's remove the committed image. That makes sense.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
rroques: I added a couple of functional tests for the cases where the container is missing, or is non existing. 

In the case of the missing container, the test does not show case the `getOrNull()` behaviour because the task validation kicks in first, and since `DockerExistingContainer#containerId` input is not optional, the validation fails the build with error `No value has been specified for property 'containerId'`.

In the case of the non existing container, then we get the error message from the docker daemon.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
+
+        then:
+        result.output.contains("Commiting image for container")
+    }
+
+    def "cannot commit image without container"() {
```
bmuschko: I'd say this test case is not necessary as we are basically testing that the Gradle `Provider` API works as expected. You can remove this test case.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+            task commitImage(type: DockerCommitImage) {
+                dependsOn startContainer
+                finalizedBy removeContainer
+            }
+        """
+
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
```
bmuschko: As a side note: You'd use `buildAndFail` in the case that you expect the build to run and fail. In those cases you won't need to catch an exception.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
```diff
@@ -0,0 +1,113 @@
...
+                dependsOn startContainer
+                finalizedBy removeContainer
+                targetContainerId "idonotexist"
+            }
+        """
+
+        buildFile << containerStart(commitImage)
+
+        when:
+        BuildResult result = build('commitImage')
```
bmuschko: Same here: This should be `buildAndFail`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
```diff
@@ -67,8 +67,10 @@ class DockerCommitImage extends DockerExistingContainer {
 
     @Override
     void runRemoteCommand(dockerClient) {
-        logger.quiet "Commiting image for container '${getContainerId()}'."
-        def commitCmd = dockerClient.commitCmd(getContainerId())
+        def localId = Objects.requireNonNull(getContainerId().getOrNull(), 'containerId must be set')
```
bmuschko: IMHO I don't think this handling is necessary. If we expect the container ID property to always container a value then we can just let Gradle fail the build with its own error message.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
rroques: The intention was to verify that the `DockerExistingContainer#containerId` is properly annotated, and that the Gradle input validation kicks in.

Happy to remove though.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImageFunctionalTest.groovy
rroques: Great! I did not know about that method.

bmuschko:
Thanks for the clarification. I am going to merge and clean it up in a couple of minutes so we can get you the fix.

bmuschko:
@rroques Thanks again for your work! We appreciated it.

Issue #674, opened 2018-10-08
Use task configuration avoidance API for task type of this plugin


JLLeitschuh:
This is different from using the `Provider` API for gradle, this is more using the lazy task configuration API like `register` instead of `create`.

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

For example, instead of `project.task(DOCKERFILE_TASK_NAME, type: Dockerfile) { ...` using
`project.tasks.register(DOCKERFILE_TASK_NAME, type: Dockerfile) {`

bmuschko:
Thanks for creating the issue. I wanted to do this as well sooner or later. We might have a wait a little though as this feature was only introduced in a recent version of Gradle. In turn all consumers will have to upgrade the Gradle version in their projects. Sometimes it's hard and long process to do this leaving behind users stuck on an older version of Gradle.

bmuschko:
@JLLeitschuh With the introduction of the `Provider` API, the minimum version of the plugin is already Gradle 4.8. I think I'd be OK with using the new API for task avoidance and raising the minimum version to 4.9. Would you be interested in providing a PR?

JLLeitschuh:
Something we may want to try because it seems to work in my plugin is something like this:
https://github.com/JLLeitschuh/ktlint-gradle/blob/64693e1b253eb10e7ad1bcbd3ddeb79127b3b03a/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintBasePlugin.kt#L18-L25
```kotlin
        val filterTargetApplier: FilterApplier = {
            if (GradleVersion.current() >= GradleVersion.version("4.9")) {
                // API only added in Gradle 4.9
                target.tasks.withType(KtlintCheckTask::class.java).configureEach(it)
            } else {
                target.tasks.withType(KtlintCheckTask::class.java, it)
            }
        }
```

bmuschko:
@JLLeitschuh Yes, that would be an option but it would make the plugin code much harder to maintain.  going forward. Apart from that `GradleVersion` is an internal class. I'd rather "force" users to upgrade to Gradle 4.9 if they want to use the latest features of the plugin. I'd don't expect a lot of breaking changes between 4.8 and 4.9.

bmuschko:
@JLLeitschuh Still interested in contributing this change? I'd be OK with raising the minimum Gradle version to 4.9 now that 5.0 is already out.

JLLeitschuh:
[...]

PR #724, opened 2018-12-17
#674 Configuration avoidance api migration for DockerRemoteApiPlugin

tylernhoward:
Relates to issue #674.

Really simple two-line PR  that starts the migration to the configuration avoidance api, following the [migration guide for configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview).

screen shot 2018-12-17 at 5 11 43 pm

Motivation was due to any tasks of type `AbstractDockerRemoteApiTask` being secretly "realized" early due to explicit task configuration occurring through the old method, which was difficult to track down when converting our tasks to use the newest gradle hotness. This will allow plugins that wrap the `DockerRemoteApiPlugin` to make use of the configuration avoidance API for tasks brought in by this plugin.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.groovy
@@ -67,7 +67,7 @@ class DockerRemoteApiPlugin implements Plugin<Project> {
 
     private void configureAbstractDockerTask(Project project, DockerExtension extension) {
         ThreadContextClassLoader dockerClassLoader = new DockerThreadContextClassLoader(extension)
-        project.tasks.withType(AbstractDockerRemoteApiTask, new Action<AbstractDockerRemoteApiTask>() {
+        project.tasks.withType(AbstractDockerRemoteApiTask).configureEach(new Action<AbstractDockerRemoteApiTask>() {
             @Override
             void execute(AbstractDockerRemoteApiTask apiTask) {
                 apiTask.with {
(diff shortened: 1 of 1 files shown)


bmuschko:
@tylernhoward Thanks for getting started on this effort. Let's pull through by using all task avoidance APIs across all code of this plugin to keep behavior consistent.

@cdancy Be aware that the minimum, runtime version of this plugin will be Gradle 5.0 after switching to the new API. Would your project be ready to use 5.0?

cdancy:
@bmuschko we don't have any current plans to bump to gradle 5.0 anytime soon unfortunately as it will be a big undertaking. Is there anything else we can really target here besides what has already been done?

bmuschko:
@cdancy If you effectively want to use the new API then you'll have to go for Gradle 5.0. I'd say let's hold off on this for a while until more users caught up to newer versions of Gradle.

bmuschko:
I am closing this PR for now. Feel free to reopen in say 6 months when the adoption rate of Gradle 5.0 will be higher.

tylernhoward:
Sounds good!

bmuschko:
@tylernhoward Thanks again for getting the ball rolling on this.

ryandens:
Hi @bmuschko! Have you reconsidered your thoughts on the Gradle 5.0 adoption rate? We're having a hard time extending this plugin to fit our use case as a result of a task of type `DockerBuildImage` getting realized too early. We're working on a workaround, but it would be simpler to make the change here if you're comfortable with moving forward on this

PR #729, opened 2019-01-10
Bump docker-java-shaded to latest and greatest and decrease wait time for DockerLiveness test.

cdancy:


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLivenessFunctionalTest.groovy
```diff
@@ -11,7 +11,7 @@ class DockerLivenessFunctionalTest extends AbstractGroovyDslFunctionalTest {
             task livenessProbe(type: DockerLivenessContainer) {
                 dependsOn startContainer
                 targetContainerId startContainer.getContainerId()
-                livenessProbe(300000, 30000, 'database system is ready to accept connections')
+                livenessProbe(300000, 1000, 'database system is ready to accept connections')
```
bmuschko: There are multiple timeouts in this test class. Let's reduce the number for all of them.

bmuschko:
Looks like something changed in the API between RC-5 and RC-7: https://scans.gradle.com/s/3kmgntfaoju44/tests/dqwcqqtprpdqc-3wa7wcubb5f4o

cdancy:
Yeah was looking at that. Fun times with those guys...

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLivenessFunctionalTest.groovy
cdancy: On it.

cdancy:
yeah looks like they moved all of that stuff into `HostConfig`

https://github.com/docker-java/docker-java/commit/580ea407cc362276e4f5a47b24593b12a0fae26e#diff-43a8d1cb8dfb2c38ae576e0528cc9793

bmuschko:
@cdancy Ballsy change in between RCs. BTW: Will they ever release a final version? ;-)

cdancy:
@bmuschko yeah who knows. It's a one man ship over there and that's a good size project for just one guy to handle on his own. Have no idea what his personal life is like but if it's anything like mine the past year (new baby, new position in company, etc) then he may be stretched thin which I can appreciate. Still ... we need to move to using the spotify client at some point. Looking on their changelog they even have support for windows named pipes coming with the next release which is something we sorely need here.

cdancy:
Weird ... this latest failure does not pop locally.

bmuschko:
@cdancy What I meant was not the amount of changes that go into a release or how long it takes to release a version. It was more about the fact that 3.1.0 seems to continuously change for months without ever getting releases as a final version. From my point of view they could just make a single change and then call it final. Using RCs for year simply send the wrong signals to users of the library.

buildSrc/src/main/kotlin/com/bmuschko/gradle/docker/FunctionalTestPlugin.kt
```diff
@@ -36,7 +36,8 @@ class FunctionalTestPlugin : Plugin<Project> {
                 events("started", "passed", "failed")
             }
         }
+        functionalTest.include("**/DockerCreateContainer*")
```
bmuschko: Why exactly is this needed?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLivenessFunctionalTest.groovy
```diff
@@ -66,7 +66,7 @@ class DockerLivenessFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 withCommand(['su', 'postgres', "-c", "/usr/local/bin/pg_ctl stop -m fast"])
                 successOnExitCodes = [0, 1, 137]
                 awaitStatusTimeout = 60000
```
bmuschko: Should this use a lower number as well?

cdancy: No. It's the probing we care about that was set to run every 10/30 seconds and unnecessarily dragging the test out.

buildSrc/src/main/kotlin/com/bmuschko/gradle/docker/FunctionalTestPlugin.kt
cdancy: It's not. I was playing around with setting `include = '...'` but it seems kotlin doesn't like that so I tried on the reference itself which seems fine. Was for dev/testing purposes. Taking out now.

cdancy:
Once things are in a working state I'll squash the commits so that we don't have these silly sentences I've been sending in that don't provide any value.

cdancy:
Can't reproduce the failing test here and I've tried on multiple machines now. @bmuschko @orzeh any thoughts? It _smells_ like an environmental issue

orzeh:
I can't reproduce it locally :confused:

bmuschko:
@cdancy @orzeh Does the test actually _run_ for you? It [requires a private registry to be set up](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy#L119). The test is skipped If the registry is not available.

orzeh:
@bmuschko spot on! With registry up and running I was able to reproduce and fix the bug. We had a workaround for docker-java bug which [was fixed](https://github.com/docker-java/docker-java/pull/1086) RC7. I removed the code and pushed changes here.

cdancy:
@orzeh good catch! Think we are good to go on this PR unless you guys have any objections?

buildSrc/src/main/kotlin/com/bmuschko/gradle/docker/FunctionalTestPlugin.kt
```diff
@@ -39,4 +39,4 @@ class FunctionalTestPlugin : Plugin<Project> {
 
         tasks["check"].dependsOn(functionalTest)
     }
-}
```
bmuschko: @cdancy Can you remove this extra new line?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLivenessFunctionalTest.groovy
bmuschko: OK

cdancy:
@bmuschko will def fix before merging but weird that this did not show on my local testing (and this one did actually run). Poking around now...

bmuschko:
@cdancy My guess is that the Travis machines are a lot slower than your machine. You may need to increase the timeout a little bit.

PR #731, opened 2019-01-18
Allow to specify additional groups in DockerCreateContainer

### LINKED ISSUE #551: How to pass additional groups to docker containers
You can use `--group-add` to pass additional groups while creating docker container. How can I achieve this functionality with the plugin?

https://docs.docker.com/engine/reference/commandline/create/#options

orzeh:
Adds `groups` property to `DockerCreateContainer`.

Closes #551

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -48,11 +48,8 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
             }
         """
 
-        when:
+        expect:
         build('inspectContainer')
-
-        then:
-        noExceptionThrown()
     }
 
     def "can override default MAC address"() {
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -21,6 +21,27 @@ import org.gradle.testkit.runner.TaskOutcome
 
 class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
+    def "can setup additional user groups"() {
+        given:
+        String containerCreationTask = """
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImageId()
+                cmd = ['groups']
+                groups = ['postgres']
+            }
+        """
+        buildFile <<
+            containerStart(containerCreationTask) <<
+            containerLogAndRemove()
+
+        when:
+        BuildResult result = build('logContainer')
+
+        then:
+        result.output.contains("postgres")
+    }
+
     def "can override default MAC address"() {
         given:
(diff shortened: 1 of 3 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -21,6 +21,27 @@ import org.gradle.testkit.runner.TaskOutcome
...
+        """
+        buildFile <<
+            containerStart(containerCreationTask) <<
+            containerLogAndRemove()
+
+        when:
+        BuildResult result = build('logContainer')
+
+        then:
+        result.output.contains("postgres")
```
bmuschko: Is there a little bit more output we could verify?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -48,6 +48,13 @@ class DockerCreateContainer extends DockerExistingImage {
     @Optional
     final Property<String> user = project.objects.property(String)
 
+    /**
+     * A list of additional groups that the container process will run as.
+     */
```
bmuschko: Can you add the `@since` annotation with the next version so users know when this field was introduced?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
orzeh: > Can we do something more specific like print out the groups of the given user and ensure a group which the user should not have is added and present?

This is what this test do right now. I just wonder if we actually should test Docker here (whether it add groups or not) or maybe we should just inspect container and see if additional groups are returned - WDYT?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
orzeh: I thought about that, but what would be next version number?

bmuschko: Correct, I already assigned the milestone to the issue.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
bmuschko: Inspecting the container is the way to go IMHO.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -21,6 +21,40 @@ import org.gradle.testkit.runner.TaskOutcome
...
+                targetContainerId startContainer.getContainerId()
+                onNext { info ->
+                    if (info.hostConfig.groupAdd != ['postgres']) {
+                        throw new GradleException("Additional group not found!")
+                    } 
+                }
+            }
+        """
+
+        when:
```
bmuschko: You can just use `expect` here and remove `noExceptionThrown`. The test case would fail anyway if an exception is thrown.

Issue #701, opened 2018-11-08
Convention plugins should be able to cache more fine-grained image layers


bmuschko:
### Expected Behavior

When creating the Dockerfile each of the following resources should be defined as individual layer.

1. External libs
2. Resource files
3. Class files

In turn we wouldn't have to build a JAR anymore - one less task to run. Layers 1-2 will likely not change very often. Therefore, Docker will be able to cache the image layers more efficiently. The java command would add all resources file to the classpath with the `*` notation.

This behavior should apply to the Docker Java Application plugin and the Docker Spring Boot application plugin.

### Current Behavior

The Docker Java Application plugin already creates a layer for the external dependencies and one for the application files. We could optimize even more to improve caching.

The Docker Spring Boot application plugin always builds the full JAR or WAR file. Any change will result in a rebuilt of the layer.

### Context

Make the produced Docker image more efficient.

PR #736, opened 2019-01-28
Define image with more fine-grained image layers

bmuschko:
Layers can be cached much better than with a single layer for an archive which changes every time.

* Plugins do not have to build an archive anymore which should speed up the build time
* The image defines separate layers for libraries, resource files and class files in the order from less likely changes to the most frequent changes
* The main class name for Spring Boot application is determined automatically (which required a dependency on a Spring Boot lib)
* We could probably even get rid of dependency on the Gradle application plugin by searching for the main class name on the classpath as well
* Less clutter in the Docker registry as layers become more cacheable
* This is a breaking change and will be marked as such in the release notes

For more information see https://github.com/bmuschko/gradle-docker-plugin/issues/701

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,6 +23,7 @@ repositories {
 }
 
 dependencies {
+    implementation("org.springframework.boot:spring-boot-loader-tools:2.1.2.RELEASE")
     testImplementation("org.spockframework:spock-core:1.2-groovy-2.5") {
         exclude(group = "org.codehaus.groovy")
     }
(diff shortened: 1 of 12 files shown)


bmuschko:
@cdancy You may want to give this out the change in one of your projects before we release it.

cdancy:
> You may want to give this out the change in one of your projects before we release it.

@bmuschko now I'm curious as to why you ask...

bmuschko:
@cdancy Because the generated `Dockerfile` changed very drastically under the hood (at least if you are using the convention plugins). Also if you later modify the generated Dockerfile then the new structure could have side-effects.

bmuschko:
@cdancy I am going to release this if you don't mind.

cdancy:
@bmuschko yeah go for it.

cdancy:
Our "main" project is still on 4.10.2. We have many internal smaller projects which are free to do as they see fit. Everything from 3.x up to the latest :)

PR #742, opened 2019-02-02
Set publishAll through hostConfig

### LINKED ISSUE #741: Cannot set publishAll property without error
Hi, thanks for making/maintaining this plugin! 
I'm running into an issue where I want to use publishAll to avoid static host ports, but when I configure my task like so
```kotlin
val createContainer by registering(DockerCreateContainer::class) {
...
    publishAll.set(true)
...
}
```
I receive this error when running the task:
> No signature of method: com.github.dockerjava.core.command.CreateContainerCmdImpl.withPublishAllPorts() is applicable for argument types: (Boolean) values: [true]

I've tried debugging this but the thread context classloader situation makes things a bit opaque.

thyming:
Closes #741 
This sets publishAll via hostConfig instead of directly on the create container command, as that option was removed in docker-java recently.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -167,6 +167,48 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
         build('inspectStoppedContainer')
     }
 
+    def "with publishAll set, all ports get a host binding"() {
+        given:
+        def exposedPorts = [1234, 2345]
+        String containerCreationTask = """
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImageId()
+                autoRemove = true
+                
+                cmd = ['sleep', '10']
+                exposePorts 'tcp', $exposedPorts
+                publishAll = true
+            }
+        """
+
+        String containerInspect = '''
+            task inspectContainer(type: DockerInspectContainer) {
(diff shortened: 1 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -167,6 +167,48 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
...
+        buildFile << containerStart(containerCreationTask)
+        buildFile << containerStop()
+        buildFile << containerInspect
+
+        when:
+        BuildResult buildResult = build('inspectContainer')
+
+        then:
+        exposedPorts.every {
+            buildResult.output.find(/$it -> \d+/) != null
```
thyming: open to a cleaner solution for capturing the output of docker inspect; not familiar with gradle's testing APIs and saw this kind of thing done elsewhere.

bmuschko: Could we verify the actual ports that we'd expect?

thyming: how do you suggest I do that?

bmuschko: Oh, I see. Then it's probably not possible. I think we should be OK with merging what we have right now.

@cdancy Do you have any other ideas here?

bmuschko:
Thanks for providing the pull request. I am going to release a new version soon.

Issue #728, opened 2019-01-06
Plugin should declare and use Docker Java on compilation classpath


bmuschko:
### Expected Behavior

No need for a thread context classloader. Docker Java classes can be imported into plugin classes. IDE support through auto-completion. No more reflection hacks.

### Current Behavior

1. Initially, the Docker Java library used by the plugin was not shaded. To mitigate the situation, the plugin used its own thread context classloader and loads Docker Java classes at runtime.
2. Later, dependencies used by Docker Java were packaged in the same JAR and move to a different package.

Given that 2. has been done, 1. is not needed anymore. @cdancy Is this a correct assessment?

### Context

Gradle does not isolate the classpath across all applied plugins in a build. As a result, dependencies required by applied plugins can conflict and result in unexpected behavior e.g. two plugins require Guava but they depend on different versions.

The shaded Docker Java library moves all external library imports into a different package to avoid conflicts with dependencies in build scripts on the build script classpath.

cdancy:
Yes this is all true. And if and when we move to using the spotify client it also has a shaded version which should shield us from any type surrounding this area.

cdancy:
So the idea here would be to remove the runtime resolution of the library like we are doing now and instead do things the more "normal" way as a proper dependency? It would make debugging things easier. The only potential gotcha I see is that in massive projects like ours (almost 200 subprojects at this point) we have points in our build pipeline that invoke the greater build but don't require docker. Furthermore those points are executed on separate nodes. If we put this into place then in theory every time our project is run, no  matter how big or small the context, the docker-java-shaded jar will be resolved which I believe is a good 15MB or so.

bmuschko:
That's correct. I'd like to use actual Docker Java classes than reflection. It also makes it easier for users to deal with callback for example.

> If we put this into place then in theory every time our project is run, no matter how big or small the context, the docker-java-shaded jar will be resolved which I believe is a good 15MB or so.

That depends on your setup. Do you use ephemeral nodes? Usually, you'd be able to just download it once and reuse it from Gradle's dependency cache. Either way, don't you have the same problem with other dependencies in your project?

cdancy:
Yeah ephemeral nodes. Same problem elsewhere but other deps tend to be smaller. This will most likely not be such a big deal for us as we have an Artifactory instance sitting side-by-side with our CI infrastrcuture so all downloads are more/less local.

cdancy:
I would say we can move forward just wanted to note it.

[...]

PR #751, opened 2019-02-16
 Use Docker Java as runtime library

bmuschko:
For more information see https://github.com/bmuschko/gradle-docker-plugin/issues/728

As a next step I would also propose that the `nextHandler` defines the type as well. I guess it would make it easier for plugin users to know what type they are actually dealing with. The only drawback is that it clearly exposes Docker Java types to the API of this plugin but at the moment we are bound to it and that's how it is. It would be a major breaking change anyway to switch to the Spotify library anyway.

This change also required renaming the custom task `DockerClient` to `DockerOperation` to avoid confusion with the Docker Java type.


bmuschko:
> Code wise things look fine (aside from CI failures). 

I'll have to investigate the build issue. I am not quite sure why that's happening. I am suspecting some sort of conflicting dependency conflict.

> It stinks in a way as this will be resolved at build time now instead of as needed if I'm not mistaken which is a good win for desktop users that don't necessarily need to run any sort of docker anything as part of building a given project locally.

I'd probably apply the plugin conditionally. From what I understand from your comment, this is only required in your CI builds?

```
if (System.env('CI')) {
    apply plugin: '...'
}
```

> Still .... it causes more confusion though hiding things in the back. I'm indifferent here mostly but leaning slightly in this direction.

It's horrible to deal with this as a plugin developer as you constantly have to deal with reflection. Seems backward. It's really something Gradle should address internally by providing a classloader per plugin. For users it's also impractical as have to know which type you are dealing with if you wanted to handle a `onNext` callback. There's no way to know except for reading the Docker Java source code.

build.gradle.kts
```diff
@@ -23,6 +23,11 @@ repositories {
 }
 
 dependencies {
+    implementation("com.aries:docker-java-shaded:3.1.0-rc-7:cglib@jar") {
+        setTransitive(false)
+    }
+    implementation("org.slf4j:slf4j-simple:1.7.5")
```
bmuschko: @cdancy Can you explain why we need the SLF4J and Activation libraries? Shouldn't those be included in the Docker Java library as shaded packages?

Issue #744, opened 2019-02-04
Refactor spring boot dependency into a different artifact


JamesXNelson:
The introduction of `implementation` dependency on spring boot caused our company's internal license checks to fail because it was not explicitly allowed.

We've simply excluded the dependency manually, but it might be nice if it was deployed as a different artifact that spring boot users would need to choose to get the plugin / dependencies they need.

bmuschko:
The dependency `org.springframework.boot:spring-boot-loader-tools` is already a transitive dependency of the Spring Boot plugin so you'd pull that in anyway (expect if you are not building a Spring Boot application - is that the case?):

```
./gradlew buildEnvironment

> Task :buildEnvironment

------------------------------------------------------------
Root project
------------------------------------------------------------

classpath
+--- org.springframework.boot:org.springframework.boot.gradle.plugin:2.1.2.RELEASE
|    \--- org.springframework.boot:spring-boot-gradle-plugin:2.1.2.RELEASE
|         +--- org.springframework.boot:spring-boot-loader-tools:2.1.2.RELEASE
|         |    +--- org.springframework:spring-core:5.1.4.RELEASE
|         |    |    \--- org.springframework:spring-jcl:5.1.4.RELEASE
|         |    \--- org.apache.commons:commons-compress:1.18
|         +--- io.spring.gradle:dependency-management-plugin:1.0.6.RELEASE
|         \--- org.apache.commons:commons-compress:1.18
\--- io.spring.dependency-management:io.spring.dependency-management.gradle.plugin:1.0.6.RELEASE
     \--- io.spring.gradle:dependency-management-plugin:1.0.6.RELEASE
```

The only option I see is to pull the code into the plugin that finds the main class in a classpath. Happy to accept pull requests. Would you be interested?

bmuschko:
Do you have any thoughts on my response?

JamesXNelson:
Ah, sorry.  We are building an application that doesn't use spring at all.  When we upgraded to the latest version of gradle-docker-plugin, we inherited the spring boot dependency that we did not need or want.  

We have license checks that fail if anything new is added without an entry confirming the license is usable (OSS or purchased); obviously spring boot wasn't a problem, but we chose to exclude it any as we don't need it.

[...]

PR #752, opened 2019-02-17
Inline main class finder

bmuschko:
Gets rid of more dependencies that most users do not need (especially if you are not using the Spring Boot Docker plugin).

For more information see https://github.com/bmuschko/gradle-docker-plugin/issues/744


bmuschko:
I am going to merge this. Let me know if you have any other comments.

PR #753, opened 2019-02-18
 Dockerfile.FileInstruction does not use flags if Dockerfile.File is passed in using a Provider (Issue #747)

### LINKED ISSUE #747: Dockerfile.FileInstruction does not use flags if Dockerfile.File is passed in using a Provider
If a `FileInstruction` is created using the constructor using a `Provider` : 

https://github.com/bmuschko/gradle-docker-plugin/blob/703ce60f498314092f2964ca6fd6fe19630bd428/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy#L1020-L1034

Then the `FileInstruction`s `flags` field is never set. However, the logic building the line for the docker file looks at `FileInstruction::flags` only, ignoring `File::flags`. 

https://github.com/bmuschko/gradle-docker-plugin/blob/703ce60f498314092f2964ca6fd6fe19630bd428/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/Dockerfile.groovy#L1048-L1050

Since `file.flags` is non-null but `this.flags` is `null` in this case, this leads to a line like
```
ADD null src dst
```
being added to the Dockerfile, which crashes the image build.

This seems to be a simple case of a forgotten check. If so, I'm happy to provide a PR to fix it. Am I missing something that would make that more complicated than it seems at first glance?

schoeneu:
closes #747

Fixes a bug where `flags` are unused in `FileInstruction`s if passed using `Provider`.
Also adds a functional test verifying this.

I did *not* add a unit test in `DockerfileTest` as this test does not have access to the gradle `Provider` factories, and directly instantiating one would be use of an internal API. However, I'm open to doing so if you think it's the right way to go.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
@@ -144,6 +144,7 @@ LABEL maintainer=benjamin.muschko@gmail.com
                 environmentVariable ENV_VAR_A: 'val_a'
                 environmentVariable ENV_VAR_B: 'val_b', ENV_VAR_C: 'val_c'
                 addFile 'http://mirrors.jenkins-ci.org/war/1.563/jenkins.war', '/opt/jenkins.war'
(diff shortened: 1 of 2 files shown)


bmuschko:
Thanks for the pull request. We should probably have a test for all instruction methods that take a `Provider`. I will take it from here. We should have had those in the first place.

PR #755, opened 2019-02-25
Shade all dependencies expect Docker Java

bmuschko:
Avoids runtime classpath conflict with other plugins.

With this change the project https://github.com/project-aries/docker-java-shaded is not needed anymore saving yet another roundtrip.

@cdancy @orzeh Please publish locally and test with your own projects.

PR #756, opened 2019-02-28
Bumped slf4j-simple

mleegwt:
Catching up on about six years of updates of SLF4J. This also resolves warnings resulting from the versions gradle  plugin about available updates in projects that depend on this plugin.


bmuschko:
Thanks for your pull request. I don't think this change will be necessary anymore after we merge https://github.com/bmuschko/gradle-docker-plugin/pull/755. IMHO the dependency isn't needed at all.

mleegwt:
#755 looks good to me. Closing this pull request.

PR #757, opened 2019-03-04
Bumped up the version of the docker-java and verified that the parsin…

### LINKED ISSUE #745: DockerWaitHealthyContainer - Unable to bind StartPeriod and Interval from HEALTHCHECK - out of range of int
### Expected Behavior
When specifying task `DockerWaitHealthyContainer`, then there is an Json mapping exception for resolving Interval and StartPeriod properties.

It should be working for bigger values specified in HEALTCHECK like 
`HEALTHCHECK --interval=1m --start-period=5m`

### Current Behavior

For (from docker inspect)

`
            "Healthcheck": {
                "Test": [
                    "CMD-SHELL",
                    "\"$ORACLE_BASE/$CHECK_DB_FILE\" >/dev/null || exit 1"
                ],
                "Interval": 60000000000,
                "StartPeriod": 300000000000
            },
`

it returns

`
11:48:14.592 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] Execution failed for task ':btn-web:waitForOracleStart'.
11:48:14.592 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter] > com.github.dockerjava.shaded.com.fasterxml.jackson.databind.JsonMappingException: Numeric value (300000000000) out of range of int
11:48:14.592 [ERROR] [org.gradle.internal.buildevents.BuildExceptionReporter]    at [Source: (com.github.dockerjava.shaded.org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 5169] (through reference chain: com.github.dockerjava.api.command.InspectContainerResponse["Config"]->com.github.dockerjava.api.model.ContainerConfig["Healthcheck"]->com.github.dockerjava.api.model.HealthCheck["StartPeriod"])
`

### Context

gradle-docker-plugin uses [docker-java](https://github.com/docker-java/docker-java) library throu [shaded version](https://github.com/project-aries/docker-java-shaded). There seems to be [fixed](https://github.com/docker-java/docker-java/blob/4eef5ed7171ef6657d93bf72d6560028bb2c3c93/src/main/java/com/github/dockerjava/api/model/HealthCheck.java#L59) in latest version. 

So, maybe it would be sufficient to bump up the version of docker-java in shaded version. But it's only my guess ...

Define any task `task waitForContainerStart(type: DockerWaitHealthyContainer) {` for container based on image which has HEALTCHECK interval or startperiod set by minutes (5m or so).

### Your Environment

MacOS, Docker 18.09.1, gradle 5.2, gradle-docker-plugin 4.4.0

mikealdo:
…g of the value for the --start-period of the HEALTHCHECK dockerfile command is passing

closes #745


bmuschko:
Thanks for providing the pull request. Can you please fix the [test failure](https://scans.gradle.com/s/zf7xthgb6hpgq/tests/dqwcqqtprpdqc-thh6g7u63jx3c)? My guess is that Docker Java made a breaking change.

mikealdo:
Yes, I was looking into it yesterday for some time but without any success. It's reproducible locally so I'll take a look on it today.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -73,7 +73,7 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
             task buildImage(type: DockerBuildImage) {
                 dependsOn dockerfile
-                inputDir = file("build/docker")
+                inputDir = file("${BUILD_DIR}")
                 tags.add("${IMAGE_ID}")
```
mikealdo: When tag does not contain repository (e.g. alpine:foo) then the string is treated as repository and the tag is always "latest".

bmuschko: I don't fully understand why this had to change just because we bumped up the Docker Java version. Can you explain in more detail?

bmuschko:
The build is still failing. I'd recommend running the full test suite on your local machine first. BTW: The build is configured to fail on the first issue. Subsequent issues won't pop up until you fix the first encountered issue.

mikealdo:
Thanks for advice. I have configured it even before and it was failing even locally but in different test method than on CI. I was trying to run it inside Intellij Idea but the resulting errors was pretty weird:

> Failed to deserialize script metadata extracted for build file '/private/var/folders/5j/4drtjh2932bdn1rn0c6jshxh0000gn/T/junit1408203699599863172/build.gradle'

> Caused by: java.io.FileNotFoundException: /private/var/folders/5j/4drtjh2932bdn1rn0c6jshxh0000gn/T/.gradle-test-kit-mdavidek/caches/5.2/scripts-remapped/build_572lw9douwij3q55bbmrawb8n/568cw6r6f2tjie5z61ukaykwe/cp_projeadbf4c451bc4876d277fc099d472843/metadata/metadata.bin (No such file or directory)
	at org.gradle.groovy.scripts.internal.DefaultScriptCompilationHandler.loadFromDir(DefaultScriptCompilationHandler.java:205)

So, after some research, I tried to completely remove the 

> /private/var/folders/5j/4drtjh2932bdn1rn0c6jshxh0000gn/T/.gradle-test-kit-mdavidek

 folder and the result was - the test was passing in intellij idea and the complete build by gradle was successful as well.

So, maybe there is some issue with the gradle test framework and his cache which isn't somehow properly cleaned before the build is executed on the Travis. Can you, @bmuschko, please check it? (As I don't see anything else which I can fix ...)

mikealdo:
I'm just looking into the [job config](https://travis-ci.org/bmuschko/gradle-docker-plugin/builds/504408350/config), so maybe adding well aimed `rm -rf` to the 'before_cache' section would resolve this issue ...

mikealdo:
Hm, and after the fixing merge conflict, the build is ok ... I don't know if you experienced similarly unexpectable problems in the past, but I'm not sure that the change won't unstabilize the plugin ...

bmuschko:
> Hm, and after the fixing merge conflict, the build is ok ... I don't know if you experienced similarly unexpectable problems in the past, but I'm not sure that the change won't unstabilize the plugin ...

Weird. Sounds like a flaky test. We will have to investigate if it happens again.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
bmuschko: Can you please explain this change?

bmuschko: @mikealdo Can you explain why this needed to change. I am not going to merge the pull request before we had a chat about it.

cdancy:
FYI: docker-java has been bumped to `3.1.1` on master on top of the `DockerSaveImageFunctionalTest` having some minor changes which will need to be fixed here.

bmuschko:
@mikealdo Are you still planning to update the PR?

bmuschko:
I am closing this issue as we haven't heard back. Either way the new release will contain the latest Docker Java version.

PR #763, opened 2019-03-08
Add capability to create tmpfs mount

### LINKED ISSUE #762: Missing support for tmpfs/mounts
Hi, I don't believe there's a way exposed through the gradle-docker-plugin to create tmpfs mounts. I'd like to leverage this feature in test database containers.

thyming:
Closes #762.
This exposes a new list property `tmpfs` that allows you to specify one or more tmpfs mounts, analogous to the `--tmpfs` flag in docker cli.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -105,6 +105,40 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
         result.output.contains("HWaddr 02:03:04:05:06:07")
     }
 
+    def "can create tmpfs mounts"() {
+        given:
+        String tmpFsMount = "/test-tmp-fs"
+        String containerCreationTask = """
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImageId()
+                cmd = ['grep', 'tmpfs $tmpFsMount', '/proc/mounts']
+                tmpfs = ['$tmpFsMount']
+                cpuset = '1'
+                labels = ["project.name": "\$project.name"]
+            }
+        """
+
+        buildFile <<
+            containerStart(containerCreationTask) <<
+            containerRemove() <<
+            """
+            task inspectContainer(type: DockerInspectContainer) {
+                dependsOn startContainer
+                finalizedBy removeContainer
+                targetContainerId startContainer.getContainerId()
+                onNext { info ->
+                    if (!info.hostConfig.tmpFs.containsKey("$tmpFsMount")) {
+                        throw new GradleException("TmpFs not found")
+                    } 
+                }
+            }
+        """
+
+        expect:
+        build('inspectContainer')
+    }
+
     def "can set multiple environment variables"() {
         given:
         String containerCreationTask = """
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -192,7 +192,7 @@ class DockerCreateContainer extends DockerExistingImage {
     @Optional
     final Property<Long> shmSize = project.objects.property(Long)
 
-    /* 
+    /*
      * Automatically remove the container when the container's process exits.
      *
      * This has no effect if {@link #restartPolicy} is set.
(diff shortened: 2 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -105,6 +105,31 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
...
+
+        buildFile <<
+                containerStart(containerCreationTask) <<
+                containerLogAndRemove()
+
+        when:
+        BuildResult result = build('logContainer')
+
+        then:
+        result.output.contains("tmpfs $tmpFsMount")
```
orzeh: IMO it would be better to check if tmpfs mounts are returned when inspecting container like in [this test](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy#L24).

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -452,6 +457,10 @@ class DockerCreateContainer extends DockerExistingImage {
         if(macAddress.getOrNull()) {
             containerCommand.withMacAddress(macAddress.get())
         }
+
+        if(tmpfs.getOrNull()) {
+            containerCommand.hostConfig.withTmpFs(tmpfs.get().collectEntries { [it, ""] })
```
orzeh: Wouldn't be better to allow users to provide also mount options?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
thyming: sure, that's better.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
thyming: you mean in addition to tmpfs or in place of that?

orzeh: In addition. Then the new property should be of type `Porperty<Map<String, String>>` and have a JavaDoc explaining what are the keys and values.

thyming: multiple mounts can be specified, though, so it also needs to be a list. i guess the map you had in mind was of the kv pairs that get passed as options to mount?

bmuschko: I suggest you implement a more complex data structure to cover all possible cases similar to what Docker Java does. Please introduce your own classes if you can so that the Docker Java classes don't bleed into the public API of the plugin if possible.

bmuschko: @thyming Are you still planning to finish this PR?

thyming: I started down the path of implementing functionality equivalent to docker cli `--mount` but found that there was quite a lot of parsing logic of the options not implemented in docker-java (as it is for ports, volumes, etc) that I didn't really have time to investigate/implement. A quick hack at it led to nasty mapping code between strings and docker-api's enums without bringing in jackson as a direct dependency.

bmuschko: Is it correct to assume that you won't have the time to follow up? I am going to close the PR soon if that's the case as the solution is not sufficient. Thanks again for your contribution.

thyming: No, I don't have time to work through a full support of the mount flag. However, I'm not sure why it's not worth merging the tmpfs option, which, while less fully featured than the mount option, is still an element of the CLI and Java API and so seems reasonable to expose here.

bmuschko: The data structure should reflect something we can enhance in the future. If we want to add something new to what you propose then this would likely change the public API of the plugin and therefore break anyone using the new property.

It's not a matter of introducing this configuring option in general. It's a matter of maintainability and future enhancement without breaking existing consumers.

thyming: In that case, I will close it.

Issue #770, opened 2019-03-16
{"message":"invalid reference format"} in DockerCreateContainer after upgrade to 4.6.2


Hubbitus:
Follow [base documentation instruction](https://bmuschko.github.io/gradle-docker-plugin/#linking_with_other_containers):
```gradle
plugins {
    id 'com.bmuschko.docker-remote-api' version '4.6.2'
}
task buildEgaisappImage(type: DockerBuildImage, dependsOn: createEgaisAppDockerfile) {
    inputDir = file(createEgaisAppDockerfile.destFile).parentFile
    remove = true
    pull = true
    ext.buildTagName = versioning.info.full.toLowerCase().replaceAll(/[^a-z0-9-_.]/, '_')
    tags = ["docreg.taskdata.work/egaisapp:${buildTagName}"]
}
task createEgaisdbContainer(type: DockerCreateContainer, dependsOn: buildEgaisappImage) {
    targetImageId 'docreg.taskdata.work/egaisdb:latest'
    portBindings = [ '15432:5432' ]
    autoRemove = true
    containerName = ("egaisdb_auto_${buildEgaisappImage.buildTagName}" as String)
}
task startEgaisdbContainer(type: DockerStartContainer, dependsOn: createEgaisdbContainer){
    targetContainerId createEgaisdbContainer.getContainerId()
}
...
```
On run got exception: `Execution failed for task ':createEgaisappContainer'.
> {"message":"invalid reference format"}`

### Current Behavior
Full trace:
```log
> Task :startEgaisdbContainer
Starting container with ID 'da7aabd3c7e5fc6bfaab127717ac253507b1a903fcb4dd832a33457c93641b2e'.

> Task :createEgaisappContainer FAILED
== Build Time Summary ==
                               0% :portal-api:core:common2java (0:00.061)
                               0% :createEgaisappContainer (0:00.061)
                               0% :portal-api:core:compileJava (0:00.085)
                               0% :portal-ui:sencha_-init (0:00.176)
                               0% :portal-ui:nodeSetup (0:00.190)
                               0% :portal-ui:extractSenchaCmd (0:00.232)
                               0% :portal-ui:sencha_init-cmd (0:00.319)
                               0% :portal-ui:npmSetup (0:00.328)
                               0% :portal-ui:downloadSenchaCmd (0:00.422)
                               0% :startEgaisdbContainer (0:00.947)
                               0% :portal-ui:reactNpmInstall (0:01.026)
[...]

PR #773, opened 2019-03-25
Bump docker-java to 3.1.1

cdancy:


build.gradle.kts
```diff
@@ -24,7 +24,7 @@ repositories {
 }
 
 dependencies {
-    shaded("com.github.docker-java:docker-java:3.1.0-rc-7")
+    shaded("com.github.docker-java:docker-java:3.1.1")
```
cdancy: @bmuschko very nifty what you did with the `shaded` config.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -60,28 +53,4 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
-                tags.add("${IMAGE_ID}")
-                labels = ["setup":"${IMAGE_ID}"]
-            }
-            
-            task removeImage(type: DockerRemoveImage) {
-                force = true
-                targetImageId buildImage.getImageId()
-            }
-        """
-    }
```
cdancy: Building this image was failing with the same pop that was discussed in #770 . Not sure what's up with that just yet but using the alpine image worked as expected and vetted the test properly.

PR #781, opened 2019-03-28
fix: Fail if the spring application class can't be found unambiguously

### LINKED ISSUE #778: Ambiguous result of DockerSpringBootApplicationPlugin.getSpringApplicationMainClassName
Hello,

thank you for the plugin!

I think, it has a potential issue. In the file **\src\main\groovy\com\bmuschko\gradle\docker\DockerSpringBootApplicationPlugin.groovy** there is the method **getSpringApplicationMainClassName**.

Its result depends on the order of the class dirs, because the method returns the first class with a 'SpringBootApplication' annotation.

To make the result not ambiguous, _all_ classes with this annotation should be found, then a check made that there is exactly one of them (failing the build when there is not).

fml2:
closes #778

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy
@@ -175,14 +175,23 @@ class DockerSpringBootApplicationPlugin implements Plugin<Project> {
     }
 
     private static String getSpringApplicationMainClassName(Project project) {
+        List<String> springMainClasses = new ArrayList<>()
+        
         for (File classesDir : getMainJavaSourceSetOutput(project).classesDirs) {
             String mainClassName = MainClassFinder.findSingleMainClass(classesDir, 'org.springframework.boot.autoconfigure.SpringBootApplication')
 
-            if (mainClassName) {
-                return mainClassName
+            if (!mainClassName) {
+            	springMainClasses.add(mainClassName);
             }
         }
+        
+        if (springMainClasses.size() == 1) {
+        	return springMainClasses[0]
+        } else if (springMainClasses.size() == 0) {
+        	throw new IllegalStateException('No Spring Application class found')
+        } else { // springMainClasses.size() > 1
+        	throw new IllegalStateException("Multiple Spring Application classes found: ${springMainClasses}")
+        }
 
-        throw new IllegalStateException('Main class name could not be resolved')
     }
 }


cdancy:
Is it _possible_, or somewhat easy/sane, to write a functional test to exercise this new code?

fml2:
It is definitely possible (everything is possible :-) ), but I don't know how to do it, and will probably not do it.

bmuschko:
One of the existing test cases is already failing with this change so we'll definitely want another one. See https://scans.gradle.com/s/h2jskjg6tksrk/tests/dqwcqqtprpdqc-5upc2rraveyzo for more info.

fml2:
I must have understood the possible contexts of this tasl wrong.

cdancy:
>I must have understood the possible contexts of this tasl wrong.

@fml2 what do you mean?

bmuschko:
Looking more deeply at your change I'd also suggest that we move the code to `MainClassFinder` and not the Gradle plugin implementation. Given that you are not planning to write a test probably the best course of action for now is to close this pull request. I am planning to create a test class for `MainClassFinder`. We can enhance functionality based on that.

bmuschko:
I just added a test class for the main class finder implementation: https://github.com/bmuschko/gradle-docker-plugin/commit/cf6b0c13a235aab758f80eb0841828eed0f7de23. Would you mind adding a test to it for your change?

fml2:
I can only say: cool! I've never seriously written code in groovy, and find the tests quite readable and understandable.

But the more I think about it the more I come to the conclusion that I don't fully understand how it should work.

Now I don't quite get why the main class shoud be searched for at all for a spring boot app. At least for one with "fat jar". Why not just extract the information from the manifest of the jar file (which was appropriately created by the native spring boot plugin, i.e. the bootJar task)?

Also, in my understanding, the case "returns null if no main class can be found" should also throw an exception. The objective is to find exactly one class with the main method and annotated with the specified annotation. If that can't be done (for whatever reason), an exception should be thrown. This is how I'd see it. But this is apparently wrong.

Hence I don't feel entitled to write tests because my understanding of how thigs work is incomplete.

fml2:
> 
> 
> > I must have understood the possible contexts of this tasl wrong.
> 
> @fml2 what do you mean?

Judging by the failing test, there are valid situations where no main class is found. I thought, in this case an exception should be thrown. That's why I think I have not understood how it's supposed to work.

bmuschko:
> Now I don't quite get why the main class shoud be searched for at all for a spring boot app. At least for one with "fat jar". Why not just extract the information from the manifest of the jar file (which was appropriately created by the native spring boot plugin, i.e. the bootJar task)?

I took the implementation from the Spring Boot libraries so the behavior should be somewhat similar. A manifest entry could be an option but I wasn't sure if that it always populated e.g. if you generate a WAR file instead.

> Also, in my understanding, the case "returns null if no main class can be found" should also throw an exception. The objective is to find exactly one class with the main method and annotated with the specified annotation. If that can't be done (for whatever reason), an exception should be thrown. This is how I'd see it. But this is apparently wrong.

I wrote the code while in transit and didn't have the time to polish everything. I actually thought the same thing when writing the this test. For now I just documented the current behavior. It makes little sense to return `null` as it would mess up the Docker image when you start a container. We should throw an exception here. Will change the behavior in the next couple of days.

fml2:
> Looking more deeply at your change I'd also suggest that we move the code to `MainClassFinder` and not the Gradle plugin implementation

To my (apparently wrong) understanding, one should not do this, because the plugin calls MainClassFinder multiple times. A single main class should be found "globally" and not per MainClassFinder call.

bmuschko:
The current implementation uses the first main class implementation that can be found on the class path. Be aware that you can define multiple source directories for a source set. That's why we have to iterate over the different classes directories [here](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy#L178). I don't necessarily see a big issue here. It's unlikely that you have two different main source directories except maybe in a legacy application. And apart from that...why would you have multiple main classes? It's an unlikely use case.

I am going to close the issue and the pull request for now. If we have someone that actually runs into the issue then we can discuss further based on their requirements. Thanks again for discussion and the proposed PR.

fml2:
Thank you for your patience with me :-)

bmuschko:
@fml2 All good. Thanks for participating.

Issue #725, opened 2018-12-18
Deprecation warning regarding Gradle 5.0 when using plugin


picaron:
### Expected Behavior

The plugin should continue working when migrating to Gradle 6.0

### Current Behavior

Currently when using Gradle 5.0, we get the following deprecation warning:

```
The DefaultSourceDirectorySet constructor has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory service to create instances of SourceDirectorySet instead.
The ProjectLayout.directoryProperty() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.directoryProperty() method instead.
        at build_9t0xk9xnqi0dg4yogioioq88k.run(/Users/pascal.chouinard/projects/product-management2/build.gradle:32)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The DefaultTask.newOutputFile() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.fileProperty() method instead.
        at build_9t0xk9xnqi0dg4yogioioq88k.run(/Users/pascal.chouinard/projects/product-management2/build.gradle:207)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The DefaultTask.newInputDirectory() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.directoryProperty() method instead.
        at build_9t0xk9xnqi0dg4yogioioq88k.run(/Users/pascal.chouinard/projects/product-management2/build.gradle:221)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
The DefaultTask.newInputFile() method has been deprecated. This is scheduled to be removed in Gradle 6.0. Please use the ObjectFactory.fileProperty() method instead.
        at build_9t0xk9xnqi0dg4yogioioq88k.run(/Users/pascal.chouinard/projects/product-management2/build.gradle:221)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

### Context

I want my builds to keep working when upgrading to Gradle 6.0

### Steps to Reproduce (for bugs)

Use the gradle wrapper 5.0 and the gradle docker plugin in a project to see the warning

### Your Environment

Using gradle wrapper 5.0

bmuschko:
[...]

PR #782, opened 2019-04-02
Fix Gradle 5.x deprecation warnings

tlusk:
This fixes issue #725

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -29,6 +29,7 @@ import com.github.dockerjava.api.model.RestartPolicy
 import com.github.dockerjava.api.model.Volume
 import com.github.dockerjava.api.model.VolumesFrom
 import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.MapProperty
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
@@ -110,7 +111,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> envVars = project.objects.property(Map)
+    final MapProperty<String, String> envVars = project.objects.mapProperty(String, String)
 
     @Input
     @Optional
@@ -153,7 +154,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> binds = project.objects.property(Map)
+    final MapProperty<String, String> binds = project.objects.mapProperty(String, String)
 
     @Input
     @Optional
@@ -204,7 +205,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> labels = project.objects.property(Map)
+    final MapProperty<String, String> labels = project.objects.mapProperty(String, String)
 
     @Internal
     final Property<String> containerId = project.objects.property(String)
@@ -265,7 +266,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     void withEnvVar(def key, def value) {
         if (envVars.getOrNull()) {
-            envVars.get().put(key, value)
+            envVars.put(key, value)
         } else {
             envVars.set([(key): value])
         }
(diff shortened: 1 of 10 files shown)


cdancy:
I was actually going to do something similar and had at least one thing in mind I wanted to change. But since you're here.... ;)

Would you mind searching through our tasks to see where we do things like `project.objects.property(Map)` or something to that effect and replace with the new `project.objects.mapProperty(String, String)`?

tlusk:
@cdancy yup, I just found that last warning during some testing (hence draft PR).  Just made the change to use `mapProperty` but it's failing some of the tests now so I'm investigating.

I'll update the PR with those changes once I get them working!

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -265,7 +266,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     void withEnvVar(def key, def value) {
         if (envVars.getOrNull()) {
-            envVars.get().put(key, value)
+            envVars.put(key, value)
```
tlusk: `MapProperty.get()` returns an `ImmutableMap` copy of the property.  Need to call `MapProperty.put()` directly in order to modify the property.

tlusk:
@cdancy all set, finished testing and not seeing any more deprecation warnings.

cdancy:
@tlusk just wondering: were there any test failures locally when running `./gradlew functionalTest`?

tlusk:
@cdancy no failures, functional tests were all clean.

cdancy:
@tlusk sounds good. CI hadn't kicked in here yet (just did now) and so was just wondering if anything ... _interesting_ ... popped.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -29,6 +29,7 @@ import com.github.dockerjava.api.model.RestartPolicy
 import com.github.dockerjava.api.model.Volume
 import com.github.dockerjava.api.model.VolumesFrom
 import org.gradle.api.provider.ListProperty
+import org.gradle.api.provider.MapProperty
```
bmuschko: The use of `MapProperty` raises the minimum compatible Gradle version to 5.1. Please update the note in the user guide to reflect that.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -110,7 +111,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> envVars = project.objects.property(Map)
+    final MapProperty<Object, Object> envVars = project.objects.mapProperty(Object, Object)
```
bmuschko: This should define the concrete types. Let's use String/String here and in the code that passes the key/value pairs to Docker Java, we'll want to remove the support for Closures. There's no more need for Closures as the Property takes care of lazy evaluation.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -153,7 +154,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> binds = project.objects.property(Map)
+    final MapProperty<Object, Object> binds = project.objects.mapProperty(Object, Object)
```
bmuschko: Same comment as above in regard to typing.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
```diff
@@ -204,7 +205,7 @@ class DockerCreateContainer extends DockerExistingImage {
 
     @Input
     @Optional
-    final Property<Map> labels = project.objects.property(Map)
+    final MapProperty<Object, Object> labels = project.objects.mapProperty(Object, Object)
```
bmuschko: Same comment as above in regard to typing.

tlusk:
@bmuschko requested changes are complete.

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
tlusk: I did wonder about the map being untyped, but I originally decided to leave it alone in the event there was an intentional reason for it.

cdancy:
Taking this in now. Thanks for the cleanup/fixes @tlusk !

Issue #786, opened 2019-04-08
Upgrade to latest docker-java client


padas2:
Our plugins are making use of gradle-docker-plugin to launch containers in parallel.
When launching in parallel, we are facing [this]( https://github.com/kohlschutter/junixsocket/issues/45) issue .

For this to be fixed, the docker-java client needs to be upgraded to use the latest junix dependencies i.e. 2.2.0 . This is already included in the latest release 3.1.2 [Link](https://github.com/docker-java/docker-java/milestone/34?closed=1).

We need the gradle-docker-plugin to use this version of docker-java client.

cdancy:
I'll spin up a PR and get this in for the next release (assuming no issues) which we are due for. Stay tuned...

cdancy:
Change has been merged to master.

padas2:
@cdancy When can we get a new version of this plugin published ?

PR #787, opened 2019-04-08
Update build.gradle.kts

cdancy:
Fix for #786


cdancy:
Failure above was re-run and passed on second time. Feel like we've seen this flaky pop before. In either event change LGTM.

PR #789, opened 2019-04-13
Rename timeout properties to waitTime so as not to clobber the exiting timeout property on gradles Task.

cdancy:
More fallout from: https://docs.gradle.org/5.0/userguide/more_about_tasks.html#sec:task_timeouts

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerWorkflowFunctionalTest.groovy
@@ -21,7 +21,7 @@ import spock.lang.Requires
 
 class DockerWorkflowFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
-    def "Can build an image, create and start a container"() {
+    def "Can build an image, create, start, stop, and restart container"() {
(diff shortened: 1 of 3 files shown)


cdancy:
@bmuschko looks like you knocked out some of the fallout a few releases ago just missed these.

bmuschko:
@cdancy Looks like you change broke the build: https://scans.gradle.com/s/4cb7gtdv24uc4/tests/dqwcqqtprpdqc-sbol5sl7njrcg. Can you please fix?

cdancy:
@bmuschko hmmm ... I did see that but re-kicked things again and was able to do a release. Let me look at things as this, like all things, wasn't failing locally :)

bmuschko:
@cdancy Any idea when you'll be able to look at it? Have a PR waiting and I don't want to it fail on CI.

cdancy:
@bmuschko I _think_ it should be good to go?!? I was able to re-run the job and it passed as expected. The task itself catches the noted exception and swallows it so in theory things should be fine.

bmuschko:
@cdancy The test probably became flaky or there's some sort of timing issue. I will let you know if it happens again.

cdancy:
Sounds good. I was thinking maybe it hit a commit that I over wrote. IDK.

Issue #761, opened 2019-03-07
com.bmuschko.docker-java-application - not able to set spring profiles


Annie0507:
We upgraded our version of 'com.bmuschko.docker-java-application' to '4.6.0' and Spring Boot version 2.1.3.RELEASE and gradle version gradle-5.2.1-all and when we are trying to run the application, we are not able to set the profiles. 

We have tried the following things: 
docker run {image_id} --spring.profiles.active=dev

dockerCreateDockerfile {
    environmentVariable 'JAVA_OPTS', '-Dspring.profiles.active=dev'
}

and also,
docker {
   springBootApplication {
        args = ['-Dspring.profiles.active=dev'] - Complains that args is not a valid argument.
   }
}

Before the upgrade, it was picking up the profile from helm templates correctly.

### Expected Behavior
When the app is deployed to dev, it should pick up the env and setup the profile.

### Current Behavior
No active profile set, falling back to default profiles: default

### Context
When out application is deployed to an environment, instead of picking up the correct application.properties file, it is reading the variables from the local application.properties.

### Steps to Reproduce (for bugs)
plugins {
    id 'java'
    id 'application'
    id 'org.springframework.boot' version '2.1.3.RELEASE'
    id 'com.bmuschko.docker-java-application' version '4.6.0'
}

mainClassName = "xxx"

docker {
    url = lookupDockerHost()
    certPath = lookupSslKeysDirectory()
    registryCredentials {
        url = registry
        username = lookupDockerRegistryUser()
        password = lookupDockerRegistryPassword()
    }

    javaApplication {
        baseImage = "${dockerBaseImage}"
        maintainer = "${teamName} ${teamEmail}"
        tag = registry + "/${group}/${name}:${project.version}"
    }
}
Then run the command:

./gradlew :{project-name}:dockerBuildImage 

then docker run with the image that was build:

docker run {image-id} ----spring.profiles.active=dev

### Your Environment
Dev, Perf, Prod

bmuschko:
Please find the details on why this changed in https://github.com/bmuschko/gradle-docker-plugin/pull/736. We are not using the generated startup script anymore provided by Gradle's application plugin. It internally sets `JAVA_OPTS` before executing the `java` command.

```
dockerCreateDockerfile.environmentVariable 'JAVA_OPTS', '-Dspring.profiles.active=dev'
```

[...]

PR #790, opened 2019-04-15
Expose extension property for configuring JVM arguments

bmuschko:
Essentially passes the JVM arguments to `java` command in the generated Dockerfile (see https://github.com/bmuschko/gradle-docker-plugin/issues/761). This PR also adds more Javadocs to the extensions.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
@@ -113,6 +113,7 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
                     }))
                     copyFile('classes', 'classes/')
                     instructions.add(dockerJavaApplication.execInstruction)
+                    defaultCommand(dockerJavaApplication.jvmArgs)
                     exposePort(dockerJavaApplication.ports)
                 }
             }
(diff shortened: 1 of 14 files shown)


bmuschko:
@cdancy Passing JVM arguments doesn't work with a CMD instruction. The JVM arguments have to be defined before the main class. Will need to change the code.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy
```diff
@@ -58,7 +58,15 @@ class DockerJavaApplicationPlugin implements Plugin<Project> {
...
-                            ["java", "-cp", "/app/resources:/app/classes:/app/libs/*", getApplicationPluginMainClassName(project)]
+                            List<String> entrypoint = ["java"]
+                            List<String> jvmArgs = dockerJavaApplication.jvmArgs.get()
+
+                            if (!jvmArgs.empty) {
+                                entrypoint.addAll(jvmArgs)
+                            }
+
+                            entrypoint.addAll(["-cp", "/app/resources:/app/classes:/app/libs/*", getApplicationPluginMainClassName(project)])
+                            entrypoint
```
rycee: I'm curious, as a Groovy newbie, why not reduce these lines to
```groovy
[
    "java",
    *dockerJavaApplication.jvmArgs.get(),
    "-cp", "/app/resources:/app/classes:/app/libs/*",
    getApplicationPluginMainClassName(project)
]
```
?

bmuschko: Achieves the same thing with less code. I noticed that developers with little Groovy and/or Gradle experience don't understand the meaning of DSL or language-specific symbolism. The spread operator is great to make things more concise but confuses users that don't know what it means (and the implication of handling empty collections). Feel free to submit a PR if you feel strongly about it.

rycee: I was mostly curious since I've recently started using Groovy and haven't yet developed an intuition for these things. So I very much appreciate that you took the time to describe your reasoning.

In any case I think the existing code is just fine and I suspect it and the example I gave above will compile to basically the same thing anyway.

Thanks a lot for developing this plugin, it's been a great help!

Issue #800, opened 2019-05-06
DockerBuildImage should be incremental and cacheable


llaforest:
### Expected Behavior
This feature is related to the one implemented in #671 and rolled-back in #672. Define a File that stores the image id in a file. Annotate the field with @OutputFile so that Gradle can build the task incrementally. Add @Cacheable annotation to the class to make enable it for use with the Gradle build cache. Ensure that all File paths are stored with its relative path by adding @PathSensitive(PathSensitivity.RELATIVE) annotation. Add an output check to ensure the image is present in the docker registry using outputs.upToDateWhen method before declaring it UP-TO-DATE.

### Current Behavior
Executing a task of type DockerBuildImage re-runs its action every time. Building more elaborate images takes longer which is unnecessary. That's especially true in a CI pipeline where "building an image" and "pushing an image" are separated into individual build steps. If "pushing an image"depends on "building an image", then the later step will rebuild the image.

### Context
Make use of incremental build functionality and the Gradle build cache to improve build times.

bmuschko:
The first step in working toward the functionality needs to make the Docker client available to the constructor of a task in some form. Right now it is created at the time of executing the task action but `outputs.upToDateWhen` can only be defined in the constructor of a custom task type.

llaforest:
My goal was to have this call inside your plugin. I did not want to have logic for the cache outside your code. I searched and found this but did not have time to test yet:
https://stackoverflow.com/questions/17793354/gradle-plugin-test-uptodatewhen-method
From inside your plugin, the Docker client is available... So if we could have the outputs.upToDate call there it would be just another step toward caching...

bmuschko:
> My goal was to have this call inside your plugin. I did not want to have logic for the cache outside your code.

We are on the same page. If we provide the functionality then it should be built into the plugin code. It doesn't change how we need to implement it though. The Docker Client currently is not available in the constructor of any of the custom task types provided by the plugin. I am talking about an incremental step toward enabling the functionality.

llaforest:
So I did some investigation and I might need some advice.
One solution to have the DockerClient inside the DockerBuildImage constructor is to have both getDockerClient and getDockerHostUrl static so they can be called from the constructor of DockerBuildImage. The drawback here is that we cannot support custom url for the DockerRepository since the property is not set yet when the constructor runs...

[...]

PR #802, opened 2019-05-09
Initialize and return Docker Java client on demand

bmuschko:
Internal refactoring to implement further changes described in https://github.com/bmuschko/gradle-docker-plugin/issues/800

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/DockerVersion.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/DockerVersion.groovy
@@ -15,12 +15,12 @@
  */
 package com.bmuschko.gradle.docker.tasks
 
-import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.model.Version
 
 class DockerVersion extends AbstractDockerRemoteApiTask {
+
     @Override
-    void runRemoteCommand(DockerClient dockerClient) {
+    void runRemoteCommand() {
         logger.quiet "Retrieving Docker version."
         Version version = dockerClient.versionCmd().exec()
         logger.quiet "Version          : $version.version"
--- a/src/docs/samples/code/remote-api-plugin/custom-task-type/groovy/build.gradle
+++ b/src/docs/samples/code/remote-api-plugin/custom-task-type/groovy/build.gradle
@@ -40,7 +40,7 @@ class DockerImageIdForName extends AbstractDockerRemoteApiTask {
     }
 
     @Override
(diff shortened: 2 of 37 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/DockerVersion.groovy
```diff
@@ -15,12 +15,12 @@
...
-import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.model.Version
 
 class DockerVersion extends AbstractDockerRemoteApiTask {
+
     @Override
-    void runRemoteCommand(DockerClient dockerClient) {
+    void runRemoteCommand() {
         logger.quiet "Retrieving Docker version."
         Version version = dockerClient.versionCmd().exec()
```
cdancy: Should this, and all other calls below, use `getDockerClient()` if only to keep the way we access things standard across the board?

bmuschko: This is just Groovy syntactic sugar. It calls the getter method under the hood. If you rather want the Java style then we can make that happen though there's no functional benefit. I am not sure if we are even consistent across the board. It's unlikely.

cdancy: The only reason I note that is because the variable is also named `dockerClient`. I've no idea how groovy will make things work guess I'm just thinking it _may_ want to go directly to the variable instead of hitting the method? If that's not the case then disregard this comment.

bmuschko: It calls the getter. Also the `dockerClient` variable is `private` so the subclasses can't access it.

cdancy: > private so the subclasses can't access 

In java maybe but with groovy ALL things are possible ;)

llaforest: I was wondering why remove the @Memoize annotation when creating dockerClient. I thought it was nice that the dockerClient gets reused for all the different needs as long as it is the same which is usually the case within the same program...

bmuschko: Yeah, I guess we already used it that way for quite a while. I was concerned about changing values between build invocations. I am not sure how Gradle caches a plugin implementation which could lead to incorrect results. I restored it.

PR #806, opened 2019-05-11
Remove exec/cmd hooks in Docker application plugin

bmuschko:
I am not convinced that those hooks are needed anymore now that we have a `jvmArgs` options. This was mostly used for the Application plugin that used to generate a start script. Our plugin doesn't use the script anymore. We are building the Java command manually to leverage layer caching.

If you still want to modify the instructions of the `Dockerfile` task then you can as described here: https://bmuschko.github.io/gradle-docker-plugin/#modifying_instructions_of_a_dockerfile_task.

One reason for removing the code is that it makes the internal far more complex than it should be. It will help us with abstracting more logic in the future. There's also an obvious gap in functionality between the Docker Application and the Spring Boot plugin. This change closes the gap by minimizing the configuration options.

Please describe your use case if you think that you still need this functionality!


bmuschko:
@cdancy Thanks. I will leave it open for a little bit so that other users can have a look at it and form an opinion.

orzeh:
LGTM, however it requires major version upgrade since it will break existing builds?

bmuschko:
@orzeh That's a good idea. We should probably batch up a couple of breaking change and release them with 5.0.0.

PR #807, opened 2019-05-13
Reinstate ImageId output file and check in Docker registry

llaforest:
Rebased on latest master, related to issue #800 incremental part.

With this code I think the task is incremental. Please review my logic but now it checks both the file and the Docker registry. My code runs in afterEvaluate closure so that internal properties of the class have the time to get populated. 

This will not do it to have the task cacheable. Still trying to figure out a way to do it.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -137,16 +137,41 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         noExceptionThrown()
     }
 
-    @PendingFeature
     def "task can be up-to-date"() {
         given:
         buildFile << imageCreation()
         buildFile << """
             task verifyImageId {
                 dependsOn buildImage
-                doLast {
-                    assert buildImage.imageId.isPresent()
-                }
+            }
+        """
+        File imageIdFile = new File(projectDir, 'build/.docker/buildImage-imageId.txt')
+
+        when:
+        BuildResult result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+
+        when:
+        result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.UP_TO_DATE
+        !result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+    }
+
+    def "task not up-to-date when no imageIdFile"() {
+        given:
+        buildFile << imageCreation()
+        buildFile << """
+            task verifyImageId {
+                dependsOn buildImage
             }
         """
         File imageIdFile = new File(projectDir, 'build/.docker/buildImage-imageId.txt')
@@ -180,6 +205,91 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         imageIdFile.text != ""
     }
 
+    def "task not up-to-date when image not in registry"() {
+        given:
+        buildFile << imageCreation()
+        buildFile << """
+            task verifyImageId {
+                dependsOn buildImage
+            }
+        """
+        File imageIdFile = new File(projectDir, 'build/.docker/buildImage-imageId.txt')
+
+        when:
+        BuildResult result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+
+        when:
+        result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.UP_TO_DATE
+        !result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+
+        when:
+        buildFile << imageRemoval(imageIdFile.text)
+        result = build('removeImage')
+
+        then:
+        result.task(':removeImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Removing image with ID")
+
+        when:
+        result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+    }
+
+    def "task not up-to-date when imageIdFile changed"() {
+        given:
+        buildFile << imageCreation()
+        buildFile << """
+            task verifyImageId {
+                dependsOn buildImage
+            }
+        """
+        File imageIdFile = new File(projectDir, 'build/.docker/buildImage-imageId.txt')
+
+        when:
+        BuildResult result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+
+        when:
+        result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.UP_TO_DATE
+        !result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+
+        when:
+        imageIdFile << " "
+        result = build('verifyImageId')
+
+        then:
+        result.task(':buildImage').outcome == TaskOutcome.SUCCESS
+        result.output.contains("Created image with ID")
+        imageIdFile.isFile()
+        imageIdFile.text != ""
+    }
+
     private static String buildImageWithShmSize() {
         """
             import com.bmuschko.gradle.docker.tasks.image.Dockerfile
@@ -203,10 +313,22 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
             task dockerfile(type: Dockerfile) {
                 from '$TEST_IMAGE_WITH_TAG'
+                runCommand('pwd')
             }
 
             task buildImage(type: DockerBuildImage) {
                 dependsOn dockerfile
+            }
+        """
+    }
+
+    private static String imageRemoval(String imageId) {
+        """
+            import com.bmuschko.gradle.docker.tasks.image.DockerRemoveImage
+
+            task removeImage(type: DockerRemoveImage) {
+                force = true
+                targetImageId '$imageId'
             }
         """
     }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -18,10 +18,13 @@ package com.bmuschko.gradle.docker.tasks.image
 import com.bmuschko.gradle.docker.DockerRegistryCredentials
 import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
 import com.bmuschko.gradle.docker.tasks.RegistryCredentialsAware
+import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.command.BuildImageCmd
+import com.github.dockerjava.api.command.ListImagesCmd
 import com.github.dockerjava.api.model.AuthConfig
 import com.github.dockerjava.api.model.AuthConfigurations
 import com.github.dockerjava.api.model.BuildResponseItem
+import com.github.dockerjava.api.model.Image
 import com.github.dockerjava.core.command.BuildImageResultCallback
 import org.gradle.api.file.DirectoryProperty
 import org.gradle.api.file.RegularFileProperty
@@ -35,6 +38,7 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.OutputFile
 
 class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCredentialsAware {
 
@@ -107,6 +111,14 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
      */
     DockerRegistryCredentials registryCredentials
 
+    /**
+     * Output file containing the image ID of the built image. 
+     * Defaults to "$buildDir/.docker/$taskpath-imageId.txt".
+     */
+    @OutputFile
+    @PathSensitive(PathSensitivity.RELATIVE)
+    final RegularFileProperty imageIdFile = newOutputFile()
+
     @Internal
     final Property<String> imageId = project.objects.property(String)
 
@@ -118,6 +130,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(new File(project.buildDir, ".docker/${safeTaskPath}-imageId.txt"))
+
+        if(imageIdFile.get().asFile.exists()){
+            // Get all images in docker registry and make sure the imageId is found in there
+            outputs.upToDateWhen{getDockerClient().listImagesCmd().exec().any{image -> image.toString().contains(imageIdFile.get().asFile.text)}}
+        }     
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -118,6 +130,15 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(new File(project.buildDir, ".docker/${safeTaskPath}-imageId.txt"))
+
+        project.afterEvaluate {
```
bmuschko: Heads up: We definitely don't want a `afterEvaluate`. It's also not required here as the Closure of `outputs.upToDateWhen` delays the evaluation.

llaforest: If the code runs outside the afterEvaluate, the class members such as url, certPath or apiVersion will not be set yet and the dockerClient might not represent the registry specified... So delaying the evaluation until the class is populated seems required to me.
Do you see some way around this?

llaforest:
Yes I will definitely add some tests. Maybe I should have done them before but I wanted you guys to validate my logic before.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: I already test cases for that so that should work as-is: https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTaskFunctionalTest.groovy#L87-L107.

llaforest: Ok removed the outputs.upToDateWhen and it seems the local variable have the time to get set. Did not know the upToDateWhen closure delayed evaluation.

llaforest:
Some tests are failing when building locally and it blocks the remaining of the test sequence. Any clue how to run "src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy" only without running all the rest?
I updated "task can be up-to-date" and added removal of the image in the docker registry to show it triggers a rebuild.

Test failing: com.bmuschko.gradle.docker.DockerWorkflowFunctionalTest - Stop container

cdancy:
@llaforest `./gradlew functionalTest --tests **DockerBuildImage**` should do the trick

llaforest:
Thank you @cdancy it works fine. I'm struggling to get something that works fine.

I have one question though... When I run the test, there is an image that gets created in my docker registry but the imageId returned by the DockerBuildImage is always the alpine one (b7c5...), not the last one that was created with the test (d87f...). By the way my repo was completely empty before executing DockerBuildImageFunctionalTest. I'm not sure I understand what is happening here can you help me?
```
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
<none>                         <none>              d87fa0353cd8        13 seconds ago      4.82MB
<none>                         <none>              96e91910db63        14 seconds ago      4.82MB
alpine                         latest              055936d39205        3 days ago          5.53MB
alpine                         3.4                 b7c5ffe56db7        3 months ago        4.82MB
test/image                     123                 b7c5ffe56db7        3 months ago        4.82MB
registry.com:5000/test/image   123                 b7c5ffe56db7        3 months ago        4.82MB
```

bmuschko:
@llaforest Some functional tests don't properly clean up the resource they created. We are aware of it and I am planning to work on it in the near future. That's probably the reason you see containers or images that don't necessarily belong to your test case.

llaforest:
Yeah cleanup is one thing but here the problem is that the image created was d87fa0353cd8 but it tells me that it was b7c5ffe56db7. Clearly b7c5ffe56db7 does not come from me, it was pulled. So now when I try to remove image b7c5ffe56db7 it fails because it has some children. It's like the DockerBuildImage returns me the parent image and not the image that was created is that possible?

llaforest:
Hello I possibly found the problem. Will come back later with explanation.

llaforest:
My problem was that my test was using alpine:3.4 as is so when trying to delete, it would complaint because some other test scenarios built images having alpine:3.4 as parent... I added a layer in the imageCreation task so that it does not take alpine:3.4 directly.

llaforest:
Now is it ready to be merged? I'm not sure if the action is on my side to resolve the 'Changes requested' @bmuschko added to the merge request. I can't find where to do it.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -11,6 +11,8 @@ import spock.lang.Unroll
 
 class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
+    private String lastImageId
```
bmuschko: This shouldn't be a class variable. It should be inlined into the test method.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -137,15 +139,14 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
-    @PendingFeature
     def "task can be up-to-date"() {
         given:
         buildFile << imageCreation()
         buildFile << """
             task verifyImageId {
                 dependsOn buildImage
                 doLast {
-                    assert buildImage.imageId.isPresent()
+                    assert buildImage.imageIdFile.get().asFile.exists()
```
bmuschko: Why was this changed exactly? We are testing this already in the assertions as far as I can see: `imageIdFile.isFile()`.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -137,15 +139,14 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         noExceptionThrown()
     }
 
-    @PendingFeature
     def "task can be up-to-date"() {
```
bmuschko: We'll need to break up this test method into multiple test cases that describe the specific use cases. At the moment I find it really hard to understand what situation are covered. At the very least I'd expect four different test cases:

- The image has been built. The image file exists. Upon the next execution the task is up to date.
- The image has been built. The image file exists but was deleted manually from the filesystem. Upon the next execution the task is not up to date.
- The image has been built. The image file exists but the image was removed from the registry. Upon the next execution the task is not up to date.
- The image has been built. The image file exists but the value inside of the file was changed manually. Upon the next execution the task is not up to date.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -118,6 +130,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(new File(project.buildDir, ".docker/${safeTaskPath}-imageId.txt"))
+
+        if(imageIdFile.get().asFile.exists()){
```
bmuschko: This needs to go into the Closure of `upToDate` when because a user could decide to set their own `imageIdFile` value. Calling `get()` here would evaluate the field too early.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -118,6 +130,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
...
+            // Get all images in docker registry and make sure the imageId is found in there
+            outputs.upToDateWhen{getDockerClient().listImagesCmd().exec().any{image -> image.toString().contains(imageIdFile.get().asFile.text)}}
```
bmuschko: Can you please make the Closure body a multi-line statement? At the moment it's not very readable.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -107,6 +111,14 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
      */
     DockerRegistryCredentials registryCredentials
 
+    /**
+     * Output file containing the image ID of the built image. 
+     * Defaults to "$buildDir/.docker/$taskpath-imageId.txt".
```
bmuschko: The logic turn the `:` character into an underscore. This statement isn't fully correct.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -118,6 +130,13 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(new File(project.buildDir, ".docker/${safeTaskPath}-imageId.txt"))
```
bmuschko: The recommend way to build something like this now is `imageIdFile.set(project.layout.buildDirectory.file(".docker/${safeTaskPath}-imageId.txt"))`. See https://docs.gradle.org/current/userguide/lazy_configuration.html.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -203,6 +247,7 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
             task dockerfile(type: Dockerfile) {
                 from '$TEST_IMAGE_WITH_TAG'
+                runCommand('pwd')
```
bmuschko: Can you explain why this was necessary?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
llaforest: imageId is @internal in DockerBuildImage class... I decided to leave the field visibility like you set it and check something else closely related. I am not using the imageId anywhere, but I'm using the imageIdFile so why check imageId? I can change it back to check imageId.isPresent() if you prefer. Or I can remove this assert altogether. It's true we check for the file presence just bellow.

Let me know what you prefer

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
llaforest: Ok I will move the variable inside the given: block of the test. But if I break the test in multiple tests, I will need the variable to be shared between tests. How can we manage to do that if the variable is not a class member?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
llaforest: If the file is not there, gradle will tell you the task is no up-to-date because the file is not there. When the image is not there, it will tell you that upToDateWhen failed. 
If I move everyting inside the upToDateWhen closure, there won't be a separate message for both cases. Is that ok with you?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
llaforest: I have to add another layer on top of the alpine image. If I add no layer it uses Alpine:3.4 as is and since this image was used to build some child images, I cannot remove when it's time to test deleting image and check up-to-date status.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
llaforest: Broke into the four test cases you mentioned. A lot of repetition but we need to build twice to get the up-to-date before testing to break it.

cdancy:
Assigned this to the `5.0` release as it could potentially be a breaking change.

llaforest:
Is there any timeframe for the 5.0 release? Many more issues tagged for 5.0?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -118,6 +131,17 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(project.layout.buildDirectory.file(".docker/${safeTaskPath}-imageId.txt"))
+
+        outputs.upToDateWhen {
+            if(imageIdFile.get().asFile.exists()) {
+                // Get all images in docker registry and make sure the imageId is found in there
+                getDockerClient().listImagesCmd().exec().any {
```
orzeh: I just wonder if this can hit performance if registry contains large number of images. Can we use here `DockerInspectImage` task?

The second thing is would if it work with private registries?

cdancy: Agreed. At the very least it would be cheaper all things considered vs iterating over everything.

llaforest: You're right for docker inspect. On a normal repo (600 images) it takes 0.6 seconds for the list, 0.03 for the inspect when image found. If image not present, inspect takes approx the same time than list. So I suppose we should use inspect and scan the return code.

I am not sure to understand the question for private registries...

llaforest: Here is the code with the inspect. Unfortunately, I found no other way than to use the exception handling in my normal workflow. This is how inspect is implemented. Still much faster though... Let me know what you think.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -108,6 +113,15 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
      */
     DockerRegistryCredentials registryCredentials
 
+    /**
+     * Output file containing the image ID of the built image. 
+     * Defaults to "$buildDir/.docker/$taskpath-imageId.txt".
+     * If path contains ':' it will be replaced by '_'.
+     */
```
bmuschko: Can you please also add `@since 5.0.0`?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -119,6 +133,20 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+        String safeTaskPath = path.replaceFirst("^:", "").replaceAll(":", "_")
+        imageIdFile.set(project.layout.buildDirectory.file(".docker/${safeTaskPath}-imageId.txt"))
+
+        outputs.upToDateWhen {
+            if(imageIdFile.get().asFile.exists()) {
```
bmuschko: Not super important but we could just reference `imageIdFile.get().asFile` once and then reuse it.

cdancy:
We could put this into the 4.9.0 release which I was going to knock out as soon as #811 got in. Do folks have any objections to putting this into `4.9.0` instead of `5.0.0`? On further thought I'd like to leave the latter for something bigger whatever that might be.

bmuschko:
@cdancy This change _should not_ be breaking if it works correctly so I guess 4.9.0 is fine.

bmuschko:
@cdancy Remember to change the `@since` annotation value depending on where this change lands.

llaforest:
Yeah, so will it go in 5.9 or 5.0? I can change it anytime.

bmuschko:
@llaforest We can change that post-merge.

cdancy:
Let's go 4.9.0. This will be a good win if we finally got things working and will remove lots of shenanigans we, and I have to imagine other places, are doing to sidestep this issue. I'll kick a release this weekend if no one beats me to it.

cdancy:
Thanks for the PR @llaforest !

cdancy:
@llaforest  `v4.9.0` has been released with this change.

PR #810, opened 2019-05-21
Avoid memory leakage by replacing addShutdownHook with Gradle.buildFinished

### LINKED ISSUE #809: addShutdownHook leaks Gradle classloaders
Hi @bmuschko,

While investigating something else, I found a memory leak in this plugin.

https://github.com/bmuschko/gradle-docker-plugin/blob/b80a981543538226593cdfb0319d3b0e757a71f3/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy#L164

These hooks are leaking the classloaders used to load the plugin, which can be very expensive. The only reliable solution here is to only register the hook when truly needed and remove the hook when it is no longer needed.

breskeby:
This PR fixes #809 

Replacing the `addShutdownHook` by a `Gradle.buildFinished` hook we can ensure the references is removed once the build has finished.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -160,11 +160,10 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         // Create client
         DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).build()
 
-        // register shutdown-hook to close kubernetes client.
-        addShutdownHook {
+        // register buildFinished-hook to close kubernetes client.
+        project.gradle.buildFinished {
             dockerClient.close()
         }
-
         dockerClient
     }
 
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -160,11 +160,10 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         // Create client
         DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).build()
 
-        // register shutdown-hook to close kubernetes client.
-        addShutdownHook {
+        // register buildFinished-hook to close kubernetes client.
```
cdancy: Care to fix this comment and change to `register buildFinished-hook to close docker client`?

breskeby: done

cdancy: Thanks. Always meant to get around to fixing that but then it always slips my mind.

cdancy:
Taking this in now. Thanks for the fix @breskeby !

breskeby:
thank you. Do you have an idea when the next release of this plugin (with that fix) is scheduled ?

cdancy:
We can release whenever. There is one outstanding PR that I'd like to get in. If no traction on that comes in the next couple days I'll do a release w/out it.

breskeby:
👍

cdancy:
@breskeby `v4.9.0` has been released with this change.

PR #812, opened 2019-05-25
API cleanup of Dockerfile task

bmuschko:
* Restructure some of the input types for specific instructions so that they become extendible
* Add `chown` parameter to ADD and COPY instruction
* More Javadocs


cdancy:
The changes here seem to require more typing and verbosity from the user now on top of breaking the existing API. While I'm sure this would be more friendly for IDE's and allow for code completion it could prove to be a pain in the butt.

bmuschko:
> The changes here seem to require more typing and verbosity from the user now on top of breaking the existing API.

Which change exactly? Are you talking about the new `with` builder methods for `File` and `From`? Having constructors with a lot of parameters is an anti-pattern. Just wondering...how would you provide the `chown` parameter?

The breaking change won't be an issue in general. This change will go into a major version change.

cdancy:
@bmuschko take this for example:
```
                from '$TEST_IMAGE_WITH_TAG'
                addFile('./aaa', 'aaa') // _previous_
                addFile(new Dockerfile.File('./aaa', 'aaa')) _current_
            }
```
The new way to do things is more verbose if I'm not mistaken. And that may be required it just feels a lot more to add on for the user and is not exactly user friendly.

bmuschko:
@cdancy The previous example still works even with this change.

bmuschko:
@cdancy With my previous comment do you still have the same opinion about this change?

cdancy:
@bmuschko +1. LGTM.

PR #813, opened 2019-05-26
#760: add parameter "target" to buildimage task

### LINKED ISSUE #760: Please add support for additional parameters to DockerBuildImage task
DockerBuildImage does not support the "--target" for multistage Docker builds.

### Expected Behavior

The DockerBuildImage task supports this parameter or it is possible to extend the DockerBuildImage task with additional parameters.

### Current Behavior

Currently it is not possible to add easily new parameters to this task.

### Context

We use Docker multistage builds to build different images from the same base image. Currently we have to use different Dockerfiles. The Docker cache is not usable in this case. As long we use the same files it is possible to use cached intermediate files.

m-raab:
Add a parameter "target" to build image task. A test for this new parameter was added to functional tests. 
The parameter "target" is used to build an image from a multistage Dockerfile with different stages.

closes #760

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -95,6 +95,16 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         noExceptionThrown()
     }
 
+    def "can build image with different targets"() {
+        buildFile << buildMultiStageImage()
+
+        when:
+        BuildResult result = build('buildTarget')
+
+        then:
+        result.output.contains("Successfully built")
+    }
+
     def "can build image using --cache-from with nothing in the cache"() {
         buildFile << buildImageUsingCacheFromWithNothingInCache()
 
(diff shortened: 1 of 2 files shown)


m-raab:
During build on travis one test failed. But this test was working on the local machine.
`com.bmuschko.gradle.docker.DockerSpringBootApplicationPluginFunctionalTest > Can create image and start container for Spring Boot application with default configuration [war plugin] FAILED`
What can I do to solve this issue?

bmuschko:
Looks like Bintray had an issue at the time this build was run: https://scans.gradle.com/s/6uyoh52xcdbd6/tests/dqwcqqtprpdqc-5upc2rraveyzo

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -95,6 +95,16 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
     }
 
+    def "can build image with different targets"() {
+        buildFile << buildMultiStageImage()
+
+        when:
+        BuildResult result = build('buildTarget')
+
+        then:
+        result.output.contains("Successfully built")
```
bmuschko: I was thinking more in terms of actually inspecting the built image. There's no point in verifying the console output here to see if the build was successful as the `build` method would throw an exception if the build failed. You will be able to find similar example if you look through some of the other functional tests.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -108,6 +108,15 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
     @Optional
     final Property<Long> shmSize = project.objects.property(Long)
 
+    /**
+     * With this parameter it is possible to build a special
+     * stage in a multi stage Docker file. This is available with Docker 17.05
+     * @since 4.9.0
```
bmuschko: This needs to be 4.10.0. The version 4.9.0 has already been released.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
m-raab: additional checks added

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
m-raab: changed version

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -117,7 +126,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
      * Output file containing the image ID of the built image. 
      * Defaults to "$buildDir/.docker/$taskpath-imageId.txt".
      * If path contains ':' it will be replaced by '_'.
-     * @since 4.9.0
+     * @since 4.10.0
```
bmuschko: You changed the wrong `@since` annotation. You can fix that?

m-raab: fixed.

Issue #814, opened 2019-05-29
Dockerfile is not part of the cache key when not defined


llaforest:
### Expected Behavior
When DockerBuildImage is called twice, it should be up-to-date. If the Dockerfile is modified, it should not be up-to-date anymore.

The dockerFile property should be set by default in the constructor so that it is part of the cache key.

### Current Behavior
When DockerBuildImage is called without specifying a dockerFile, it picks by default the one located in $inputDir/Dockerfile.
If Dockerfile is changed and the task called again, it will not run because up-to-date check failed to capture the change in dockerfile.

### Steps to Reproduce (for bugs)
- Create a docker file and save it in build/docker/Dockerfile
- Create a gradle task buildImage of type DockerBuildImage
- Run buildImage twice to get up-to-date status
- Modify manually the Dockerfile
- Run buildImage and the status will still be up-to-date

### Suggestion
By default the inputDir is already set to build/docker. Lets set the dockerFile to build/docker/Dockerfile by default as well. This way it will always be part of cache key.

PR #815, opened 2019-05-29
Set the dockerFile property in the constructor to update cache key

llaforest:
#814 
Set the dockerFile property in the constructor to update cache key

* Remove the @Optional annotation since always set in constructor
* Remove logic in the execution that checks for dockerFile since always set

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -53,7 +53,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
     /**
      * The Dockerfile to use to build the image.  If null, will use 'Dockerfile' in the
-     * build context, i.e. "$inputDir/Dockerfile".
+     * build context, i.e. "$inputDir/Dockerfile". It must be within inputDir.
      */
     @InputFile
     @PathSensitive(PathSensitivity.RELATIVE)


bmuschko:
@llaforest Please add all possible test cases you can think. It's OK to break the API if necessary if the result solved the problem properly. You initial changes already break one of the existing test cases.

llaforest:
I have fixed the issue caused by my change, that your test caught...
I will try to think of how and where to test this before we close this yes.

llaforest:
Ok all this uncovered another problem I did not want to address in this merge request:
Basically, if I set the dockerFile in the constructor, it is part of the cach key. But if later on the inputDir is changed, the default path to the docker file is not changed as well and it will fail. I cannot set the dockerFile in the execute because it has to be set in config time.
What I need to do is change the property dockerFile so that it is not an absolute path. IT has to be relative to the inputDir. So far the provider API automatically appends the $projectDir to it if I do not specify a full path.

The question is: How can we make the dockerFile property a path relative to the inputDir and not an absolute one. Anyway, it has to be relative because otherwise the java docker will not allow it.

llaforest:
Ok new development. We can close this PR and delete the branch.
The dockerFile has to be within the inputDir for java docker api to work. Since dockerFile is part of inputDir, it will always be part of cache key. If someone tries to use one outside the inputDir, the java docker api will fail anyway so it cannot work.
I still think it is not well documented so my only change would be the comment above the property...

cdancy:
@llaforest so ... closing this PR?

llaforest:
Yes we close it.

Issue #816, opened 2019-05-30
Manifest information not to invalidate cache key


llaforest:
### Expected Behavior
It is usually good practice to have some kind of manifest in artefacts. One or many of the following: Version, Build Date, Build Id, Commit Id, etc... This kind of information changes all the time but it should not trigger a rebuild all the time.

To have such a thing in a docker image, it could be saved either as a manifest file, or in image labels for example. The manifest file is tricky because it must be part of the staging dir and it will invalidate the cache key all the time. The labels as we know them will also invalidate the build unless another type of labels is created...

Let's say there is a new property called 'notInvalidatingLabels' in DockerBuildImage, these labels would be added to the buildImageCommand.withLabels in the execution but they would not be tagged @input. This way we append the information to the image but only when it rebuilds.

Would this be accepted?

### Current Behavior
So far nothing can be added to an image without invalidating cache key.

bmuschko:
I can think of multiple viable solutions here. I do not like the proposed solution in the PR as it will be very confusing to end users.

1. A user can disable up-to-date checking for any field that shouldn't be considered. Currently, I believe the `TaskInputs` or `TaskOutputs` API does not allow removing an existing input or output marker. That would be my preferable solution as it wouldn't require a code change to the plugin.
2. We remove the input annotation for that specific field. Any change to the values in the Map would not be considered for up-to-date checking.
3. Users with special needs will have to implement their own task. It's more work for the build author but the plugin stays multi-purpose.

llaforest:
I like what you propose it is much more generic. I just have no idea how we could enable or disable annotations on demand from the gradle task that extends DockerBuildImage. I will do some research but if you know how let me know I could give it a try.

hugares:
@bmuschko If I understand correctly, in option 2, you are suggesting to remove the @Input annotation on the existing `labels` field?

Would you be OK with that option as it would be easier/shorter than option 1 (which requires to implement a new feature in Gradle)?

bmuschko:
> I like what you propose it is much more generic. I just have no idea how we could enable or disable annotations on demand from the gradle task that extends DockerBuildImage. I will do some research but if you know how let me know I could give it a try.

[...]

PR #817, opened 2019-05-30
Add property buildLabels to add build info to image

llaforest:
#816 
Adding labels into this property will not invalidate the cache key.
It might be useful to add manifest information without trigerring
a new build every time.
But when a build is triggered, the information goes in.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -73,6 +73,46 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         gradleTaskDefinition << [imageCreationWithBuildArgs(), imageCreationWithLabelParameter()]
     }
 
+    def "can build image changing buildLabels without forcing to rebuild"() {
+        buildFile << imageCreationWithInspection()
+
+        when:
+        buildFile << """
+            buildImage {
+                buildLabels = ['buildLabel1':'test1', 'buildLabel2':'test2']
+            }
+        """
+        BuildResult result = build('inspectImage')
+
+        then:
+        result.output.contains("buildLabel1:test1, buildLabel2:test2")
+
+        when:
+        buildFile << """
+            buildImage {
+                buildLabels = ['buildLabel1':'test1', 'buildLabel2':'test2', 'buildLabel3':'test3']
+            }
+        """
+        result = build('inspectImage')
+
+        then:
+        result.output.contains("buildLabel1:test1, buildLabel2:test2")
+        !result.output.contains("buildLabel3:test3")
+
+        when:
+        buildFile << """
+            buildImage {
+                labels = ['label1':'test5', 'label2':'test6']
+            }
+        """
+        result = build('inspectImage')
+
+        then:
+        result.output.contains("buildLabel1:test1, buildLabel2:test2")
+        result.output.contains("buildLabel3:test3")
+        result.output.contains("label1:test5, label2:test6")
+    }
+
     def "can build image with multiple tags"() {
         buildFile << buildImageWithTags()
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -87,6 +87,9 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
     @Optional
     final MapProperty<String, String> labels = project.objects.mapProperty(String, String)
 
+    @Optional
```
hugares: Shouldn't this property be annotated with @Internal to indicate that it is neither an input nor an output.

llaforest: Yes I think it should be annotated with @Internal

llaforest:
Any comments on this feature?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -87,6 +87,14 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
     @Optional
     final MapProperty<String, String> labels = project.objects.mapProperty(String, String)
 
+    /**
+     * The buildLabels are labels that will not invalidate the cache key when changed.
+     * Usually used to store in the image build information that changes all the time.
+     * They will only be applied when the image builds, otherwise they are ignored.
+     */
+    @Internal
+    final MapProperty<String, String> buildLabels = project.objects.mapProperty(String, String)
```
cdancy: I'm not sure how else to address this but it _could_ become confusing as folks may not know they can actually use it given the annotation.

cdancy: Can you **also** add an @input and @optional along with keeping the @internal ?

llaforest: The goal of this feature is 'not to be part of the inputs'
Let say you need some references in your image about the date it was build, the git id of the commit you used, stuff like that will change all the time and will invalidate the cache key all the time if part of the input.
My claim is that we need a property that enables us to save some information in the image when it is rebuilt, but it must not trigger the rebuild.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -382,6 +422,27 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+            }
+
+            task buildImage(type: DockerBuildImage) {
+                dependsOn dockerfile
+            }
+
+            task inspectImage(type: DockerInspectImage) {
+                dependsOn buildImage
+                targetImageId buildImage.getImageId()
+            }
```
orzeh: Can you add a cleanup task, which will remove the image?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -192,7 +201,11 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         }
 
         if (labels.getOrNull()) {
-            buildImageCmd.withLabels(labels.get().collectEntries { [it.key, it.value.toString()] })
+            buildLabels.putAll(labels.get())
```
orzeh: So `labels` can silently override `buildLabels`? This may cause some nasty bugs I guess.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -142,6 +150,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             if(file.exists()) {
                 try {
                     getDockerClient().inspectImageCmd(file.text).exec()
+                    imageId.set(file.text)
```
orzeh: This change is unrelated to the scope of this PR.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: A task property can only be ignored for up-to-date checking by not annotating it or annotating it with `@Internal`. It doesn't make sense to combine `@Internal` with any other input or output annotation.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
llaforest: I could but non of the test cleanup so far. I was more thinking of implementing a cleanup for the whole test case.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
llaforest: I agree it could be moved to another issue

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
llaforest: No they will add to each other. They are complementary. If some labels are really part of the build image inputs, then they should be 'labels'. If they are just complementary informations, they should be 'buildLabels'

PR #819, opened 2019-06-04
Always set imageId within DockerBuildImage on success (#818)

### LINKED ISSUE #818: DockerBuildImage should not return null on up-to-date executions
### Expected Behavior
The contract to get the built imageId from the docker image is honoured if the DockerBuildImage task ran successfully by being able to query the "internal" getImageId() member in subsequent tasks. Such as was fixed within https://github.com/bmuschko/gradle-docker-plugin/issues/686
 
### Current Behavior
Seems like 4.9.0 broke this behaviour as I now get back null when the build task is considered up-to-date. Looks like https://github.com/bmuschko/gradle-docker-plugin/pull/807 might have introduced this breakage. 

From a very quick glance at the code within the outputs.upToDateWhen block it should read the image id from the file and set the imageId variable within it, since it's not the variable is never set.

### Context
We have a specific Dockerfile that we build within the gradle build to compile some C++ binaries for various linux versions. So the docker image itself will get compiled once, the first time but will be re-used/up-to-date on subsequent gradle executions as the docker image simply mounts the codebase the developers are editing within it.

Additional context:
Downgrading back to 4.8.1 version does not this problem, the DockerBuildImage task always runs and returns an imageId value.

```
task buildDevEnvImage(type: DockerBuildImage) {
    def uid = new com.sun.security.auth.module.UnixSystem().getUid()
    def gid = new com.sun.security.auth.module.UnixSystem().getGid()
    def username = System.properties['user.name']

    inputDir = file('buildenv/')
    tags.add("${project.name}-build:latest")
    buildArgs = ['UID': "${uid}",
                 'GID': "${gid}",
                 'USERNAME': username]
}

task createDevEnvContainer(type: DockerCreateContainer) {
    def uid = new com.sun.security.auth.module.UnixSystem().getUid()
    def gid = new com.sun.security.auth.module.UnixSystem().getGid()

    dependsOn buildDevEnvImage
    targetImageId buildDevEnvImage.getImageId()
    autoRemove = true
    binds = ["${project.rootDir}": "${project.rootDir}"]
    user = "${uid}:${gid}"
    cmd = ['cat', '-']
    tty = true
}
```

### Your Environment
Latest Centos 7 + Docker version 18.09.6, build 481bc77156

stevendpclark:
- Add a unit test that exposes the issue when the DockerBuildImage
   task is considered up to date an another task attempts to fetch
   the imageId value.
 - When the task is considered up-to-date, use the image id that was
   persisted to the file to set the value of imageId and return to
   other tasks to use it

closes #818

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -5,6 +5,7 @@ import com.bmuschko.gradle.docker.TestConfiguration
 import com.bmuschko.gradle.docker.TestPrecondition
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Issue
 import spock.lang.Requires
 import spock.lang.Unroll
 
@@ -264,6 +265,7 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         imageIdFile.text != ""
     }
 
+    @Issue('https://github.com/bmuschko/gradle-docker-plugin/issues/818')
     def "task not up-to-date when imageIdFile changed"() {
         given:
         buildFile << imageCreation()
@@ -316,6 +318,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 dependsOn dockerfile
                 shmSize = 128000L
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -332,6 +336,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
             task buildImage(type: DockerBuildImage) {
                 dependsOn dockerfile
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -370,6 +376,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 dependsOn buildImage
                 targetImageId buildImage.getImageId()
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -392,6 +400,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 dependsOn buildImage
                 targetImageId buildImage.getImageId()
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -415,6 +425,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 dependsOn dockerfile
                 tags.add('test/image:123')
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -437,6 +449,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 cacheFrom.add('$uniqueTag:latest')
                 tags.add('$uniqueTag:latest')
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -486,6 +500,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 inputDir = dockerfile.destFile.get().asFile.parentFile
                 ${useCacheFrom ? "cacheFrom.add('$uniqueTag:latest')" : ""}
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -502,6 +518,8 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
                 dependsOn dockerfile
                 network = 'host'
             }
+
+            ${imageIdValidation()}
         """
     }
 
@@ -533,6 +551,26 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
             }
 
             buildTarget.finalizedBy tasks.removeImage
+
+            ${imageIdValidation()}
+        """
+    }
+
+    private static String imageIdValidation() {
+        """
+            tasks.withType(DockerBuildImage) { Task task ->
+                def assertTask = tasks.create("assertImageIdFor\${task.name.capitalize()}"){
+                    def dependantTask = task
+                    doLast {
+                        def value = dependantTask.getImageId().getOrNull()
+                        if(value == null || !(value ==~ /^\\w+\$/)) {
+                            throw new GradleException("The imageId property was not set from task \${dependantTask.name}")
+                        }
+                    }
+                }
+
+                task.finalizedBy assertTask
+            }
         """
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -177,7 +177,9 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             File file = imageIdFile.get().asFile
             if(file.exists()) {
                 try {
-                    getDockerClient().inspectImageCmd(file.text).exec()
+                    def fileImageId = file.text
+                    getDockerClient().inspectImageCmd(fileImageId).exec()
+                    imageId.set(fileImageId)
                     return true
                 } catch (DockerException e) {
                     return false


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -270,6 +270,12 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         buildFile << """
             task verifyImageId {
                 dependsOn buildImage
+
+                doLast {
+                    // Expose issues with getImageId returning null
+                    // https://github.com/bmuschko/gradle-docker-plugin/issues/818
+                    buildImage.getImageId().get()
```
bmuschko: We'll need to assert that the value is not null. We should also have this check in the different test cases at every step along the way.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -150,7 +150,9 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
             File file = imageIdFile.get().asFile
             if(file.exists()) {
                 try {
-                    getDockerClient().inspectImageCmd(file.text).exec()
+                    def fileImageId = file.text
+                    getDockerClient().inspectImageCmd(fileImageId).exec()
+                    imageId.set(fileImageId)
```
bmuschko: The value of the `imageId` property should be set before making the call to Docker. In case Docker communication throws an exception, we already have the field populated.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -270,6 +270,12 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
         buildFile << """
             task verifyImageId {
                 dependsOn buildImage
+
+                doLast {
+                    // Expose issues with getImageId returning null
+                    // https://github.com/bmuschko/gradle-docker-plugin/issues/818
```
bmuschko: I propose you annotate the test method with `@Issue` and point to the relevant Github issue as value.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
stevendpclark: Interesting, okay I performed the set afterwards in case the inspectImageCmd had thrown due to the image id not being found, setting it to an invalid value would have seemed incorrect in that use case. 

I hadn't thought about the use case that the communication to docker itself had failed. 

Is it worth making the distinction? I.E. set if the failure is due to a communication issue and assuming it's still a valid image id and not setting it if the image doesn't exist. A quick look through the docker-java seems to mention a specific NotFoundException (would need to validate this)

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
stevendpclark: Ah neat, I didn't know about this annotation. Will do.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
llaforest: @bmuschko : I think we should set the imageIdt after the inspect fails because when it fails the task is set as not up-to-date and the imageId should not be set. It should be set only when the inspect succeeds in my opinion.
I will add the check in the tests though it is a good point.

@stevendpclark : At first I was catching only NotFoundException but then I realized that I could get more in other situations, miscommunication being one of them.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
stevendpclark: @bmuschko Would you have any additional feedback on how I implemented this validation across the workflow in the review feedback commit?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -533,6 +551,26 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
             
             buildTarget.finalizedBy tasks.removeImage
+
+            ${imageIdValidation()}
+        """
+    }
+
+    private static String imageIdValidation() {
+        """
+            tasks.withType(DockerBuildImage).all { Task task ->
```
bmuschko: The call to `all` is not required here. It will work without. Would you mind taking it off?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -533,6 +551,26 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+                def assertTask = tasks.create("assertImageIdFor\${task.name}"){
```
bmuschko: Would probably make sense to upper case the first letter of the task name so that the naming sticks to camel case notation.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -533,6 +551,26 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+                    def dependantTask = task
+                    doLast {
+                        def value = dependantTask.getImageId().getOrNull()
+                        if(value == null || !(value ==~ /^\\w+\$/)) {
```
bmuschko: Just as a side note - not required to change. It's probably good enough to use `assert` instead of having to throw a `GradleException`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
bmuschko: @llaforest I think you are right here. Sorry, let's change this back.

stevendpclark:
Review feedback was implemented, I rebased my branch and squashed the commits into a single one. Please let me know if there is any additional modifications you would like me to make.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
```diff
@@ -533,6 +551,26 @@ class DockerBuildImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
             buildTarget.finalizedBy tasks.removeImage
+
+            ${imageIdValidation()}
+        """
+    }
+
+    private static String imageIdValidation() {
+        """
+            tasks.withType(DockerBuildImage) { Task task ->
+                def assertTask = tasks.create("AssertImageIdFor\${task.name}"){
```
bmuschko: What I meant here was "upper case the value of the expression $task.name". Right now the task name starts with a upper case which is not standard convention. So this should look like `assertImageIdFor\${task.name.capitalize()}`. Can you please change that?

hugares:
Can this PR be included in v4.10.0 ?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
stevendpclark: Yeah that makes more sense. Okay will tweak the commit now. 
Thanks for all the time for the reviews!

stevendpclark:
Seems like the travis-ci build failed but for another functional test

> com.bmuschko.gradle.docker.tasks.container.DockerCreateContainerFunctionalTest > with autoRemove set, the container is removed after stopping FAILED
    org.spockframework.runtime.ConditionFailedWithExceptionError at DockerCreateContainerFunctionalTest.groovy:170
        Caused by: java.lang.NullPointerException at DockerCreateContainerFunctionalTest.groovy:170

Is this a known issue with a test running in the ci build system?

bmuschko:
@hugares Should be fine to include in 4.10.0.
@stevendpclark I have seen this test fail before. It's somewhat flaky. We'll need to investigate.

PR #820, opened 2019-06-04
imageId must be set when task is up-to-date

llaforest:
Related to #818


orzeh:
@llaforest this duplicates #819 
Let's continue the work under that PR.

PR #821, opened 2019-06-05
Remove final modifier on properties to allow overrides

llaforest:
Related to issue #816


bmuschko:
We'll just want to change that single field.

PR #823, opened 2019-06-07
Remove final from DockerBuildImage.labels property

### LINKED ISSUE #816: Manifest information not to invalidate cache key

hugares:
This allows build authors to remove the labels from the up-to-date
check by doing something like this:

```
import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage

class CustomDockerBuildImage extends DockerBuildImage {
    @Override
    @Internal
    MapProperty<String, String> getLabels() {
        super.getLabels()
    }
}
```

closes #816

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
@@ -94,10 +94,29 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
     /**
      * Labels to attach as metadata for to the image.
+     * <p>
+     * This property is not final to allow build authors to remove the labels from the up-to-date
+     * check by extending {@code DockerBuildImage} and annotating the overrided {@code getLabels()} method
+     * with {@code @Internal}, example:
+     *
+     * <pre>
+     * import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+     *
+     * class CustomDockerBuildImage extends DockerBuildImage {
+     *     @Override
+     *     @Internal
+     *     MapProperty<String, String> getLabels() {
+     *         super.getLabels()
+     *     }
+     * }
+     * </pre>
+     *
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -82,7 +82,7 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
 
     @Input
     @Optional
-    final MapProperty<String, String> labels = project.objects.mapProperty(String, String)
+    MapProperty<String, String> labels = project.objects.mapProperty(String, String)
```
bmuschko: Can you please add a comment via Javadoc that documents the design decision? While we are at it, please also explain `labels` represents as part of the comment.

hugares: Sure, Here is another PR to add documentation for all properties. Once it is merged, I will rebase the current MR on top of it and address you comment

bmuschko: Merged it yesterday.

hugares: Done

hugares:
Done, I added a functional test

PR #824, opened 2019-06-07
Document DockerBuildImages properties

hugares:


bmuschko:
Maybe I wasn't clear in https://github.com/bmuschko/gradle-docker-plugin/pull/823/files#r291587712 but what I meant was "document why we do not assign `final` and what the `labels` field does". I didn't necessarily propose to document all fields in this class though I doesn't hurt.

hugares:
> Maybe I wasn't clear in https://github.com/bmuschko/gradle-docker-plugin/pull/823/files#r291587712 but what I meant was "document why we do not assign final and what the labels field does". I didn't necessarily propose to document all fields in this class though it doesn't hurt.

I know you did not suggest that but instead of documenting only the label property(including reason why it is not final), I thought it would be good to document them all and in my other PR, I will add the section to explain why the label property is not final.

bmuschko:
@hugares Great, thanks! Much needed work.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -64,34 +64,58 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
     @Optional
     final SetProperty<String> tags = project.objects.setProperty(String)
 
+    /**
+     * When <code>true</code>, do not use docker cache when building the image.
```
bmuschko: The option would be to use `{@code}`. Would you mind using that instead?

hugares: Done

PR #825, opened 2019-06-10
Fix DockerBuildImageFunctionalTest failing when re-executing them

hugares:
When re-executing the tests for the seconds time, the test "task not
up-to-date when image not in registry" was failing with the following
error:
```
Task :removeImage FAILED
Removing image with ID 'd90aa9bfe0e6'.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':removeImage'.
> {"message":"conflict: unable to delete d90aa9bfe0e6 (cannot be forced) - image has dependent child images"}
```

The problem was that the tests are interfering with each other because
the images created were not unique. Make sure the image definition
created by imageCreation() method will be unique by adding a layer which
echo a UUID.


hugares:
I know I did unrelated formatting fix in this PR. I did them in a separate commit in the same PR instead of 2 PRs.

I was not sure how you guys prefer to work, 1 commit per PR or if you are OK with more that one as long as each is a logical unit. Let me know if you prefer that extract the commit removing the trailing spaces to its own PR.

bmuschko:
The smaller a PR, the easier and faster it is to review and merge. Thanks!

PR #833, opened 2019-06-21
Make Dockerfile.ItemJoinerUtil inner class static

hugares:
This inner class does not use any variables from the outer so it can be
made static. Eclipse report this as an error to have static methods in a
inner class that is not static.


cdancy:
Why not ;)

cdancy:
Thanks for the PR!!!

bmuschko:
IMHO the whole code around `ItemJoiner` has to be cleaned up. I looked at it multiple times and still don't understand why we have to have this complex logic there. Also `ItemJoiner` bleeds into the public API which is unnecessary from my perspective.

PR #835, opened 2019-06-21
ISSUE-834: Add multiple tags support application plugins

tanelso2:
Only being able to specify one tag for these projects when you can
specify multiple in the DockerBuildImage tasks is a real drag. It means
you would have to add a custom DockerBuildImage or DockerTagImage task
when the whole point of the plugin is to abstract all that out.

Tested by:
    `./gradlew clean build`

### DIFF
--- a/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
+++ b/src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
@@ -111,7 +111,7 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
         when:
         ExtensionAware dockerExtension = (ExtensionAware) project.extensions.getByType(DockerExtension)
         DockerJavaApplication dockerJavaApplicationExtension = dockerExtension.extensions.getByType(DockerJavaApplication)
-        dockerJavaApplicationExtension.tag.set("some-test-tag")
+        dockerJavaApplicationExtension.tags.set(["some-test-tag"])
 
         then:
         DockerBuildImage task = project
@@ -128,13 +128,30 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
         applyDockerJavaApplicationPluginAndApplicationPlugin(project)
         when:
         project.docker.javaApplication {
-            tag.set(testTagName)
+            tags.set([testTagName])
         }
         then:
         DockerBuildImage task = project.tasks.findByName(DockerJavaApplicationPlugin.BUILD_IMAGE_TASK_NAME)
         Set<String> tags = task.tags.get()
         tags.size() == 1
         tags.first() == testTagName
+    }
+
+    def "Can configure the dockerJava.javaApplication extension dynamically and set multiple tags"() {
+        given:
+        String testTagName1 = "some-test-tag"
+        String testTagName2 = "another-test-tag"
+        applyDockerJavaApplicationPluginAndApplicationPlugin(project)
+        when:
+        project.docker.javaApplication {
+            tags.set([testTagName1, testTagName2])
+        }
+        then:
+        DockerBuildImage task = project.tasks.findByName(DockerJavaApplicationPlugin.BUILD_IMAGE_TASK_NAME)
+        Set<String> tags = task.tags.get()
+        tags.size() == 2
+        tags.contains(testTagName1)
+        tags.contains(testTagName2)
     }
 
     private void applyDockerJavaApplicationPluginWithoutApplicationPlugin(Project project) {
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -64,11 +64,11 @@ class DockerJavaApplication {
     final ListProperty<Integer> ports
 
     /**
-     * The tag used for the Docker image.
+     * The tags used for the Docker image.
      * <p>
-     * Defaults to {@code <project.group>/<applicationName>:<project.version>}.
+     * Defaults to {@code [<project.group>/<applicationName>:<project.version>]}.
      */
-    final Property<String> tag
+    final ListProperty<String> tags
 
     /**
      * The JVM arguments used to start the Java program.
@@ -86,7 +86,7 @@ class DockerJavaApplication {
         maintainer.set(System.getProperty('user.name'))
         ports = objectFactory.listProperty(Integer)
         ports.set([8080])
-        tag = objectFactory.property(String)
+        tags = objectFactory.listProperty(String)
         jvmArgs = objectFactory.listProperty(String).empty()
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -15,12 +15,10 @@
  */
 package com.bmuschko.gradle.docker
 
-
 import groovy.transform.CompileStatic
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.ListProperty
 import org.gradle.api.provider.Property
-
 /**
  * The extension for configuring a Java application via the {@link DockerJavaApplicationPlugin}.
  * <p>
@@ -67,7 +65,7 @@ class DockerJavaApplication {
      * The tag used for the Docker image.
      * <p>
      * Defaults to {@code <project.group>/<applicationName>:<project.version>}.
-     * @deprecated use {@link #tags}
+     * @deprecated use {@link #tags} - will be removed in 6.0.0
      */
     @Deprecated
     final Property<String> tag
@@ -78,6 +76,16 @@ class DockerJavaApplication {
      * Defaults to {@code [<project.group>/<applicationName>:<project.version>]}.
      */
     final ListProperty<String> tags
+
+    /**
+     * Tags that will be used for building the Docker image, but will not be pushed to the repo.
+     * In order for these tags to be built, they should be specified in the {@link #tags} property.
+     * e.g. for rapid local development, "busybox:latest" would be useful, but becomes confusing when
+     * pushed to a remote repo
+     * <p>
+     * Defaults to {@code []}
+     */
+    final ListProperty<String> localOnlyTags
 
 
     /**
@@ -98,6 +106,7 @@ class DockerJavaApplication {
         ports.set([8080])
         tag = objectFactory.property(String)
         tags = objectFactory.listProperty(String)
+        localOnlyTags = objectFactory.listProperty(String).empty()
         jvmArgs = objectFactory.listProperty(String).empty()
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
@@ -64,11 +64,21 @@ class DockerJavaApplication {
     final ListProperty<Integer> ports
 
     /**
-     * The tags used for the Docker image.
+     * The tag used for the Docker image.
+     * <p>
+     * Defaults to {@code <project.group>/<applicationName>:<project.version>}.
+     * @deprecated use {@link #tags}
+     */
+    @Deprecated
+    final Property<String> tag
+
+    /**
+     * The tag used for the Docker image.
      * <p>
      * Defaults to {@code [<project.group>/<applicationName>:<project.version>]}.
      */
     final ListProperty<String> tags
+
 
     /**
      * The JVM arguments used to start the Java program.
@@ -86,6 +96,7 @@ class DockerJavaApplication {
         maintainer.set(System.getProperty('user.name'))
         ports = objectFactory.listProperty(Integer)
         ports.set([8080])
+        tag = objectFactory.property(String)
         tags = objectFactory.listProperty(String)
         jvmArgs = objectFactory.listProperty(String).empty()
     }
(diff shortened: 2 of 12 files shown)


src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -64,11 +64,11 @@ class DockerJavaApplication {
...
 
     /**
-     * The tag used for the Docker image.
+     * The tags used for the Docker image.
      * <p>
-     * Defaults to {@code <project.group>/<applicationName>:<project.version>}.
+     * Defaults to {@code [<project.group>/<applicationName>:<project.version>]}.
      */
-    final Property<String> tag
+    final ListProperty<String> tags
```
orzeh: We should document here that **only** the first tag in the list will be pushed.

cdancy:
While this is aiming at `5.0.0` I feel torn on ripping the old `tag` option out. For massive code-bases like what we have (think the linux kernel times 3 and that's not an exaggeration) migrating to newer versions of plugins is something we actively want to do but often can't due to breaking changes and so things end up moving very slow. I don't know how others feel but I'd like to keep both options if at all possible, deprecate `tag`, and put in a note that we will remove it in `6.0.0` or something like that. That would give us, and others in the same boat, the ability to upgrade to the latest and greatest version of this plugin, on top of the time needed to migrate things over before a next major release. 

@bmuschko thoughts?

tanelso2:
I agree that it would be good to keep backwards compatibility, however I'm not too familiar with Gradle plugin development so I would need some guidance on how to do that correctly.

Definitely willing to do it, just might need some guidance

orzeh:
@tanelso2 please start with looking at #380

tanelso2:
> To preserve backward compatibility if tag is defined, tags are ignored.

Oh that makes a lot of sense. Easy enough

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
cdancy: Should we iterate over all potential tags and attempt to push? Maybe have `DockerPushImage` also take in a list of tags and we attempt to push each?

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
```diff
@@ -55,9 +55,18 @@ class DockerSpringBootApplication {
      * The tag used for the Docker image.
      * <p>
      * Defaults to {@code <project.group>/<applicationName>:<project.version>}.
+     * @deprecated use {@link #tags}
      */
+    @Deprecated
```
cdancy: Lets put in a comment that `tag` will be removed with version `6.0.0`.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
bmuschko: @cdancy I do like that idea in case we want to move forward with it.

tanelso2:
Hey everyone, sorry I've been away from this PR for a while. 

What is the current status? Is this change fine or is there more that I need to add to it?

bmuschko:
@cdancy @orzeh Can you get guide this PR to the finish line? I already added my comments above.

cdancy:
@tanelso2 it looks like you haven't addressed the comments in questions yet. Please correct me if I'm wrong. I'd go back over the _diff_ and address comments and/or provide opinions. Would be very cool to have this functionality. Reach out if you need any help.

cdancy:
> We just made an opinionated gradle plugin for our build system at work and I'm absolutely swamped evangelizing it to other teams.

@tanelso2 no rush and trust me: I know the feeling all too well ;)

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -31,8 +32,15 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
      * The image name e.g. "bmuschko/busybox" or just "busybox" if you want to default.
      */
     @Input
+    @Optional
     final Property<String> imageName = project.objects.property(String)
```
cdancy: Lets add an @deprecated here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -45,51 +53,64 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
...
+            PushImageCmd pushImageCmd = dockerClient.pushImageCmd(imageName)
 
-        if(tag.getOrNull()) {
-            pushImageCmd.withTag(tag.get())
-            logger.quiet "Pushing image with name '${imageName.get()}:${tag.get()}'."
-        } else {
-            logger.quiet "Pushing image with name '${imageName.get()}'."
-        }
+            if (tag.getOrNull()) {
+                pushImageCmd.withTag(tag.get())
```
cdancy: Lets make the tag parameter @deprecated as well.

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
```diff
@@ -55,9 +55,28 @@ class DockerSpringBootApplication {
...
+
+    /**
+     * Tags that will be used for building the Docker image, but will not be pushed to the repo.
+     * In order for these tags to be built, they should be specified in the {@link #tags} property.
+     * e.g. for rapid local development, "busybox:latest" would be useful, but becomes confusing when
+     * pushed to a remote repo
+     * <p>
+     * Defaults to {@code []}
+     */
+    final ListProperty<String> localOnlyTags
```
cdancy: Do we really need the `localOnlyTags`? Just trying to think what the usecase is here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -31,8 +32,15 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
...
      */
     @Input
+    @Optional
     final Property<String> imageName = project.objects.property(String)
 
+    /**
+     * The image names to push e.g. ["bmuschko/busybox"]
+     */
+    @Input
+    final ListProperty<String> imageNames = project.objects.listProperty(String)
```
cdancy: hmmm ... guess I was thinking we would be keeping things more in line with what we did elsewhere in the PR and make the `tag` parameter plural and create a `tags` parameter and deprecate the former.

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
tanelso2: So I was thinking of my team's specific usecase, where we build "<application>:latest" and reference it in all of our kubernetes manifests so that its fairly easy to do local development.

However, we don't want those "latest" tags to ever go to our docker repo, we just want them to stay local. This was just the first way I thought of doing it, I'm open to other suggestions.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
tanelso2: Yeah that makes sense. I can look into doing it that way. 

I originally did it like this because the plugin was autowiring the imageName based on the results of the DockerBuildImage task

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
cdancy: Ok I can see that being useful and handy for others. Can we change the name though to something more descriptive? Maybe `skipPushTags` or `skipPushOnTags` or if there is something better.

tanelso2: Those names are definitely better than `localOnlyTags`.

How about `noPushTags`?

cdancy: Hmmm ... I'm on the line for `noPushTags` and `skipPushTags`. I leave the ball in your court ;)

src/integTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginIntegrationTest.groovy
```diff
@@ -137,6 +137,23 @@ class DockerJavaApplicationPluginIntegrationTest extends AbstractIntegrationTest
...
+        applyDockerJavaApplicationPluginAndApplicationPlugin(project)
+        when:
+        project.docker.javaApplication {
+            tags.set([testTagName1, testTagName2])
+        }
+        then:
+        DockerBuildImage task = project.tasks.findByName(DockerJavaApplicationPlugin.BUILD_IMAGE_TASK_NAME)
+        Set<String> tags = task.tags.get()
+        tags.size() == 2
+        tags.contains(testTagName1)
```
bmuschko: You can probably just say `tags == [testTagName1, testTagName2]` here to make it simpler.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
```diff
@@ -67,9 +65,29 @@ class DockerJavaApplication {
...
+
+    /**
+     * Tags that will be used for building the Docker image, but will not be pushed to the repo.
+     * In order for these tags to be built, they should be specified in the {@link #tags} property.
+     * e.g. for rapid local development, "busybox:latest" would be useful, but becomes confusing when
+     * pushed to a remote repo
+     * <p>
+     * Defaults to {@code []}
+     */
+    final ListProperty<String> localOnlyTags
```
bmuschko: I don't think we should introduce this functionality. It makes understanding the API harder and users will likely get confused as they don't look the API. I can see your use case but why not go with a concrete tag (e.g. semantic versioning) and use that for local testing? I am just not a big fan of `latest` in Docker as it leads people to believe that it actually is the latest in terms of a SNAPSHOT.

src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplication.groovy
```diff
@@ -55,9 +55,28 @@ class DockerSpringBootApplication {
...
+     * @deprecated use {@link #tags} - will be removed in 6.0.0
      */
+    @Deprecated
     final Property<String> tag
 
+    /**
+     * The tags used for the Docker image.
+     * <p>
+     * Defaults to {@code [<project.group>/<applicationName>:<project.version>]}.
+     */
```
bmuschko: Should also have the `@since` annotation.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -31,8 +32,15 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
      * The image name e.g. "bmuschko/busybox" or just "busybox" if you want to default.
      */
     @Input
+    @Optional
     final Property<String> imageName = project.objects.property(String)
 
+    /**
+     * The image names to push e.g. ["bmuschko/busybox"]
+     */
```
bmuschko: Missing `@since`.

src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplication.groovy
cdancy: @bmuschko checkout the convo we had elsewhere. It wouldn't necessarily be for latest (once the name is changed) but for any tag we don't want pushed remotely. For example I create 10 tags but only want to push 5 of them for any number of reasons.

bmuschko: @cdancy Really trying to understand the "reasons" and identify if there are other ways to make this happen without having to change the code as proposed. Can someone list the concrete use cases?

cdancy: May be naive at this point but I took this along the same lines as the change for `DockerSpringBootApplication` which is discussed further above.

bmuschko: I guess you are referring to this?

> So I was thinking of my team's specific usecase, where we build ":latest" and reference it in all of our kubernetes manifests so that its fairly easy to do local development.

It doesn't really describe the use case. It only describes the "what" and not the "why". It will need more context for me to understand it. I am strongly voting for making the change for allowing building/pushing multiple tags first. After that we can have another discussion about which of those tags should be pushed and why. The changes in this PR try to implement multiple use cases and gets far too confusing to discuss.

Here's what I am trying to really figure out:

- What is the actual use case for not pushing all tags that were declared?
- Can this be solved in a different way without making these changes?
- Is this a common use case that would justify making the code more complex and harder to understand for other users? Maybe this use case is very specific to this one team and it might be better to let them build their own plugin.

tanelso2: > What is the actual use case for not pushing all tags that were declared?

Balancing strict governance policies for Docker repos with developer productivity. The `:latest` tag is useful for doing local development with Helm and Kubernetes, so that we can test our manifests locally without having to edit the version every time. However, having a `:latest` tag is not allowed in our repo (everything must be versioned), and if we tried to push all of the tags, that one would be rejected and the whole task would fail.

bmuschko:
@tanelso2 After a couple of refactorings to unify the logic of the convention plugins, I opened a new PR to support multiple tags. Please have a look at it and let us know what you think. Closing this one. Thanks for your initial work!

PR #836, opened 2019-06-23
Removed duplicated code and extracted use of ItemJoiner from public API

bmuschko:
* There's no point in having different treatment for single vs. multiple items. Both notations are supported (see https://docs.docker.com/engine/reference/builder/#env).
* `SingleItemJoiner` duplicated most of the code from `MultiItemJoiner`
* `ItemJoiner` really is an implementation detail and shouldn't bleed into the public API (see constructor of `MapInstruction`)

Next question from my end: Why do we even have such complex escape logic? Couldn't the end user provide the proper syntax and escape on their end if needed? That's what you have to do in Java in general. We can address the question after reviewing/merging this PR.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerfileFunctionalTest.groovy
@@ -165,7 +165,7 @@ LABEL maintainer=benjamin.muschko@gmail.com
 RUN echo deb http://archive.ubuntu.com/ubuntu precise universe >> /etc/apt/sources.list
 CMD ["echo", "some", "command"]
 EXPOSE 8080 14500
-ENV ENV_VAR_KEY envVarVal
+ENV ENV_VAR_KEY=envVarVal
 ENV ENV_VAR_A=val_a
 ENV ENV_VAR_B=val_b ENV_VAR_C=val_c
 ADD http://mirrors.jenkins-ci.org/war/1.563/jenkins.war /opt/jenkins.war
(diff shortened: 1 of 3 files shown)


cdancy:
Code LGTM but it's making me scratch my head as to why it was needed and/or asked for and/or helpful in the first place. Can we do a git blame on the lines removed, see which PR/author put this in, and further discuss from there? While this may pass testing I would hate to leave one of our end-users out in the cold if this breaks an existing use-case.

bmuschko:
See https://github.com/bmuschko/gradle-docker-plugin/issues/415 but I still understand why this change was needed.

cdancy:
>  Couldn't the end user provide the proper syntax and escape on their end if needed?

So that looks to be the crutch of it: making things slightly easier for the end-user. The multiple ways the user can pass along env-vars for the Dockerfile itself is what this was done for and to try and minimize the pain around it which I can appreciate.

IDK. I'm on the fence as to removing the escaping code as it can be helpful and might force users to now have to amend things on their end.

bmuschko:
> IDK. I'm on the fence as to removing the escaping code as it can be helpful and might force users to now have to amend things on their end.

This needs to be documented if we want to keep the code. There's no way for a user to know how this behaves under the hood. Most users likely just do the escaping themselves as there's no indication that the task will do some special handling here.

bmuschko:
@cdancy So from what I understand we are good with merging this PR and leave the rest of the code in place for now?

bmuschko:
I am assuming this means "yes". Merging!

cdancy:
I'm indifferent. I can see the yes and the no reasons. If you want it OK if not then OK as well ;)

PR #867, opened 2019-10-06
Multi-tag support for push operation and convention plugins

bmuschko:
This change takes a step-wise approach for the work started in https://github.com/bmuschko/gradle-docker-plugin/pull/835.

This is a breaking change and will be delivered with the next major version. The property for defining a single tag has been removed without a deprecation cycle. Unify other custom tasks to simplify and unify API.

- In its current implementation there's a one to one map from built tags to pushed tags. A later PR might introduce logic to make the push operation more selective about which tags to push.
- The API of the push task didn't align with the build task. Separating the repository from the tag led to misunderstandings from end users. Unifying both APIs makes a much easier to hand over the tags property values. The same goes for other tasks.
- Removed the method `DockerPullImage.getImageId()`. The returned value didn't really represent the image ID and therefore wasn't accurate. In fact you'd have to represent a list of image IDs if done right. You can simply ask for the Provider value of the `tag` property if required.

### DIFF
--- a/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
+++ b/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
@@ -12,7 +12,7 @@ docker {
         baseImage = 'dockerfile/java:openjdk-7-jre'
         maintainer = 'Benjamin Muschko "benjamin.muschko@gmail.com"'
         ports = [9090, 5701]
-        tag = 'jettyapp:1.115'
+        tags = ['jettyapp:1.115']
         jvmArgs = ['-Xms256m', '-Xmx2048m']
     }
 }
--- a/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
+++ b/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
@@ -12,7 +12,7 @@ docker {
         baseImage = 'dockerfile/java:openjdk-7-jre'
         maintainer = 'Benjamin Muschko "benjamin.muschko@gmail.com"'
         ports = [9090, 5701]
-        tags = ['jettyapp:1.115']
+        images = ['jettyapp:1.115']
         jvmArgs = ['-Xms256m', '-Xmx2048m']
     }
 }
--- a/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
+++ b/src/docs/samples/code/application-plugin/basic/groovy/build.gradle
@@ -12,7 +12,7 @@ docker {
         baseImage = 'dockerfile/java:openjdk-7-jre'
         maintainer = 'Benjamin Muschko "benjamin.muschko@gmail.com"'
         ports = [9090, 5701]
-        images = ['jettyapp:1.115']
+        images = ['jettyapp:1.115', 'jettyapp:latest']
         jvmArgs = ['-Xms256m', '-Xmx2048m']
     }
 }
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
@@ -23,18 +23,11 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 
 class DockerCommitImage extends DockerExistingContainer {
-    /**
-     * Repository.
-     */
-    @Input
-    @Optional
-    final Property<String> repository = project.objects.property(String)
 
     /**
-     * Commit tag.
+     * The tag used to save the image e.g. {@code my-java-app:1.2.3}.
      */
     @Input
-    @Optional
     final Property<String> tag = project.objects.property(String)
 
     /**
@@ -70,14 +63,7 @@ class DockerCommitImage extends DockerExistingContainer {
     void runRemoteCommand() {
         logger.quiet "Committing image for container '${getContainerId().get()}'."
         CommitCmd commitCmd = dockerClient.commitCmd(getContainerId().get())
-
-        if(repository.getOrNull()) {
-            commitCmd.withRepository(repository.get())
-        }
-
-        if(tag.getOrNull()) {
-            commitCmd.withTag(tag.get())
-        }
+        commitCmd.withTag(tag.get())
 
         if(message.getOrNull()) {
             commitCmd.withMessage(message.get())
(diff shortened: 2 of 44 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -47,49 +41,45 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
...
+            PushImageCmd pushImageCmd = dockerClient.pushImageCmd(currentTag)
 
-        if(tag.getOrNull()) {
-            pushImageCmd.withTag(tag.get())
-            logger.quiet "Pushing image with name '${imageName.get()}:${tag.get()}'."
-        } else {
-            logger.quiet "Pushing image with name '${imageName.get()}'."
-        }
+            if(registryCredentials) {
+                AuthConfig authConfig = new AuthConfig()
```
orzeh: `AuthConfig` can be created only once outside `each` loop?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
```diff
@@ -47,49 +41,45 @@ class DockerPushImage extends AbstractDockerRemoteApiTask implements RegistryCre
 
     @Override
     void runRemoteCommand() {
-        PushImageCmd pushImageCmd = dockerClient.pushImageCmd(imageName.get())
+        tags.get().each { currentTag ->
+            logger.quiet "Pushing image with name '${currentTag}'."
```
orzeh: Maybe we should change _name_ to _tag_.

bmuschko: Good idea. Will change that. Technically, `tag` is the wrong term here as well as we are talking about `<repository>/<image-name>:<tag>`. I don't think there's really a blanket term in Docker that would cover all of this.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
bmuschko: Going to look into it.

bmuschko:
> My only concern with this feature is that we break an assumption that each task is an atomic interaction with Docker API. If we have 3 tags and pushing the second one fails, the build ends up in inconsistent state, meaning you don't know which tags were pushed successfully. I guess we have to provide some hook for users to compensate in case of an error. But probably we can add this later if someone asks for this.

Docker Java doesn't offer an API for providing multiple tags at the moment. I am not sure how we'd make this atomic. Let's say the first push worked but the second push failed, I am not sure how we'd properly handle the situation. I am not aware of a call that can delete a remote Docker image so there's no real rollback or atomic commit functionality.

I am not sure how a user-facing callback could help. There's nothing a user can do to delete a remote image either. Maaaaybe write out a custom error message but I am hoping that Docker Java provides a useful message here.

Let's me know how you think this could work as atomic operation and we can discuss options.

orzeh:
> Docker Java doesn't offer an API for providing multiple tags at the moment. I am not sure how we'd make this atomic. Let's say the first push worked but the second push failed, I am not sure how we'd properly handle the situation. I am not aware of a call that can delete a remote Docker image so there's no real rollback or atomic commit functionality.

Probably that's the reason why this option is not built into Docker itself :)

I guess we don't have to worry about it that much, but maybe we should just add one sentence to the User Guide so that people are aware.

src/docs/samples/code/application-plugin/basic/groovy/build.gradle
```diff
@@ -12,7 +12,7 @@ docker {
         baseImage = 'dockerfile/java:openjdk-7-jre'
         maintainer = 'Benjamin Muschko "benjamin.muschko@gmail.com"'
         ports = [9090, 5701]
-        tag = 'jettyapp:1.115'
+        tags = ['jettyapp:1.115']
```
cdancy: Maybe provide multiple tags here so that users will know how things work? Possibly the same comment for other doc-related changes.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
```diff
@@ -23,18 +23,11 @@ import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 
 class DockerCommitImage extends DockerExistingContainer {
-    /**
-     * Repository.
-     */
-    @Input
-    @Optional
-    final Property<String> repository = project.objects.property(String)
```
cdancy: If we're removing this property a "better" name might be to just use `image` as that technically encompasses the entire URL. It's at least what is used within the docker docs all over the place.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
cdancy: I would argue just to say `Pulling image ${currentImage}` (goes inline with comments I made above).

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
bmuschko: done

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.groovy
bmuschko: done

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.groovy
bmuschko: done

src/docs/samples/code/application-plugin/basic/groovy/build.gradle
bmuschko: done

cdancy:
LGTM and I like the new functionality. Have at least one place we'll be able to take advantage of this internally for which we have custom code to do this (though admittedly not as clean :) )

PR #869, opened 2019-10-07
Gradle 6.0 compatibility fixes

ljacomet:
Resolve issues related to Gradle 6.0 compatibility

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
@@ -97,6 +97,11 @@ class DockerLogsContainer extends DockerExistingContainer {
     @Optional
     Writer sink
 
+    // Allows subclasses to carry their own logic
+    protected Date getInternalSince() {
+        return since.getOrNull()
+    }
+
     DockerLogsContainer() {
         stdOut.set(true)
         stdErr.set(true)
@@ -200,8 +205,9 @@ class DockerLogsContainer extends DockerExistingContainer {
             logsCommand.withTail(tailCount.get())
         }
 
-        if (since.getOrNull()) {
-            logsCommand.withSince((int) (since.get().time / 1000))
+        Date since = getInternalSince()
+        if (since) {
+            logsCommand.withSince((int) (since.time / 1000))
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
```diff
@@ -97,6 +97,11 @@ class DockerLogsContainer extends DockerExistingContainer {
     @Optional
     Writer sink
 
+    // Allows subclasses to carry their own logic
+    protected Date getInternalSince() {
+        return since.getOrNull();
```
bmuschko: We can clean it up but let's remove the comma at the end.

src/main/groovy/com/bmuschko/gradle/docker/utils/IOUtils.groovy
```diff
@@ -44,6 +44,5 @@ final class IOUtils {
         ProgressLoggerFactory factory = registry.get(ProgressLoggerFactory)
         ProgressLogger progressLogger = factory.newOperation(Objects.requireNonNull(clazz))
         progressLogger.setDescription("ProgressLogger for ${clazz.getSimpleName()}")
-        progressLogger.setLoggingHeader(null)
```
bmuschko: Will this change make it completely Gradle 6.0-compatible? The creation of a ProgressLogger is still OK?

src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
ljacomet: Ah, sorry, too much Java muscle memory 😉

src/main/groovy/com/bmuschko/gradle/docker/utils/IOUtils.groovy
ljacomet: Yes, only some methods have been removed. See [this commit](https://github.com/gradle/gradle/commit/4ca2ff6c4e8f1bb406dcc9a051899256a2e3586e).

cdancy: So technically it's still there just the setter is removed. In theory we should be able to do:
```
progressLogger.loggingHeader = null
```
Though that seems a redundant operation. Just putting it out there should folks stumble upon this in the future.

bmuschko:
Thanks for sending in the pull request. It has been merged.

Issue #861, opened 2019-09-25
Add support of ipcMode to DockerCreateContainer


jberglund-BSFT:
Would like to see support of ipcMode added
This is an existing setting in the docker-java HostConfig:
https://github.com/docker-java/docker-java/blob/master/docker-java/src/main/java/com/github/dockerjava/api/model/HostConfig.java#L1140

Docker run reference:
https://docs.docker.com/engine/reference/run/#ipc-settings---ipc

This would be a minor change, something like https://github.com/jberglund-BSFT/gradle-docker-plugin/commit/0d0b39d2642be71a0e364d9860ed3b0d833ea6e0

### Expected Behavior
Able to set the ipcMode

### Current Behavior
Not able to set the ipcMode

### Context
Need to be able to set the ipcMode to share memory segments for an in-memory database running in a sidecar.

### Steps to Reproduce (for bugs)
(not a bug)

### Your Environment
(not a bug)

jberglund-BSFT:
also interested in adding sysctls

cdancy:
@jberglund-BSFT care to send in a PR? If so I would break things up into multiple PR's for each feature requested to making reviewing easier.

jberglund-BSFT:
@cdancy would love to, was intending that the PR template asked for an issue first. PR coming shortly.

cdancy:
Sounds good. Looking forward to it.

bmuschko:
@jberglund-BSFT Can you please move the relevant properties into an inner class called `HostConfig` similar to `LogConfig` and `ExposedPort`? A flat list of properties simply isn't maintainable.

jberglund-BSFT:
> @jberglund-BSFT Can you please move the relevant properties into an inner class called `HostConfig` similar to `LogConfig` and `ExposedPort`? A flat list of properties simply isn't maintainable.
Do you want all the existing hostConfig properties moved there too, or just new ones?

bmuschko:
> Do you want all the existing hostConfig properties moved there too, or just new ones?

Yes, absolutely. If there are other ones that actually belong to `HostConfig` then they should move as well.

jberglund-BSFT:
@bmuschko Do you want the DockerCreateContainer task configuration to contain a `hostConfig` input  and move the properties there, breaking any backwards compatibility?

bmuschko:
@jberglund-BSFT Yes, please. We'll bump up the version appropriately. Let's not go through a deprecation cycle just for this.

[...]

PR #873, opened 2019-10-15
Extracts host configuration for better maintainability

bmuschko:
For more context, see https://github.com/bmuschko/gradle-docker-plugin/issues/861.

### DIFF
--- a/src/docs/samples/code/remote-api-plugin/functional-test/groovy/build.gradle
+++ b/src/docs/samples/code/remote-api-plugin/functional-test/groovy/build.gradle
@@ -13,8 +13,8 @@ task buildMyAppImage(type: DockerBuildImage) {
 task createMyAppContainer(type: DockerCreateContainer) {
     dependsOn buildMyAppImage
     targetImageId buildMyAppImage.getImageId()
-    portBindings = ['8080:8080']
-    autoRemove = true
+    hostConfig.portBindings = ['8080:8080']
+    hostConfig.autoRemove = true
 }
 
 task startMyAppContainer(type: DockerStartContainer) {
--- a/src/docs/samples/code/remote-api-plugin/functional-test/kotlin/build.gradle.kts
+++ b/src/docs/samples/code/remote-api-plugin/functional-test/kotlin/build.gradle.kts
@@ -13,8 +13,8 @@ val buildMyAppImage by tasks.creating(DockerBuildImage::class) {
 val createMyAppContainer by tasks.creating(DockerCreateContainer::class) {
     dependsOn(buildMyAppImage)
     targetImageId(buildMyAppImage.getImageId())
-    portBindings.set(listOf("8080:8080"))
-    autoRemove.set(true)
+    hostConfig.portBindings.set(listOf("8080:8080"))
+    hostConfig.autoRemove.set(true)
 }
 
 val startMyAppContainer by tasks.creating(DockerStartContainer::class) {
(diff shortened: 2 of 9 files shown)


cdancy:
Hmmm ... IDK. I'm on the fence. At least in this PR it seems to be a lot of change for getting the exact same thing whilst also putting in yet another backwards breaking change.

bmuschko:
We are targeting a major version change here so this is our chance to make things more maintainable. Yes, it means breaking the API as-is but with other benefits attached to it. IMHO this has not been addressed in the past and we need to pay a higher price now as we added more and more properties over time.

Generally speaking, it's not too hard to explain to users why the change needed to happen. In consuming build scripts, the change is straightforward: use the `hostConfig` property.

BTW: This change has already been merged so if you don't want to revert the change, we are already in the game. ;-)

cdancy:
Oh you merged this into master ... that's a bummer as the next release has to be a major version bump. Which also means no minor/patch versions can go out. Was hoping to get the "netty config" stuff merged in as it's not anything near breaking but even if I do that we'd still have to bump to next major version and our pipeline is massive and would require lots of work to account for this change. That's a huge bummer. Will have to think on this...

orzeh:
I guess we can create a HF branch and release minor/patch versions from there?

cdancy:
@orzeh I suppose that could work. Let me worry about the "netty config" stuff first and get a PR out before I stress out over this :)

bmuschko:
> Oh you merged this into master ... that's a bummer as the next release has to be a major version bump. Which also means no minor/patch versions can go out. Was hoping to get the "netty config" stuff merged in as it's not anything near breaking but even if I do that we'd still have to bump to next major version and our pipeline is massive and would require lots of work to account for this change. That's a huge bummer. Will have to think on this...

I am planning to release the next major version within the next week or two. So it's not something you'll have to wait on for long. Yes, there are some API changes but depending on what part of the plugin you are using, it won't be as much work.

cdancy:
I wish I could explain the size and scope of our build pipeline. It's the biggest most massive thing I've ever seen. LOADS of stuff being built across teams across the globe which use this plugin to varying degrees.

Could a patch release work like what Orzeh suggested above from prior to this merge?

jberglund-BSFT:
@bmuschko @cdancy  I'm not a groovy expert, but also had misgivings about the api breaking change.. would it be possible in groovy to flatten the hostConfig props ontop of the CreateContainerCmd, prefer the `hostConfig.nnn` props, and add a warning that the flattened ones will be removed in the next release?

bmuschko:
> Could a patch release work like what Orzeh suggested above from prior to this merge?

Sure, that could work but you'd probably have to run the release steps manually and merge back into `master` afterward. But before you do so, please identify the actual issue first.

jberglund-BSFT:
@bmuschko I noticed in the docker-java project, HostConfig is also used in the inspect container response, I wonder if it shouldn't eventually be extracted outside of the DockerCreateContainerCmd to better match the lower layer in order to re-use

bmuschko:
@jberglund-BSFT Yes, it should be done across the board. But I'd need some help from my team mates. ;-) This isn't my full time job.

cdancy:
Ours either! That's why big breaking changes like this are scary!
> I'm not a groovy expert, but also had misgivings about the api breaking change.. would it be possible in groovy to flatten the hostConfig props ontop of the CreateContainerCmd, prefer the hostConfig.nnn props, and add a warning that the flattened ones will be removed in the next release?

I'm honestly not sure but sounds like a reasonable idea.

PR #877, opened 2019-10-17
ISSUE-876: Expose env-var to optionally use netty-exec-cmd-factory.

### LINKED ISSUE #876: Expose "netty config" on docker-client
So we've started running into [THIS](https://github.com/docker-java/docker-java/issues/698) issue here in our build pipeline. A user chimed in on that thread with a supposed solution:
```
DockerCmdExecFactory factory= new NettyDockerCmdExecFactory().withConnectTimeout(100000);
DockerClient docker = DockerClientBuilder.getInstance(<some_config>).withDockerCmdExecFactory(factory).build();
```
Was thinking we could wire up the config to expose letting folks programatically do this. Something like:
```
docker {
    url = 'https://192.168.59.103:2376'
    certPath = new File(System.properties['user.home'], '.boot2docker/certs/boot2docker-vm')
    netty {
        connectTimeout = 100000
    }
}
```
Basically if the `netty` closure is present then use it otherwise it's business as usual.

Thoughts/Opinions?

cdancy:
Resolves #876

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -18,8 +18,10 @@ package com.bmuschko.gradle.docker.tasks
 import com.bmuschko.gradle.docker.DockerExtension
 import com.bmuschko.gradle.docker.DockerRemoteApiPlugin
 import com.github.dockerjava.api.DockerClient
+import com.github.dockerjava.api.command.DockerCmdExecFactory
 import com.github.dockerjava.core.DefaultDockerClientConfig
 import com.github.dockerjava.core.DockerClientBuilder
+import com.github.dockerjava.netty.NettyDockerCmdExecFactory
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
 import org.gradle.api.Action
@@ -36,6 +38,8 @@ import org.gradle.api.tasks.TaskAction
 
 @CompileStatic
 abstract class AbstractDockerRemoteApiTask extends DefaultTask {
+
+    private static final String ENV_VAR_GDP_CONN_NETTY = "GDP_CONN_NETTY"
 
     /**
      * Docker remote API server URL. Defaults to "http://localhost:2375".
@@ -134,6 +138,7 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     @Internal
     @Memoized
     DockerClient getDockerClient() {
+
         DockerClientConfiguration dockerClientConfiguration = createDockerClientConfig()
         DockerExtension dockerExtension = (DockerExtension) project.extensions.getByName(DockerRemoteApiPlugin.EXTENSION_NAME)
         String dockerUrl = getDockerHostUrl(dockerClientConfiguration, dockerExtension)
@@ -157,8 +162,9 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
 
         DefaultDockerClientConfig dockerClientConfig = dockerClientConfigBuilder.build()
 
-        // Create client
-        DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).build()
+        boolean useNettyExecFactory = Boolean.valueOf(System.getenv(ENV_VAR_GDP_CONN_NETTY))
+        DockerCmdExecFactory execFactory = useNettyExecFactory ? new NettyDockerCmdExecFactory() : null
+        DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).withDockerCmdExecFactory(execFactory).build()
 
         // register buildFinished-hook to close docker client.
         project.gradle.buildFinished {


src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -157,8 +162,9 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
 
         DefaultDockerClientConfig dockerClientConfig = dockerClientConfigBuilder.build()
 
-        // Create client
-        DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).build()
+        boolean useNettyExecFactory = Boolean.valueOf(System.getenv(ENV_VAR_GDP_CONN_NETTY))
+        DockerCmdExecFactory execFactory = useNettyExecFactory ? new NettyDockerCmdExecFactory() : null
```
cdancy: If null then default jersey factory is used.

cdancy:
Not sure if there is a "clean way" to write a test for this which is why it's omitted.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
bmuschko: I am wondering why we need to switch the `DockerCmdExecFactory` completely instead of configuring the the Jersey implementation. Apparently, the `JerseyDockerCmdExecFactory` doesn't set an unlimited read and connect timeout. Otherwise you wouldn't see the issue with current implementation. I looked around in the code a bit but couldn't find where it does change the defaults. Maybe it's even a somewhat unrelated bug with the `JerseyDockerCmdExecFactory` in general.

Popping in `NettyDockerCmdExecFactory` just so an unlimited timeout is set (which seems to be the default for it) sounds a little backwards. Moreover, I'd personally consider switching out the communication layer pretty high risk (especially because it's not the default in Docker Java right). I don't think we can be sure that the Netty implementation just works the same way as the Jersey layer without hick ups.

cdancy: @bmuschko I think we're getting caught up on the `timeout` property so let me stop mentioning it. It's the the implementation in general that seems to be causing things to pass ... for whatever reason. I honestly don't know. When I run the tests locally with this change, and our pipeline where things do fail, things pass. The process itself takes maybe 10 minutes at most and doesn't trip any timeout anyway. Again ... I don't know what to make of it other than it just works.

cdancy:
Also, and if we move forward with this, I'd also like to check sys-prop and project property as passing along an env-var, at least in our setup, is a huge pain whilst passing along a project property is much easier to make happen.

Thinking something like: `GDP_INTERNAL_CONN_DRIVER`, `gdp.internal.conn.driver` and `gdpInternalConnDriver`. Again ... we won't expose this in our documentation but it's there should folks want to switch implementation for whatever reason.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: FWIW: following the code for either jersey or netty impl the timeouts, if not set, will default to infinity.

bmuschko: I think I caught up by the example you gave initially. I didn't quite understand what you meant by "the timeout is just an example". It's interesting that Docker Java still didn't address this crucial error after 3 years.

BTW: Did you see this comment https://github.com/docker-java/docker-java/issues/698#issuecomment-403502833 that could work around the issue without have to switch to Netty? Maybe that's an easier, less invasive solution to the problem.

bmuschko:
> Also, and if we move forward with this, I'd also like to check sys-prop and project property as passing along an env-var, at least in our setup, is a huge pain whilst passing along a project property is much easier to make happen.

Feel free to add that. As a sanity check I'd propose to at least run through the whole test suite with Netty enabled. I don't understand why Docker Java has that a communication option but Jersey is still the default. Somewhat not trustworthy. ;-)

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: That's an interesting solution... let me give it a go and see how things fare. I would still like to get this in though.

cdancy:
> Feel free to add that. As a sanity check I'd propose to at least run through the whole test suite with Netty enabled. I don't understand why Docker Java has that communication option but Jersey is still the default. Somewhat not trustworthy. ;-)

Yeah the thought did cross my mind. It begs the question: what was the reason for going down this path in the first place and setting up a netty backend? Was that to replace things in the future just never had the time to complete? IDK. Might ask those guys on IRC to see if they say anything.

And am going to try and circle back on things this week (hopefully today actually) to finish knock this out. Tried with the above solution, in creating the file ahead of time with a good size, but still things pop.

cdancy:
@bmuschko give things a look over when you get a chance. I've added the @ignore tag to the env-var test as we're currently on an older version of gradle that doesn't support your above suggestion. FWIW I did test in the hard-code way and that worked. I also ran a complete end-to-end test with this on by default locally and all functionalTests passed locally.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -157,8 +160,10 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
 
         DefaultDockerClientConfig dockerClientConfig = dockerClientConfigBuilder.build()
 
-        // Create client
-        DockerClient dockerClient = DockerClientBuilder.getInstance(dockerClientConfig).build()
+        DockerClient dockerClient = DockerClientBuilder
+            .getInstance(dockerClientConfig)
+            .withDockerCmdExecFactory(getDockerCmdExecFactory())
```
cdancy: If null this defaults back to jersey.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerNettyDockerCmdExecFactoryFunctionalTest.groovy
```diff
@@ -0,0 +1,57 @@
...
+            task dockerInfo(type: DockerInfo)
+        """
+
+        when:
+        BuildResult result = build('--debug', 'dockerInfo')
+
+        then:
+        result.output.contains(NETTY_DRIVER_IN_USE)
+    }
+}
```
cdancy: The whole tester amounts to simple outputs of docker-info but checking each test to see if the debug statement was output correctly.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -167,6 +172,17 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
...
 
+    private DockerCmdExecFactory getDockerCmdExecFactory() {
+        boolean useNettyExecFactory = Boolean.valueOf(getProject().findProperty('gdpNettyExecFactory').toString())
+            ?: Boolean.valueOf(System.getProperty('gdp.netty.exec.factory'))
+            ?: Boolean.valueOf(System.getenv('GDP_NETTY_EXEC_FACTORY'))
+
+        if (useNettyExecFactory) {
+            this.logger.debug("Using " + NettyDockerCmdExecFactory.class.simpleName + " as driver for " + DockerClient.class.simpleName)
+        }
+        return useNettyExecFactory ? new NettyDockerCmdExecFactory() : null
```
cdancy: FWIW: all timeouts default to infinity if not specified. I probably should not have mentioned that for fear of confusion but just putting it out there for completeness sake should we ever stumble back upon this issue down the road.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -134,6 +136,7 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
     @Internal
     @Memoized
     DockerClient getDockerClient() {
+
```
bmuschko: Can you remove the extra new line?

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -167,6 +172,17 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         dockerClient
     }
 
+    private DockerCmdExecFactory getDockerCmdExecFactory() {
+        boolean useNettyExecFactory = Boolean.valueOf(getProject().findProperty('gdpNettyExecFactory').toString())
+            ?: Boolean.valueOf(System.getProperty('gdp.netty.exec.factory'))
+            ?: Boolean.valueOf(System.getenv('GDP_NETTY_EXEC_FACTORY'))
+
+        if (useNettyExecFactory) {
+            this.logger.debug("Using " + NettyDockerCmdExecFactory.class.simpleName + " as driver for " + DockerClient.class.simpleName)
```
bmuschko: There's no need for the explicit usage of `this` here.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -167,6 +172,17 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         dockerClient
     }
 
+    private DockerCmdExecFactory getDockerCmdExecFactory() {
+        boolean useNettyExecFactory = Boolean.valueOf(getProject().findProperty('gdpNettyExecFactory').toString())
```
bmuschko: I am surprised that the method call of `toString()` doesn't fail if the `findProperty` method return null. You will likely want to catch that with via `?.toString()`.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
bmuschko: That's fine for now.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerNettyDockerCmdExecFactoryFunctionalTest.groovy
```diff
@@ -0,0 +1,57 @@
+package com.bmuschko.gradle.docker
+
+import com.github.dockerjava.api.DockerClient
+import com.github.dockerjava.netty.NettyDockerCmdExecFactory
+import org.gradle.testkit.runner.BuildResult
+import spock.lang.Ignore
+
+class DockerNettyDockerCmdExecFactoryFunctionalTest extends AbstractGroovyDslFunctionalTest {
```
bmuschko: You will want to add a test case for the negative case. If none of your Netty properties is set, then it should use the Jersey implementation and not render the message.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: That's interesting that did not fail. Maybe some groovy magic? IDK. I'll put in this fix.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: Done.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
cdancy: Done.

src/functTest/groovy/com/bmuschko/gradle/docker/DockerNettyDockerCmdExecFactoryFunctionalTest.groovy
cdancy: Done.

cdancy:
> Do you want this change to be released as part of v5.3.0 before all the 6.0.0 features go out? If you do, did you create the branch off of the latest, released tag (v5.2.0)? You can't really assume that master didn't already contain other changes after the tag.

Have not yet but can do so. My thinking was that once this PR was merged, and the commits squashed, I'd cherry-pick that commit back to 5.2.0 and release off of that.

cdancy:
All PR comments should now be addressed.

cdancy:
@bmuschko also pinged the maintainers of `docker-java` on gitter yesterday and asked about why netty is even a thing here in the first place but haven't gotten a response yet. If anything comes up I'll reply back with the response here.

cdancy:
Thanks @bmuschko . I'll fix the existing conflict now and try a release off of the aforementioned branch and see how things fare. Stay tuned....

PR #865, opened 2019-10-02
Plugin resolves and uses Docker credential helper

kgrech:
This pr closes #743.

The PR solves the problem that PushImage and PullImage build task are not able to get the credential information from the ``~/.docker/config.json`` file

PR #879, opened 2019-10-27
Improve DSL for configuring registry credentials for custom tasks

bmuschko:
* Always creates an instance of `DockerRegistryCredentials` for custom tasks
* Pass down values from extension
* Custom tasks can configure registry credential values with a DSL-like API

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractGroovyDslFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractGroovyDslFunctionalTest.groovy
@@ -68,8 +68,10 @@ abstract class AbstractGroovyDslFunctionalTest extends AbstractFunctionalTest {
 
         if (dockerPrivateRegistryUrl) {
             buildFile << """
-                docker.registryCredentials {
-                    url = '$dockerPrivateRegistryUrl'
+                docker {
+                    registryCredentials {
+                        url = '$dockerPrivateRegistryUrl'
+                    }
                 }
             """
         }
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPluginFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerRemoteApiPluginFunctionalTest.groovy
@@ -43,6 +43,37 @@ class DockerRemoteApiPluginFunctionalTest extends AbstractGroovyDslFunctionalTes
         build('verify')
     }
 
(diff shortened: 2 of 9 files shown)


cdancy:
Very cool and thanks for the PR!

cdancy:
So the PR looks OK to me but I feel like there is room to potentially use a groovy trait to encapsulate a lot of this common code and have each task implement it. Something like what I did [HERE](https://github.com/bmuschko/gradle-kubernetes-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/kubernetes/plugin/common/RetryAware.groovy).

cdancy:
Oh you merged this already? Bummer ... if I'm not mistaken the code looks like this will re-configure the existing extension point "registry credentials" instead of using a new one defined specifically for the task? If that's true it would require the user to then re-init the extension point "registry credentials" unless I'm confusing things.

bmuschko:
I am confused about what you are asking how the behavior should look like. Here's how it works:

1. The remote API plugin copies the registry credentials values of the extension to all custom tasks. Every custom task has their own instance of `DockerRegistryCredentials`.
2. If you want to overwrite the values for a specific custom task, you can do so by setting the registry credentials value for that particular task. It does not change the values of the extension.

Your initial request made it sound like you wanted a DSL-like way to set registry credentials values for custom tasks. That's why I asked to create a new issue that would specific the expected behavior.

BTW: I am planning to release 6.0.0 very soon after merging https://github.com/bmuschko/gradle-docker-plugin/pull/865 to `master`. So if you want an in-between release then you should do it soon from a different branch other than `master`.

cdancy:
Apologies for the confusion: your explanation clears things up. _I_ was getting confused by another comment I saw and flipping back and forth between PR's in tabs.

Thoughts on just using a groovy `trait` for `RegistryCredentialsAware` and not having to add these DSL `registryCredentials` methods to each task? It would simplify things and make it easier to adopt this functionality elsewhere in the future IMO.

cdancy:
Also ... working on that other PR at the moment while I have some spare cycles...

bmuschko:
> Thoughts on just using a groovy trait for RegistryCredentialsAware and not having to add these DSL registryCredentials methods to each task? It would simplify things and make it easier to adopt this functionality elsewhere in the future IMO.

A trait sounds like a bit of overkill for this logic. The method really only is one line of code. Sure, it's duplicated across a couple of classes but it's not complex. The interface `RegistryCredentialsAware` indicates where the method should be implemented. I don't see it as a pressing issue.

Issue #891, opened 2019-11-23
Unable to have multiple main methods in project when building image


Robouzou:
### Expected Behavior
Running `dockerBuildImage` should either auto-select a main method to use an the entry point (as it did when using Application Plugin), or get this from some configuration value provided to it. Maybe through `mainClassName`, though some kind of per-image configuration might be useful to people wanting to build multiple images.

### Current Behavior
Whenever you have multiple files that have `main` functions, `dockerBuildImage` will give this error:
```
Unable to find a single main class from the following candidates [{classes with main methods}]
```

### Context

The project in which this is a problem has a `Main` class, with the actual `main` entry point for the application, but it also houses a few other tools which are used occasionally by themselves, each with their own `main`.

### Steps to Reproduce (for bugs)

[multiple_mains.zip](https://github.com/bmuschko/gradle-docker-plugin/files/3882760/multiple_mains.zip)

Running `gradle dockerBuildImage` in this project generates the issue.

### Your Environment

This is with Gradle 5.6.4, openjdk 11.0.4, and any version of the plugin after commit 9cfac7f2c0f58c8247c3231629d76edfaf9b3621 / v5.1.0.

bmuschko:
The plugin you are using is looking for a [main class that is annotated with `org.springframework.boot.autoconfigure.SpringBootApplication`](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy#L32). Can you explain why your Spring Boot application does not use this annotation?

Robouzou:
That's because we're not using the Spring Boot Application Plugin, but the Java Application Plugin, [which doesn't search for any annotation to specificity the main class.](https://github.com/bmuschko/gradle-docker-plugin/blob/1910be61fab63a8da578bc2a085064ba4804e724/src/main/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPlugin.groovy#L43)

bmuschko:
Ah, OK. I must have misread your initial comment. I can [reproduce the issue](https://github.com/bmuschko/gradle-docker-plugin/commit/2f6b3374e6a3ba4b203dd96513442992dd2bcf90). I thought that we are selecting one main class from the list, however, that might not be the one you want. I'll look into some alternatives.

ypoluektovich:
[...]

PR #892, opened 2019-12-04
Allow configuring the main class name for convention plugins

bmuschko:
Covers the corner case where a project's main source set contains more than one main class file. In those situations, the end user will want to select which main class to use for starting the application.

For more information, see https://github.com/bmuschko/gradle-docker-plugin/issues/891.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionApplicationPlugin.groovy
@@ -122,7 +122,7 @@ abstract class DockerConventionApplicationPlugin<EXT extends DockerConventionApp
                                 entrypoint.addAll(jvmArgs)
                             }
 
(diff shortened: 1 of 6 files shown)


src/main/groovy/com/bmuschko/gradle/docker/DockerConventionApplicationPlugin.groovy
```diff
@@ -190,7 +190,11 @@ abstract class DockerConventionApplicationPlugin<EXT extends DockerConventionApp
         })
     }
 
-    private String getApplicationMainClassName(Project project) {
+    private String getApplicationMainClassName(Project project, EXT extension) {
+        if (extension.mainClassName.isPresent()) {
+            return extension.mainClassName.get()
```
cdancy: Should we check to see if the class is actually present or just wait until things fall over at a later time?

bmuschko: I think that would only work if the main class can be found in the same Gradle project. If the main lives in a different Gradle project e.g. as a project dependency then finding the main class would fail (see https://github.com/bmuschko/gradle-docker-plugin/issues/893).

PR #894, opened 2019-12-06
ISSUE-887: Fix error logging on docker push task when auth config fil…

mikeykoelbl:
…e is not present

This fixes some of the problems mentioned in ISSUE-887, that logging 

* If the authentication configuration file cannot be found, just return the default configuration without logging (either loglevel warn or loglevel error)
* If the authentication configuration file cannot be parsed, we will go ahead with the registry push (as before) put log as a warn. The loglevel "warn" is correct for the situation "we found a problem but tried to continue"
* I have added tests that verify that logs are written to the correct log level.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
@@ -106,6 +106,9 @@ class RegistryAuthLocator {
             configFile.exists() ? 'exists' : 'does not exist',
             commandPathPrefix)
 
+        if (!(configFile.exists())) {
+            return defaultAuthConfig
+        }
         try {
             Map<String, Object> config = slurper.parse(configFile) as Map<String, Object>
 
@@ -127,7 +130,8 @@ class RegistryAuthLocator {
             }
 
         } catch(Exception ex) {
-            logger.error('Failure when attempting to lookup auth config ' +
+            logger.warn('Failure when attempting to lookup auth config ' +
+                // logger.warn('Failure when attempting to lookup auth config ' +
                 '(docker repository: {}, configFile: {}. ' +
                 'Falling back to docker-java default behaviour',
                 repository,
--- a/src/test/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocatorTest.groovy
+++ b/src/test/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocatorTest.groovy
@@ -22,7 +22,7 @@ class RegistryAuthLocatorTest extends Specification {
         config.getRegistryAddress() == 'https://index.docker.io/v1/'
         !config.getUsername()
         !config.getPassword()
-        0 * logger.error(*_)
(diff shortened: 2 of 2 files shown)


mikeykoelbl:
I have noticed that you have changed the code I was previously working on - I shall iron our the merge conflict and push again.

mikeykoelbl:
okay, I don't know why the functional tests have failed. I have not yet found anything but will look into it later in a few hours' time.

mikeykoelbl:
Okay, I cannot reproduce the same error in the functional tests that was shown in this build:
```
com.bmuschko.gradle.docker.tasks.container.DockerWaitHealthyContainerFunctionalTest > Invoke onNext periodically passing the health status FAILED
```
My test run properly, but only when I start with a clean docker slate and with max-workers=1
```
lucinda:gradle-docker-plugin mrk-hre$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G95
```
Docker version 19.03.1
Java  1.8.0_232 (AdoptOpenJDK)
There is no $HOME/.docker/

If you need any other information, please let me know

bmuschko:
The failing functional tests is probably a flaky test case. You can ignore that for now. I will look into the root cause and fix for it soon. Moreover, I will review your code over the weekend.

src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
```diff
@@ -127,7 +130,8 @@ class RegistryAuthLocator {
             }
 
         } catch(Exception ex) {
-            logger.error('Failure when attempting to lookup auth config ' +
+            logger.warn('Failure when attempting to lookup auth config ' +
```
bmuschko: For now, I'd not want to change the log level for anything that might go wrong. It probably indicates a real issue. If the parsing of the JSON file doesn't work properly then that's a bug.

So I'd still want to see an error for the following cases:
1. The JSON cannot be parsed.
2. The auth config cannot be determined.
3. The credentials cannot be resolved.

IMHO those are all valid use cases. So if you have a config file then I'd expect it to work.

src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
```diff
@@ -106,6 +106,9 @@ class RegistryAuthLocator {
             configFile.exists() ? 'exists' : 'does not exist',
             commandPathPrefix)
 
+        if (!(configFile.exists())) {
```
bmuschko: I am OK with adding this in, however, it will only cover some of the issues users reported.

bmuschko: This should use `isFile()` instead.

bmuschko:
Thanks for the pull request. I just committed the change separately: https://github.com/bmuschko/gradle-docker-plugin/commit/fe2c7bca71cba75fd07530c1bccc60f363a5ae29

PR #898, opened 2019-12-12
ISSUE 887

kgrech:
Changes:
- Slightly improved the test case to be close to real ``docker-credential-desktop`` invocation. The response provided by the user is parsed with no error reported.
- Added a debug line so users can report the issue with credentials helper by  running the plugin with a debug flag and see the output

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
@@ -249,6 +249,7 @@ class RegistryAuthLocator {
         String credentialHelperName = commandPathPrefix + credHelper
 
         String data = runCommand(hostName, credentialHelperName)
+        logger.debug('Credential helper response: {}', data)
         Map<String, String> helperResponse = slurper.parseText(data) as Map<String, String>
         logger.debug('Credential helper provided auth config for: {}', hostName)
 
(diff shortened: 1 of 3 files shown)


src/test/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocatorTest.groovy
```diff
@@ -80,16 +80,18 @@ class RegistryAuthLocatorTest extends Specification {
         0 * logger.error(*_)
     }
 
-    def "AuthLocator returns default config for Docker Desktop config without existing credentials"() {
+    def "AuthLocator works for Docker Desktop config without existing credentials"() {
```
bmuschko: We should have both test cases:

1. The existing one: Config file can be read but the credentials couldn't be resolved
2. The config file can be read and the credentials as well.

Would you mind changing the code.

src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
```diff
@@ -249,6 +249,7 @@ class RegistryAuthLocator {
         String credentialHelperName = commandPathPrefix + credHelper
 
         String data = runCommand(hostName, credentialHelperName)
+        logger.debug('Credential helper response: {}', data)
```
bmuschko: I am sure this will be helpful in identifying the underlying issues reported by users.

src/test/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocatorTest.groovy
kgrech: @bmuschko done

Issue #905, opened 2020-01-17
JDK 13 Issues


andrewpakhomov:
Have following error when using with JDK 13 and Gradle 6.1
```
 Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':dev-infra:stopLocalDbContainer'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:205)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:203)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:184)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
        at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:52)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$CallableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:416)
[...]

PR #906, opened 2020-01-17
Issue 905 fix with 

andrewpakhomov:
Fix for https://github.com/bmuschko/gradle-docker-plugin/issues/905

Forcing version of junixsocket-core to 2.2.1
Exactly same thing as https://github.com/docker-java/docker-java/pull/1280/commits, but without next release of  docker-java.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,8 +23,13 @@ repositories {
     jcenter()
 }
 
+configurations.all {
+    resolutionStrategy.force("com.kohlschutter.junixsocket:junixsocket-core:2.2.1")
+}
+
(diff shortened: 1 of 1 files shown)


bmuschko:
This needs to be mainly fixed in Docker Java. In fact, it [has already been changed in master](https://github.com/docker-java/docker-java/blob/master/pom.xml#L71). I believe it's going to be released soon as 3.2.0-rc2 is already out. For more info, see https://github.com/docker-java/docker-java/issues/1245.

I don't really want to get into the business of forcing another version of a transitive dependency and then ensuring the correct version going forward.

PR #932, opened 2020-04-10
ISSUE-931: use a MapProperty for volumesFrom, allowing access modes

d-vesely:
This allows users of the plugin to mount volumes from running containers (using the volumesFrom / 
`--volumes-from` functionality) in read-only mode as well, instead of only the default read-write access mode. They can do this by specifying the desired access mode, either `"ro"` or `"rw"`, in the map passed assigned to `hostConfig.volumesFrom`.
An example would be:
```
    imageId = exampleImage
    hostConfig.volumesFrom = ['anotherContainer':'ro']
    containerName = exampleContainer
}
```

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.groovy
@@ -271,7 +271,7 @@ class DockerCreateContainer extends DockerExistingImage {
         }
 
         if(hostConfig.volumesFrom.getOrNull()) {
-            List<VolumesFrom> createdVolumes = hostConfig.volumesFrom.get().collect { new VolumesFrom(it) }
+            List<VolumesFrom> createdVolumes = hostConfig.volumesFrom.get().collect { VolumesFrom.parse([it.key, it.value].join(':')) }
             containerCommand.hostConfig.withVolumesFrom(createdVolumes)
         }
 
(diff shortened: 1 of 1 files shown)


bmuschko:
Thanks for providing the PR. Seems like the CI build is failing. Did you execute the tests locally?

In addition, we'll need to address the following points:

1. This is a breaking change as the public API has been changed. We may want to delay the introduction of the change to the next major version of the plugin.
2. We'll likely want to introduce a new POJO as inner class that captures the properties of a Volume so that it becomes more maintainable in the future....something like `ListProperty<Volume>`. At the moment, a user _has_ to provide an access mode and you'd need to know what the default is.
3. If there were no functional tests before, we should add a new one.

d-vesely:
According to the [CI build logs](https://travis-ci.org/github/bmuschko/gradle-docker-plugin/jobs/673368197), one test fails due to the changed API. To be honest, this PR was rather hastily done, my mistake.

One possibility for mitigating the problem of a changed API (at least until the next major version, as you mentioned), would be to add something along the lines of `hostConfig.volumesFromReadOnly` which would work exactly like `volumesFrom` (a list of containers) but would use the constructor with an access mode argument (of course fixed to "ro"), instead of the one `volumesFrom` uses, which just takes the container name as a parameter and assigns the default access mode "rw".

If you think that solution would be better for now I could submit a PR, including a new functional test.

bmuschko:
Your change already breaks one of the functional tests which would have to be fixed. For now, let's wait until we release v7.0.0 of the plugin. I am pretty sure this will break too many consumers. Closing this for now but will pick it up later.

PR #935, opened 2020-05-06
Fixes Issue-648

nirud:
Hi,

this fixes the Issue-931.

The problem was that the remote path was passed to project.file() which will convert a relative path to an absolute path based on the project dir.

Under windows, a Linux path like '/tmp' will be interpreted as a relative path and converted to an absolute path like 'C:\code\project\tmp' which of course does not exist in the linux container.

So now the remote path is passed to the container without modification.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileToContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileToContainer.groovy
@@ -83,9 +83,9 @@ class DockerCopyFileToContainer extends DockerExistingContainer {
         }
 
         if (copyFileToContainer.remotePath instanceof Closure) {
-            containerCommand.withRemotePath(copyFileToContainer.remotePath.call())
+            containerCommand.withRemotePath(copyFileToContainer.remotePath.call() as String)
         } else {
-            containerCommand.withRemotePath(copyFileToContainer.remotePath)
+            containerCommand.withRemotePath(copyFileToContainer.remotePath as String)
         }
 
         if (copyFileToContainer.isTar) {
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileToContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileToContainer.groovy
@@ -82,14 +82,12 @@ class DockerCopyFileToContainer extends DockerExistingContainer {
             localHostPath = project.file(copyFileToContainer.hostPath)
         }
 
-        def localRemotePath
         if (copyFileToContainer.remotePath instanceof Closure) {
-            localRemotePath = project.file(copyFileToContainer.remotePath.call())
+            containerCommand.withRemotePath(copyFileToContainer.remotePath.call())
         } else {
-            localRemotePath = project.file(copyFileToContainer.remotePath)
+            containerCommand.withRemotePath(copyFileToContainer.remotePath)
         }
 
-        containerCommand.withRemotePath(localRemotePath.path)
         if (copyFileToContainer.isTar) {
             containerCommand.withTarInputStream(localHostPath.newInputStream())
         } else {


cdancy:
LGTM and would definitely fix a long standing issue for windows users.

I'll let @bmuschko and @orzeh chime in for the final say however.

nirud:
Hi,

the `DockerCopyFileToContainerFunctionalTest` would have failed if it would have run on a windows system. 

But neither test nor functionalTest can run under windows for different reasons.
I tried to quickly fix the functional test because the first error is just about escaping `\` in path strings but after that, the next error appeared (somewhere in the test suite a 'unix://...' URL is used what fails on windows)

So because the test would have caught the issue if the would have run on windows and because I cannot run the tests I decided not to change any tests.

The `DockerCopyFileFromContainer` class looks fine and is working at least in my use case on windows.

I can open an issue for the test on windows problem. Because that should be fixed and maybe the travis ci should also test if the test runs on windows. I would be willing to try to work on the test.

cdancy:
Can travis run things on windows? If so, and if it's easy enough to wire-up, that would be good to get in place IMO as it would allow us to test out the functionality of using windows npipes if available.

nirud:
As far as I can see they have an "early stage" support of windows (https://docs.travis-ci.com/user/reference/windows/)

The node comes with chocolatey which can install docker and a jdk so in theory it should be doable.

bmuschko:
> As far as I can see they have an "early stage" support of windows (https://docs.travis-ci.com/user/reference/windows/)

I tried it once but AFAIR upon execution it failed. From my perspective Travis is dead in the long run anyway given that they let go most of their engineering team. When I get a chance, I wanted to try out GitHub Actions.

Some people I worked with before, said that https://www.appveyor.com/ was working well for CI on Windows.

bmuschko:
> But neither test nor functionalTest can run under windows for different reasons.

So how do you know if your change will even work?

nirud:
I build the plugin, deployed it to my local maven repository, and tried it out.

I also quickly fixed the `DockerCopyFileToContainerFunctionalTest` so that I can run it in isolation from IntelliJ. The quick fix was just appending `.replace('\', '\\')` to all file path strings witch is not a good solution so I did not want to commit this and it would not make the whole test suite run on windows.

You could of course fist make the test suites run on windows then see the `DockerCopyFileToContainerFunctionalTest` failing and then merge this pull request. That would ensure that my changes would not break any other test.

bmuschko:
@nirud I set up GitHub Action builds for Linux and Windows. As expected, Windows is a PITA. For some reason, it cannot resolve the JCenter or Maven Central host.

```
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not resolve org.eclipse.jetty.aggregate:jetty-all:9.2.5.v20141112.
     Required by:
         project :
      > Could not resolve org.eclipse.jetty.aggregate:jetty-all:9.2.5.v20141112.
         > Could not get resource 'https://jcenter.bintray.com/org/eclipse/jetty/aggregate/jetty-all/9.2.5.v20141112/jetty-all-9.2.5.v20141112.pom'.
            > Could not GET 'https://jcenter.bintray.com/org/eclipse/jetty/aggregate/jetty-all/9.2.5.v20141112/jetty-all-9.2.5.v20141112.pom'.
               > No such host is known (jcenter.bintray.com)
      > Could not resolve org.eclipse.jetty.aggregate:jetty-all:9.2.5.v20141112.
         > Could not get resource 'https://repo.maven.apache.org/maven2/org/eclipse/jetty/aggregate/jetty-all/9.2.5.v20141112/jetty-all-9.2.5.v20141112.pom'.
            > Could not GET 'https://repo.maven.apache.org/maven2/org/eclipse/jetty/aggregate/jetty-all/9.2.5.v20141112/jetty-all-9.2.5.v20141112.pom'.
               > No such host is known (repo.maven.apache.org)
```

Either way, I think we'll want to transition to GitHub Actions long-term. If someone knows what's wrong with the Windows build on GitHub Actions, let me know.

bmuschko:
@nirud I assume you have a Windows machine at hand? Could you please execute the build on `master` and post the failures into https://github.com/bmuschko/gradle-docker-plugin/issues/938?

bmuschko:
This PR aims for more complete solution: https://github.com/bmuschko/gradle-docker-plugin/pull/943. I am going to close this PR here.

PR #937, opened 2020-05-19
Default to docker hub when no registry is explicitly given

### LINKED ISSUE #936: RegistryAuthLocator fails to locate docker hub auth credentials
### Expected Behavior

Given a credentials file containing:
```
{
  "auths" : {
    "https://index.docker.io/v1/" : {
      "auth": "dXNlcm5hbWU6c2VjcmV0"
    }
  }
}
```

The `RegistryAuthLocator` should be able to locate these credentials when using an image name that does not include the Docker Hub registry prefix, e.g. `org/repo`.

### Current Behavior

It fails to find the credentials.

### Context

We are trying to run this on our CI which used to work in a previous version of the plugin but which broke in 6.0:
```
docker login -u "${username}" -p "${password}"
./gradlew pushDockerImage
```

https://github.com/bmuschko/gradle-docker-plugin/pull/937/files#diff-4a6de77d733db46d24fa79f55eb0880bR196-R212

steinybot:
Closes #936

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
@@ -94,14 +94,14 @@ class RegistryAuthLocator {
      * no credentials found
      */
     AuthConfig lookupAuthConfig(String image, AuthConfig defaultAuthConfig) {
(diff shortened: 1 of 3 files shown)


bmuschko:
Thanks for providing the pull request. I will look at it in the course of this week.

PR #940, opened 2020-06-11
Use configuration avoidance API

### LINKED ISSUE #674: Use task configuration avoidance API for task type of this plugin

gaiazov:
closes #674

The plugin already implements all the necessary logic for configuration avoidance. However, all `RegistryCredentialsAware` were still eagerly configured.

As mentioned in [Configuration avoidance migration guide](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview), instead of 
```
withType(java.lang.Class, org.gradle.api.Action)
```
one should use 
```
withType(java.lang.Class).configureEach(org.gradle.api.Action)
```

Confirmed locally that tasks are no longer eagerly configured.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
@@ -27,12 +27,12 @@ import groovy.transform.CompileStatic
 import org.gradle.api.Action
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.file.Directory
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Sync
 import org.gradle.api.tasks.TaskProvider
-
 /**
  * The abstract class for all conventional JVM application plugins.
  *
@@ -146,7 +146,12 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
                     group = DockerRemoteApiPlugin.DEFAULT_TASK_GROUP
                     description = "Copies the distribution resources to a temporary directory for image creation."
                     dependsOn project.tasks.getByName(JavaPlugin.CLASSES_TASK_NAME)
-                    into(createDockerfileTask.get().destDir)
+                    into(project.provider(new Callable<Provider<Directory>>() {
+                        @Override
+                        Provider<Directory> call() throws Exception {
+                            createDockerfileTask.get().destDir
+                        }
+                    }))
                     with(createAppFilesCopySpec(project))
                 }
             }
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
@@ -146,10 +146,10 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
                     group = DockerRemoteApiPlugin.DEFAULT_TASK_GROUP
                     description = "Copies the distribution resources to a temporary directory for image creation."
                     dependsOn project.tasks.getByName(JavaPlugin.CLASSES_TASK_NAME)
-                    into(project.provider(new Callable<Provider<Directory>>() {
+                    into(project.provider(new Callable<Directory>() {
                         @Override
-                        Provider<Directory> call() throws Exception {
-                            createDockerfileTask.get().destDir
+                        Directory call() throws Exception {
+                            createDockerfileTask.get().destDir.get()
                         }
                     }))
                     with(createAppFilesCopySpec(project))
@@ -195,7 +195,12 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
                     group = DockerRemoteApiPlugin.DEFAULT_TASK_GROUP
                     description = 'Pushes created Docker image to the repository.'
                     dependsOn dockerBuildImageTask
-                    images.set(dockerBuildImageTask.get().getImages())
+                    images.convention(project.provider(new Callable<Set<String>>() {
+                        @Override
+                        Set<String> call() throws Exception {
+                            dockerBuildImageTask.get().getImages().get()
+                        }
+                    }))
                 }
             }
         })
--- a/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
@@ -15,6 +15,11 @@
  */
 package com.bmuschko.gradle.docker
 
+import static com.bmuschko.gradle.docker.internal.ConventionPluginHelper.createAppFilesCopySpec
+import static com.bmuschko.gradle.docker.internal.ConventionPluginHelper.getMainJavaSourceSetOutput
+
+import java.util.concurrent.Callable
+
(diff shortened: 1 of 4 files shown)


bmuschko:
Thanks for providing the pull request. Your change is not sufficient. We'll want to use use the configuration avoidance API across all plugins in the code base. Are you planning to work on it? If not, we'll close this PR as it will likely cause more issues than solve them due to its inconsistent use.

gaiazov:
@bmuschko I pushed a second commit. Is that what you had in mind?

bmuschko:
Along those lines but I am sure you'll have to make addition tweaks. Add a functional test similar to [this one](https://github.com/gradle/playframework/blob/master/src/integTest/groovy/org/gradle/playframework/WellBehavedPluginTest.groovy#L19-L35). The list of realized tasks should be empty if you got it right.

gaiazov:
@bmuschko added two tests to existing functional test class. Had to update `DockerConventionPluginFixture` but otherwise no unexpected changes were needed.

Lets see if CI passes 🤞

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
```diff
@@ -176,15 +182,15 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
...
+    private static void registerPushImageTask(Project project, TaskProvider<DockerBuildImage> dockerBuildImageTask) {
+        project.tasks.register(PUSH_IMAGE_TASK_NAME, DockerPushImage, new Action<DockerPushImage>() {
             @Override
             void execute(DockerPushImage pushImage) {
                 pushImage.with {
                     group = DockerRemoteApiPlugin.DEFAULT_TASK_GROUP
                     description = 'Pushes created Docker image to the repository.'
                     dependsOn dockerBuildImageTask
-                    images.set(dockerBuildImageTask.getImages())
+                    images.set(dockerBuildImageTask.get().getImages())
```
bmuschko, line 193: Can you explain why this `get()` is needed? I'd think that it would actually realize the value during configuration.

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
```diff
@@ -132,23 +138,23 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
...
+    private static TaskProvider<Sync> registerSyncBuildContextTask(Project project, TaskProvider<Dockerfile> createDockerfileTask) {
+        project.tasks.register(SYNC_BUILD_CONTEXT_TASK_NAME, Sync, new Action<Sync>() {
             @Override
             void execute(Sync sync) {
                 sync.with {
                     group = DockerRemoteApiPlugin.DEFAULT_TASK_GROUP
                     description = "Copies the distribution resources to a temporary directory for image creation."
                     dependsOn project.tasks.getByName(JavaPlugin.CLASSES_TASK_NAME)
-                    into(createDockerfileTask.destDir)
+                    into(createDockerfileTask.get().destDir)
```
bmuschko, line 149: I don't think the `get()` is needed here.

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
```diff
@@ -67,16 +68,21 @@ abstract class DockerConventionJvmApplicationPlugin<EXT extends DockerConvention
...
 
         project.plugins.withType(JavaPlugin) {
-            Dockerfile createDockerfileTask = createDockerfileTask(project, extension)
-            Sync syncBuildContextTask = createSyncBuildContextTask(project, createDockerfileTask)
-            createDockerfileTask.dependsOn syncBuildContextTask
-            DockerBuildImage dockerBuildImageTask = createBuildImageTask(project, createDockerfileTask, extension)
-            createPushImageTask(project, dockerBuildImageTask)
+            TaskProvider<Dockerfile> createDockerfileTask = registerDockerfileTask(project, extension)
+            TaskProvider<Sync> syncBuildContextTask = registerSyncBuildContextTask(project, createDockerfileTask)
+            createDockerfileTask.configure(new Action<Dockerfile>() {
```
bmuschko, line 73: I can look at this myself but it somewhat looks like a circular dependency between these tasks of some sort.

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
gaiazov: You're right, the `get()` call forces the `dockerBuildImage` task to be realized. However, this code is inside the closure of `tasks.register('dockerPushImage')`. This closure is only executed if the `dockerPushImage` task is to be executed.

And if the `dockerPushImage` is to be executed, the `dockerBuildImage` will be executed as well (since it is a dependency). So it is OK to call the `get()` method to force the dependent task to be configured

I agree its not the most elegant thing though

bmuschko: Sorry, I think you can either leave off `get()` or use `convention()`. Why do you _have_ to call `get()` here? You can set a Provider for a Provider.

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
gaiazov: Yeah it does look like that. I debugged it and stepped through the code.

Here's what happens when `dockerCreateDockerfile` task is executed
1. During the configuration phase, the `dockerCreateDockerfile`
2. During configuration, `dockerCreateDockerfile` tasks is configured to depend on `dockerSyncBuildContext`. However, the dependency is done via `TaskProvider`, so when the dependency is declared, the `dockerSyncBuildContext` has not yet been configured
3. The `dockerSyncBuildContext` is configured
4. The `dockerCreateDockerfile.get()` call just returns the configured task (which finished in step 3)

I'm open for suggestions to how improve this, but it does appear to work just fine

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
gaiazov: It is needed to get the `destDir` property which only exists on `Dockerfile` task and not `TaskProvider<Dockerfile>`

What happens when `dockerSyncBuildContext` is executed
1. `dockerSyncBuildContext` is configured
2. During configuration, the `createDockerfileTask.get()` call configures `dockerCreateDockerfile` task
3. `dockerCreateDockerfile` configuration is finished
4. `dockerSyncBuildContext` configuration is finished
5. `dockerSyncBuildContext` is executed
6. `dockerCreateDockerfile` is NOT executed. It only gets configured so we can calculate the `destDir` property correctly

gaiazov:
@bmuschko comments addressed to best of my ability!

src/main/groovy/com/bmuschko/gradle/docker/DockerConventionJvmApplicationPlugin.groovy
gaiazov: Ok I _think_ I understand. Can you look at the last 2 commits?

bmuschko:
Thanks for the pull request. I has been merged.

PR #942, opened 2020-06-12
Default to communication transport to Apache HttpClient 5

### LINKED ISSUE #917: Support for npipes in docker-java upcoming. Any plans for integrating this?
### Current Behavior
As of now, the gradle-docker-plugin requires that on Windows 10 tcp://localhost:2375 is exposed (cf. #854). This is due to limitations of docker-java (explained in #769).

As a consequence, you cannot use this plugin under Windows 10 with Linux containers, because in recent versions of Docker Desktop exposing the port [only works with Windows containers](https://github.com/docker/for-win/issues/453). (I've never tried this [horrible workaround](https://youtrack.jetbrains.com/issue/WI-32733#comment=27-1549346).)

Surprisingly, after 3 years of waiting, there's a [release candidate of docker-java](https://github.com/docker-java/docker-java/issues/765#issuecomment-584524438) that's supposed to fix the underlying issues.

It would be very helpful, if a)someone could verify that using this RC of docker-java fixes the problem on Windows and b)maybe provide a rough estimate about how long it would take until a release that uses the new docker-java version after it's released.

Thanks

astudentofblake:
replaces the default transport with okhttp to allow the use of npipes

this is a like for like swap and the existing tests exercise the transport
there are a couple of tweaks to the tests

closes #917

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
@@ -16,7 +16,7 @@ class DockerMultipleClientFunctionalTest extends AbstractGroovyDslFunctionalTest
                 }
             }
             task dockerClient2(type: DockerOperation) {
-                url = 'tcp://docker.corp.com'
+                url = 'tcp://docker.corp.com:2375'
                 onNext { client ->
                     if (client != null) {
                         logger.quiet "config: " + client.dockerClientConfig.toString()
@@ -24,7 +24,7 @@ class DockerMultipleClientFunctionalTest extends AbstractGroovyDslFunctionalTest
                 }
             }
             task dockerClient3(type: DockerOperation) {
-                url = 'tcp://docker.school.edu'
+                url = 'tcp://docker.school.edu:2375'
                 certPath = new File('/tmp')
                 onNext { client ->
                     if (client != null) {
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
@@ -22,6 +22,8 @@ import com.github.dockerjava.api.DockerClient
 import com.github.dockerjava.api.command.DockerCmdExecFactory
 import com.github.dockerjava.core.DefaultDockerClientConfig
 import com.github.dockerjava.core.DockerClientBuilder
+import com.github.dockerjava.core.DockerClientConfig
+import com.github.dockerjava.httpclient5.ApacheDockerHttpClient
 import com.github.dockerjava.netty.NettyDockerCmdExecFactory
 import groovy.transform.CompileStatic
 import groovy.transform.Memoized
@@ -165,11 +167,7 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
 
         DefaultDockerClientConfig dockerClientConfig = dockerClientConfigBuilder.build()
 
-        DockerClient dockerClient = DockerClientBuilder
-            .getInstance(dockerClientConfig)
-            .withDockerCmdExecFactory(getDockerCmdExecFactory())
-            .build()
-
+        DockerClient dockerClient = buildDockerClient(dockerClientConfig)
         // register buildFinished-hook to close docker client.
         project.gradle.buildFinished {
             dockerClient.close()
@@ -177,15 +175,27 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
         dockerClient
     }
 
-    private DockerCmdExecFactory getDockerCmdExecFactory() {
+    private DockerClient buildDockerClient(DockerClientConfig config) {
         boolean useNettyExecFactory = Boolean.valueOf(getProject().findProperty('gdpNettyExecFactory')?.toString())
             ?: Boolean.valueOf(System.getProperty('gdp.netty.exec.factory'))
             ?: Boolean.valueOf(System.getenv('GDP_NETTY_EXEC_FACTORY'))
 
         if (useNettyExecFactory) {
             logger.debug("Using " + NettyDockerCmdExecFactory.class.simpleName + " as driver for " + DockerClient.class.simpleName)
+            DockerClientBuilder
+                .getInstance(config)
+                .withDockerCmdExecFactory(new NettyDockerCmdExecFactory())
+                .build()
         }
-        return useNettyExecFactory ? new NettyDockerCmdExecFactory() : null
+        else {
+            DockerClientBuilder
+                .getInstance(config)
+                .withDockerHttpClient(new ApacheDockerHttpClient.Builder()
+                    .dockerHost(config.getDockerHost())
+                    .sslConfig(config.getSSLConfig())
+                    .build())
+                .build()
+        }
     }
 
     /**
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,7 +24,8 @@ repositories {
 }
 
 dependencies {
-    shaded("com.github.docker-java:docker-java:3.2.0")
+    shaded("com.github.docker-java:docker-java:3.2.2")
+    shaded("com.github.docker-java:docker-java-transport-httpclient5:3.2.2")
     shaded("javax.activation:activation:1.1.1")
     shaded("org.ow2.asm:asm:7.3.1")
     testImplementation("org.spockframework:spock-core:1.2-groovy-2.5") {
--- a/src/docs/asciidoc/12-extension.adoc
+++ b/src/docs/asciidoc/12-extension.adoc
@@ -73,5 +73,5 @@ include::{samplesCodeDir}/remote-api-plugin/basic/kotlin/build.gradle.kts[tags=e
 
 On Unix the Docker daemon listens by default on `unix:///var/run/docker.sock`.
 
-On Windows the Docker daemon listens by default on `npipe:////./pipe/docker_engine` though this is not currently supported.
-We instead fall back to `tcp://127.0.0.1:2375`.
+On Windows the Docker daemon listens by default on `npipe:////./pipe/docker_engine` if this exists.
+We fall back to `tcp://127.0.0.1:2375` where the pipe does not exist.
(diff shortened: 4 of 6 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
```diff
@@ -16,15 +16,15 @@ class DockerMultipleClientFunctionalTest extends AbstractGroovyDslFunctionalTest
                 }
             }
             task dockerClient2(type: DockerOperation) {
-                url = 'tcp://docker.corp.com'
+                url = 'tcp://docker.corp.com:2375'
```
bmuschko, line 19: Why was this change required here?

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -165,27 +167,36 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
...
 
-    private DockerCmdExecFactory getDockerCmdExecFactory() {
+    private DockerClient buildDockerClient(DockerClientConfig config) {
         boolean useNettyExecFactory = Boolean.valueOf(getProject().findProperty('gdpNettyExecFactory')?.toString())
             ?: Boolean.valueOf(System.getProperty('gdp.netty.exec.factory'))
             ?: Boolean.valueOf(System.getenv('GDP_NETTY_EXEC_FACTORY'))
 
         if (useNettyExecFactory) {
             logger.debug("Using " + NettyDockerCmdExecFactory.class.simpleName + " as driver for " + DockerClient.class.simpleName)
+			DockerClientBuilder
```
bmuschko, line 185: Can you fix the indentation here?

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
```diff
@@ -165,27 +167,36 @@ abstract class AbstractDockerRemoteApiTask extends DefaultTask {
...
+			.build()
         }
-        return useNettyExecFactory ? new NettyDockerCmdExecFactory() : null
+        else {
+			DockerClientBuilder
+			.getInstance(config)
+			.withDockerHttpClient(new OkDockerHttpClient.Builder()
+                            .dockerHost(config.getDockerHost())
+                            .sslConfig(config.getSSLConfig())
+							.connectTimeout(30 * 100)
```
bmuschko, line 196: Was there a reason why you provided a different connect timeout here than whatever is default?

astudentofblake: I wasn't too sure what the timeout should be: -
  netty's timeout seems to default to 30 seconds
  jersey's timeout seems to default to 0
  okhttp's default timeout is 10 seconds

using the default 10 seems fine, unless you want to be consistent between netty and okhttp (in which case it should probably be (30 * 1000), or want to keep the same behaviour as jersey and set to 0.

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
astudentofblake: yes, of course

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
astudentofblake: Unlike the JerseyDockerHttpClient 
`case "tcp":
                try {
                    dockerHost = new URI(dockerHost.toString().replaceFirst("tcp", protocol));`
 
the OkDockerHttpClient 
`           case "tcp":
                baseUrlBuilder = new HttpUrl.Builder()
                    .scheme(isSSL ? "https" : "http")
                    .host(dockerHost.getHost())
                    .port(dockerHost.getPort());
                break;`
has a specific call to getPort. This returns -1 where no port is set, and you get a failure trying to connect to port -1
|      * What went wrong:
|      Execution failed for task ':dockerInfo'.
|      > unexpected port: -1

I could alter the URI to set a default port, or detect this and output a more accurate message: -
`			if (config.getDockerHost().getScheme() == "tcp" && config.getDockerHost().getPort() == -1) {
				throw new GradleException("Dockerhost uri with tcp scheme must contain port")
			}`
and add a test, but neither seemed to add much.

bsideup:
FTR I would recommend trying the Apache HttpClient5 transport instead. See https://github.com/docker-java/docker-java/blob/master/docs/transports.md for more info

src/main/groovy/com/bmuschko/gradle/docker/tasks/AbstractDockerRemoteApiTask.groovy
bsideup: @bmuschko good question.

@astudentofblake I would recommend the same rule as with JVM's flags. "Don't touch them if you don't need/understand them" :D So, unless you have a good motivation to change the defaults, I would recommend keeping them.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
bsideup: The change is correct. The port should be specified explicitly, there is no guarantee that `2375` will be implicitly used, and it is there in Jersey only for historical reasons

bmuschko:
@bsideup Great documentation about this topic.
@astudentofblake Can we please switch to the HttpClient implementation?

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/DockerMultipleClientFunctionalTest.groovy
bmuschko: Got it, thanks.

astudentofblake:
I made the switch to  Apache HttpClient 5 and updated the PR.

This seems to expose an issue in the liveness probe, which causes a test failure when running the tests in windows, which I have documented here: -

[Gap in logs for liveness probe #946](https://github.com/bmuschko/gradle-docker-plugin/issues/946)

bmuschko:
@cdancy Once we release this, I'd ask you to complete switch over to HttpClient instead of Netty. My assumption is that you had issues earlier as Jersey has issues internally. I'd love to through out the Netty code path if HttpClient just works fine.

bsideup:
@bmuschko FYI I am about to release `docker-java` 3.2.5 containing a fix for a leak, ETA ~1-2h, consider waiting for it.

cdancy:
@bmuschko I'll give things a go when a new version is kicked.

bsideup:
Update: 3.2.5 is out now :)

bmuschko:
@bsideup Thanks, https://github.com/bmuschko/gradle-docker-plugin/commit/171a0ca1fd601229d00d9a395100bce0e4f3c905.

bmuschko:
@bsideup I see the following stack trace quite often:

```
org.apache.hc.core5.http.StreamClosedException: Stream already closed
at org.apache.hc.core5.http.impl.io.IdentityInputStream.read(IdentityInputStream.java:85)
at org.apache.hc.core5.http.io.EofSensorInputStream.read(EofSensorInputStream.java:118)
at com.github.dockerjava.core.FramedInputStreamConsumer.accept(FramedInputStreamConsumer.java:53)
at com.github.dockerjava.core.FramedInputStreamConsumer.accept(FramedInputStreamConsumer.java:12)
at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:275)
at java.base/java.lang.Thread.run(Thread.java:834)
```

Is this somewhat of a known issue? For more info, see https://github.com/bmuschko/gradle-docker-plugin/runs/809758405.

bsideup:
@bmuschko from the link provided, I only see the following failure:
```java
com.bmuschko.gradle.docker.tasks.container.DockerExecContainerFunctionalTest > Execute multiple commands within running container FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at DockerExecContainerFunctionalTest.groovy:63
```

Can you share a reproducer?

bmuschko:
@bsideup Sorry, meant to post the link to the build scan: https://scans.gradle.com/s/czprsbfyzcoyy/tests/:functionalTest/com.bmuschko.gradle.docker.tasks.container.DockerExecContainerFunctionalTest/Execute%20multiple%20commands%20within%20running%20container#1. Couldn't reproduce locally yet. The test might have become flaky by some underlying change. I will need to look it more deeply.

bsideup:
@bmuschko ah, okay, I see it now :) This doesn't look good, I will check on my side meanwhile. We haven't seen it in Testcontainers (yet?).

Thanks for letting me know!

bsideup:
@bmuschko to unblock the release, you may want to use the OkHttp transport which is more battle tested, with an option to switch to AHC5 for the early adopters

bmuschko:
@bsideup It's absolutely possible that there's an issue in the implementation of `com.bmuschko.gradle.docker.tasks.container.DockerExecContainer` which has been exposed by a different transport protocol. Yeah, we might want to go for OkHttp for now. I will make that change over the weekend. Certainly let me know in case you uncover an underlying issue. Thanks!

bmuschko:
@bsideup There seems to be a similar issue with OkHttp. Any ideas?

```
Error during callback
java.io.IOException: closed
at okio.RealBufferedSource$1.read(RealBufferedSource.java:434)
at com.github.dockerjava.core.FramedInputStreamConsumer.accept(FramedInputStreamConsumer.java:53)
at com.github.dockerjava.core.FramedInputStreamConsumer.accept(FramedInputStreamConsumer.java:12)
at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:275)
at java.base/java.lang.Thread.run(Thread.java:834)
```

bsideup:
@bmuschko oh no... Is it something that you can reproduce? I've never seen in in docker-java's nor testcontainers' tests :(

bmuschko:
@bsideup Yeah, check out `master` and run `./gradlew clean fT --tests com.bmuschko.gradle.docker.tasks.container.DockerExecContainerFunctionalTest."Execute multiple commands within running container"`. I was looking at the implementation of the task yesterday but couldn't pinpoint an apparent reason for the failure.

bsideup:
@bmuschko since I need to debug `docker-java`'s internals to understand it better, would it be possible for you to provide an isolated reproducer (even better if it will be in Java 😅)?

bmuschko:
@bsideup Yeah, will look into it. Will probably need a couple of days until I get to it.

bmuschko:
@bsideup Looks like that was an issue with the test cases. The container finished before the the command could be completed.

Issue #946, opened 2020-06-24
Gap in logs for liveness probe


astudentofblake:
In the calls in the liveness probe to get the logs, there can be gaps in the logs caused by the since clause.

### Expected Behavior
DockerLivenessFunctionalTest, Can start a container and probe it for liveness should pass

### Current Behavior
DockerLivenessFunctionalTest, Can start a container and probe it for liveness fails

### Context
This only reproduces in windows, with a new transport + a certain config, but it seems like a flaw in how the liveness probe works.

### Steps to Reproduce (for bugs)
failing in the functional test: -
  Class com.bmuschko.gradle.docker.tasks.container.DockerLivenessFunctionalTest
     Can start a container and probe it for liveness

### Your Environment
In windows, 
using npipes
with apache httpclient 5 transport (https://github.com/bmuschko/gradle-docker-plugin/pull/942)
with docker desktop and linux containers

### details

the liveness probe takes the time A,
reads the entire log, checks for the text
doesn't find the text.
takes the time B
reads the log since time A (before the first call for the logs), checks for the text
doesn't find the text.
takes the time C
reads the log since time B, checks for the text

logs for this test show: -

since forever (first read of entire log) gets

The files belonging to this database system will be owned by user "postgres".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.utf8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

fixing permissions on existing directory /var/lib/postgresql/data ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default timezone ... UTC
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok

next log read with since: -
since Wed Jun 24 18:38:31 2020
logLine:
since Wed Jun 24 18:38:36 2020
logLine:

docker logs -t shows
2020-06-24T18:38:23.603749500Z The files belonging to this database system will be owned by user "postgres".
2020-06-24T18:38:23.603778200Z This user must also own the server process.
2020-06-24T18:38:23.603785300Z
2020-06-24T18:38:23.603789400Z The database cluster will be initialized with locale "en_US.utf8".
2020-06-24T18:38:23.603793500Z The default database encoding has accordingly been set to "UTF8".
2020-06-24T18:38:23.603797200Z The default text search configuration will be set to "english".
2020-06-24T18:38:23.603801100Z
[...]

Issue #886, opened 2019-11-16
DockerLivenessContainer assumes that field `sink` is of type StringWriter


bmuschko:
### Expected Behavior

The class `DockerLivenessContainer` handles the field `sink` appropriately. At the moment the implementation assumes that field `sink` is typed as `StringWriter`. It calls the method `getBuffer()` on it. There needs to be appropriate handling for it.

### Current Behavior

Passing an instance of `Writer` to the field `sink` can potentially break at runtime. The method call [`getSink().getBuffer()`](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/extras/DockerLivenessContainer.groovy#L115) would throw an exception if the field is not of type `StringWriter`.

### Context

This issue can easily be revealed by adding the `@CompileStatic` annotation to the class `DockerLivenessContainer`.

cdancy:
@bmuschko I'll knock this out within the next week as we have an existing feature request we'd like to get in for this specific task (liveness based on the truthiness of a passed closure).

bmuschko:
@cdancy Were you planning to submit a PR for this soon?

bmuschko:
Plan is to deprecated and remove those classes.

PR #964, opened 2020-08-22
Change to more concrete data type

bmuschko:
Subclasses calls method on `StringWriter` which wouldn't work with a plain `Writer`. See https://github.com/bmuschko/gradle-docker-plugin/issues/886.

I believe there's a bug in the code in the same area as reported with https://github.com/bmuschko/gradle-docker-plugin/issues/946. Probably somewhere here:

```
logLine = null
getSink().getBuffer().setLength(0)
```

BTW: What do you use this class for? Is it even being used in your project?

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerLogsContainer.groovy
@@ -99,7 +99,7 @@ class DockerLogsContainer extends DockerExistingContainer {
      */
     @Input
     @Optional
-    Writer sink
+    StringWriter sink
 
     // Allows subclasses to carry their own logic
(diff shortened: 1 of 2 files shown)


bmuschko:
@cdancy Making this work would require significant changes to the underlying tasks.

FYI: My plan is to deprecated the extra tasks and remove them with the next major version of the plugin. They are better suited for a custom plugin that builds on top of this plugin. You will want to move that code into your own infrastructure.

cdancy:
@bmuschko that's fair. As you've said elsewhere: I'm not sure how much the community gets from them but we used them internally for various purposes and thought they could help/benefit the community. We can always pull them back in :)

cdancy (approved):

bmuschko:
Plan is to deprecate and remove those classes.

PR #978, opened 2020-12-10
Add support for HEALTHCHECK

### LINKED ISSUE #754: Introduce a healthCheck property to DockerCreateContainer task type
Docker-java library supports setting health check instruction on a created container even though image doesn't have a `HEALTHCHECK` instruction set but currently `DockerCreateContainer`task  doesn't allow any method for setting the underlying `containerCommand.withHealthcheck(...)`

https://docs.docker.com/engine/reference/builder/#healthcheck

https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/command/CreateContainerCmd.java#L54

fluxroot:
This PR adds support for the HEALTHCHECK command. So I guess it fixes #754.


bmuschko (approved):
This looks pretty good.

PR #980, opened 2020-12-18
Added subnet functionality to DockerCreateNetwork.

### LINKED ISSUE #979: Add "subnet" to the DockerCreateNetwork task
### Expected Behavior

Adding the capability for the DockerCreateNetwork task to allow for `subnet` as a configuration value. From PR #449 it looks like this was meant to be added for use with the `ipV4Address` feature but never made it in.

### Current Behavior

Presently DockerCreateNetwork only allows for configuring the network name.

### Context

We have a usecase for setting up containers with specific `ipV4Address` values for networking-related tests, however we ran into an issue requiring for the network subnet to be specified before we're able to set the IPs.

N/A

### Your Environment
N/A

Will submit a corresponding PR soon :)

wrijeff:
From the related issue #979 : Adding the capability for the DockerCreateNetwork task to allow for subnet as a configuration value.

**On adding new code**
_Any new code added must contain relevant functionalTest(s) exercising all possible paths the code may take._

New functional test added to test subnet functionality. Left existing functional tests in place to confirm that _not_ specifying a subnet does not cause an error (subnet is optional). 

Test results:
```
[/workplace/github/gradle-docker-plugin] ./gradlew functionalTest

> Configure project :
Inferred project: gradle-docker-plugin, version: 6.6.2-SNAPSHOT

> Task :functionalTest
WARNING: An illegal reflective access operation has occurred
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.vmplugin.v7.Java7$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create and tear down a network STARTED

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create and tear down a network PASSED

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create and tear down a network with a specified subnet STARTED

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create and tear down a network with a specified subnet PASSED

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create a container and assign a network and alias STARTED

com.bmuschko.gradle.docker.tasks.network.DockerNetworkFunctionalTest > can create a container and assign a network and alias PASSED
```

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
@@ -1,11 +1,14 @@
 package com.bmuschko.gradle.docker.tasks.network
 
 import com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask
+import com.github.dockerjava.api.command.CreateNetworkCmd
 import com.github.dockerjava.api.command.CreateNetworkResponse
+import com.github.dockerjava.api.model.Network
 import groovy.transform.CompileStatic
 import org.gradle.api.provider.Property
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.Optional
 
 @CompileStatic
 class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
@@ -17,6 +20,10 @@ class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
     @Input
     final Property<String> networkName = project.objects.property(String)
 
+    @Input
+    @Optional
+    final Property<String> subnet = project.objects.property(String)
+
     /**
      * The id of the created network.
      */
@@ -25,7 +32,14 @@ class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
 
     void runRemoteCommand() {
         logger.quiet "Creating network '${networkName.get()}'."
-        CreateNetworkResponse network = dockerClient.createNetworkCmd().withName(networkName.get()).exec()
+
+        CreateNetworkCmd networkCmd = dockerClient.createNetworkCmd().withName(networkName.get())
+
+        if (subnet.getOrNull()) {
+            networkCmd.withIpam(new Network.Ipam().withConfig(new Network.Ipam.Config().withSubnet(subnet.get())))
+        }
+
+        CreateNetworkResponse network = networkCmd.exec()
 
         if (nextHandler) {
             nextHandler.execute(network)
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
```diff
@@ -25,7 +32,14 @@ class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
 
     void runRemoteCommand() {
         logger.quiet "Creating network '${networkName.get()}'."
-        CreateNetworkResponse network = dockerClient.createNetworkCmd().withName(networkName.get()).exec()
+
+        CreateNetworkCmd networkCmd = dockerClient.createNetworkCmd().withName(networkName.get())
+
+        if (subnet.getOrNull()) {
+            networkCmd.withIpam(new Network.Ipam().withConfig(new Network.Ipam.Config().withSubnet(subnet.get())))
```
bmuschko, line 39: We have to model this functionality thinking forward into the future. We should introduce an inner class `Ipam` and then have a list of `Config` similar to what Docker Java does. Instead of a field `subnet`, we'd have a field called `ipam`. With that we'll be able to easily add more fields going forward.

bmuschko (changes requested):

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
wrijeff: Pushed a commit to add the inner Ipam class which expects a list of maps to parse (similar to how [docker-java's network ipam](https://github.com/docker-java/docker-java/blob/c3661f3da85ebf4f9caea5a75d1c9224d2937876/docker-java-api/src/main/java/com/github/dockerjava/api/model/Network.java#L138) has a list of Configs). I had wanted to use `@Nested` within `@Nested` so we would have 
```
@Nested
@Optional
ListProperty<Config> config
```

Instead of the generic map, but had trouble getting that to work. Please let me know if you have any suggestions, I'm very new to both groovy and the Gradle API.

bmuschko: Hhmmm, did you try to create that with `objectFactory.newInstance` as well? Maybe that's a Gradle limitation. What exactly didn't work?

src/main/groovy/com/bmuschko/gradle/docker/tasks/network/DockerCreateNetwork.groovy
```diff
@@ -35,4 +62,36 @@ class DockerCreateNetwork extends AbstractDockerRemoteApiTask {
...
         logger.quiet "Created network with ID '$createdNetworkId'."
     }
+
+
+    static class Ipam {
+        private static final String SUBNET = "subnet"
+
+        @Input
+        @Optional
+        ListProperty<Map> config
```
bmuschko, line 72: This should be a List of complex objects, in this case `Config`. A Map is too generic. I can touch this up on my end.

Issue #977, opened 2020-11-28
Cannot set `DockerRemoveImage.targetImageId` using `DockerBuildImage.imageId`


rkrisztian:
I believe, ever since `DockerBuildImage.images` was introduced, it is not possible anymore to reuse `DockerBuildImage.imageId`. So in fact, none of the documented examples relying on it should work, e.g.:

```groovy
import com.bmuschko.gradle.docker.tasks.container.*
import com.bmuschko.gradle.docker.tasks.image.*

plugins {
    id 'com.bmuschko.docker-remote-api' version '6.6.1'
}

task buildMyAppImage(type: DockerBuildImage) {
    inputDir = file('docker/myapp')
    images.add('test/myapp:latest')
}

task createMyAppContainer(type: DockerCreateContainer) {
    dependsOn buildMyAppImage
    targetImageId buildMyAppImage.getImageId()
    hostConfig.portBindings = ['8080:8080']
    hostConfig.autoRemove = true
}
```

Because I get:

```none
Caused by: org.gradle.api.InvalidUserDataException: No value has been specified for property 'imageId'.
```

What works for me is:

```groovy
task removeMyAppImage(type: DockerRemoveImage) {
	force = true
	targetImageId { buildImage.images.get().first() }
}
```

Can you tell me if I'm doing something wrong? Like, I can set up an example project with Gradle 6.7.1 if needed. But I was hoping you could quickly reproduce the error.

bmuschko:
The task `DockerBuildImage` doesn't deal with a single image ID anymore. The underlying data type is now a list. `DockerBuildImage.imageId` has been removed from the API. What you are doing right now is the correct way.

rkrisztian:
The **docs are misleading**/outdated then. They make us believe that the `getImageId()` still works, but not. Look at 2.5.3 and 2.5.4. As I said before:

> none of the documented examples relying on it should work`.

And I wanted to clarify that first.

bmuschko:
I didn't really look at this project for months now. My previous statement was not correct.

Whenever you can and when in doubt, look at the API docs. Examples are really only examples and show a specific use case. BTW: All examples in the user guide are tested and should work as-is.

[`DockerBuildImage.html#imageId`](https://bmuschko.github.io/gradle-docker-plugin/api/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.html#imageId): The id of the image built.

[...]

PR #982, opened 2020-12-30
Allow configuration of `DockerExistingImage#targetImageId` with a build task.

rkrisztian:
Based on discussion #977.

This is like a convenience functionality where the image ID of
`DockerBuildImage` task will be used if the image was built in the same Gradle
build, otherwise the first image name/tag will be used from
`DockerBuildImage#images` (assuming it exists on the local filesystem).

I believe that this kind of auto-configuration should fit well for the majority
of the use cases.


bmuschko:
This change couples the `DockerBuildImage` task too strongly to the abstract class that only knows about the image ID. I'd rather not make the API more complicated and avoid merging this change. I see where you are coming from but this can easily be handled by consumers of the plugin. You may want to wrap the plugin with your own functionality that adds something like this. Thanks again for your work.

Issue #990, opened 2021-02-26
improve Dockerfile.addFile(File) method documentation


matteomazza91:
### Expected Behavior
documentation of addFile(File) method in Dockerfile class should link to [Dockerfile.File](https://bmuschko.github.io/gradle-docker-plugin/api/com/bmuschko/gradle/docker/tasks/image/Dockerfile.File.html) instead of [java.io.File
](https://docs.oracle.com/javase/8/docs/api/java/io/File.html)

### Current Behavior
at the moment, both in the summary and in the method description of this page: https://bmuschko.github.io/gradle-docker-plugin/api/com/bmuschko/gradle/docker/tasks/image/Dockerfile.html#addFile(java.io.File),  
the link for the File type of the addFile parameter is pointing to java.io.File: https://docs.oracle.com/javase/8/docs/api/java/io/File.html

### Context
Improve documentation

bmuschko:
Oooops, care to send a PR?

matteomazza91:
today I tried to checkout the repo and generate the doc. I got this error in compilation:
```
'forRemoval'is not part of the annotation java.lang.Deprecated -> java.lang.Deprecated in @java.lang.Deprecated
 @ line 37, column 1.
   @Deprecated(forRemoval = true)
```
I saw there is 'sourceCompatibility = JavaVersion.VERSION_1_8' on the buildscript but `forRemoval` is not present in java 8. 

anyway, I guess I would need a newer JDK.

bmuschko:
What task did you run? `javadoc`?

matteomazza91:
I tried both `javadoc` and `groovydoc` actually. The error happens at compilation time in `compileGroovy` task

bmuschko:
You are right. The attribute has been introduced with Java 9: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Deprecated.html#forRemoval(). Looks like `sourceCompatibility` doesn't have an effect on `groovydoc`. We'll probably want to build that code with Java 11 anyway. There's no point in sticking with Java 8 anymore.

bmuschko:
Master is building with source and target compatibility 11 now. You can go ahead and work on a fix.

bmuschko:
I just fixed it: https://github.com/bmuschko/gradle-docker-plugin/commit/dcad57f2bad1be13a34bcfb9a3836858d8e6be8c

matteomazza91:
sorry for the delay on this.  
[...]

PR #997, opened 2021-03-19
Explicitly spell out fully qualified class references to inner classes Dockerfile.File and Dockerfile.CopyFile

matteomazza91:
additional doc changes for #990 
- fixed some groovydoc links
- fixed CopyFile doc


bmuschko (approved):
The references to `Dockerfile.File` were always correct. Given that `File` is an inner class, there was no need to spell out the pull path.

PR #1002, opened 2021-04-16
Upgrade ASM for JDK 16 support

### LINKED ISSUE #999: Support for JDK 16
The current version, 6.7.0, does not work with projects using JDK 16.

The task dockerCreateDockerFile fails with `Unsupported class file major version 60`

The issue seems to be the dependency, org.ow2.asm:asm, where this plugin uses version 7.3.1. When updating this to use the latest version, 9.1, the plugin appears to be working.

abrudin:
This PR upgrades asm to its latest version 9.1, in order to support JDK 16

closes #999


bmuschko (approved):

bmuschko:
Thanks, merged!

Issue #970, opened 2020-09-10
How to pass docker environment variables into java application?


vleushin:
Hello! As part of migrating from Java 12 to Java 14 I'm updating `gradle-docker-plugin` straight from `3.6.2` to `6.6.1`.

While overall everything seems OK, we can remove a lot of old quirks (like setting  and pushing `latest` tag) the only problem we currently have is that it is not clear how to pass docker environment variable into java application.

Old Dockerfile:
```
FROM openjdk:12
LABEL maintainer=vleushin
ADD myapp-node /myapp-node
ADD app-lib/myapp-node-2.18.14.jar /myapp-node/lib/myapp-node-2.18.14.jar
ENTRYPOINT ["/myapp-node/bin/myapp-node"]
EXPOSE 8080
```

New Dockerfile:
```
FROM openjdk:14
LABEL maintainer=vleushin
WORKDIR /app
COPY libs libs/
COPY resources resources/
COPY classes classes/
ENTRYPOINT ["java", "-cp", "/app/resources:/app/classes:/app/libs/*", "com.example.MyAppNode"]
EXPOSE 8080
```

As I understood, as part of optimisation and layering improvements -- there are new layers -- libs, resources, classes. That's all great. But the problem is `ENTRYPOINT`.

In old Dockerfile java application startup script has following line at the end:
```
# Collect all arguments for the java command, following the shell quoting and substitution rules
eval set -- $DEFAULT_JVM_OPTS $JAVA_OPTS $MYAPP_NODE_OPTS -classpath "\"$CLASSPATH\"" com.example.MyAppNode "$APP_ARGS"
```
So it is passing `DEFAULT_JVM_OPTS`, `JAVA_OPTS` and `MYAPP_NODE_OPTS` into java application. Which is great way to customize docker image to run on specific environment or with specific features. We use it heavily afterwards in Kubernetes -- when we specify in deployment image, and using environment variable to pass different java flags, DEV/PROD environment, etc.

Quick example of Kubernetes deployment:
```
      containers:
      - name: myapp-node
        image: myapp/myapp-node:2.18.20
        env:
        - name: MYAPP_ENV
          valueFrom:
            configMapKeyRef:
              name: myapp-config
              key: env
        - name: JAVA_OPTS
[...]

PR #1021, opened 2021-06-23
Enable putting environment variable as jvmArgs

fuminchao:
#### Problem:
Environment variables in jvmArgs are not working

Related issue https://github.com/bmuschko/gradle-docker-plugin/issues/970

```
// Environment variables in jvmArgs are not working
// E.g. docker run -e "JAVA_OPTS=xxx" images
docker {
    springBootApplication {
        baseImage = 'openjdk:8-alpine'
        ports = [9090, 8080]
        images = ['awesome-spring-boot:1.115', 'awesome-spring-boot:latest']
        jvmArgs = ['$JAVA_OPTS']
    }
}
```

#### Cause by:
The entrypoint for java plugins is "java ...jvmArgs -cp xxxx Main", which don't resolve the environment variables

#### Solution:
Change the entrypoint as ```sh -c "java ...jvmArgs -cp xxxx Main"```

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/DockerJavaApplicationPluginFunctionalTest.groovy
@@ -165,7 +165,7 @@ LABEL maintainer=benjamin.muschko@gmail.com
 WORKDIR /app
 COPY libs libs/
 COPY classes classes/
-ENTRYPOINT ["java", "-cp", "/app/resources:/app/classes:/app/libs/*", "com.bmuschko.gradle.docker.application.JettyMain"]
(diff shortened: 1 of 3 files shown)


bmuschko:
I don't think that's how I want to support JVM parameters necessarily. The JVM process would inherit everything from the environment which might cause unexpected side-effects.

fuminchao:
> I don't think that's how I want to support JVM parameters necessarily. The JVM process would inherit everything from the environment which might cause unexpected side-effects.

But as the issue https://github.com/bmuschko/gradle-docker-plugin/issues/970 mentioned, there is no way to pass jvm_options dynamically, according to different deploySpec

bmuschko:
> But as the issue #970 mentioned, there is no way to pass jvm_options dynamically, according to different deploySpec

Yes, but I think we'll want to the solve the problem in a different kind of fashion. I'd need to dig into myself a little bit. I will try to think about solutions sometime next week.

bmuschko:
I don't have time to look into this more deeply right now but I don't think we'll want to solve the problem as proposed.

PR #1024, opened 2021-06-28
Introduce platform parameter for build image task

bmuschko:


RicardoMonteiroSimoes:
@bmuschko as noted above, there appears to be 0 indication of where/how to use the new platform property.

bmuschko:
@RicardoMonteiroSimoes It's here: https://bmuschko.github.io/gradle-docker-plugin/current/api/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.html#platform

veda-s4dhak:
@bmuschko is there a way to specify the platform label in the build.gradle. Read the link you provided but its unclear if the indented way to specify platform is in the build.gradle or another way. 

docker {
	springBootApplication {
		// within here?
	}
}

bmuschko:
@veda-s4dhak You would define the platform in the build script but not with the help of the plugin's extension. You need to set the property for the task of type `DockerBuildImage`.

```
dockerBuildImage.platform = 'linux/s390x'
```

PR #1034, opened 2021-08-11
Issue 1029: Stop DockerLogsContainer with StopExecutionException

Bubesz:
This is my idea about how the plugin should handle GradleExceptions and particularly StopExecutionException. Throw it to Gradle.

Unfortunately, I was not able to test it in a meaningful way. A lot of other tests failed for me. I'm sorry about it, I'm not a groovy developer, not a Gradle plugin developer, I'm just a Java developer who was surprised by the way of exception handling.

The failing test is the reason why I wrote a tricky assertation part. I got NPE at the end of each test because `result.task(':logContainer')` was null for me, even in other tests. (I had to comment out tests so that I can verify mine at least runs.)

I really hope it will work (pass the tests) 🤞 but if not, I would like to require assistance! 🚑
Unfortunately, I don't have the time to learn groovy-based Gradle plugin development from the beginning. 😞


bmuschko:
Thanks for your pull request. I needed to make a bunch of changes so I replaced your PR with https://github.com/bmuschko/gradle-docker-plugin/pull/1083.

PR #1037, opened 2021-09-02
Add support for docker tmpfs mounts

### LINKED ISSUE #1036: Add support for docker tmpfs mounts
### Expected Behavior

It should be possible to configure a `tmpfs` filesystem (https://docs.docker.com/storage/tmpfs/) in the docker container using `CreateDockerContainer` task. The docker java api provides this functionality via [hostconfig.tmpFs](https://github.com/docker-java/docker-java/blob/28e32b0aca77ef71958eec3ad75212e1bd2dcfdf/docker-java-api/src/main/java/com/github/dockerjava/api/model/HostConfig.java#L281). A previous PR https://github.com/bmuschko/gradle-docker-plugin/pull/763 was rejected as full `--mount` support was requested for better future extensibility and maintenance. The full mount support is more involved to implement and from my point of view mount support can be an additional feature. The `--mount` api is more complex and should be possibly treated as separate distinguished feature.  

As part of the `CreateDockerContainer` it will be possible to add a tmpfs to the container like that:

```Groovy
    hostConfig.tmpFs = ['/testdata': 'rw,noexec,nosuid,size=2m']
}
```

### Current Behavior

It is not possible to add a `tmpfs` for the docker container 

### Context
A tmpfs inside the container is quite useful for testing and not having files persisted on the host as it is the case with mounted directories. In our scenario the docker container stores the files in a mounted directory with a different userid, which led to the problem that the build server cannot delete those files created in the docker container

jonnybbb:
This change add tmpfs support for the `CreateDockerContainer` task

closes #1036

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -82,6 +82,45 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
         build('inspectContainer')
     }
 
+    def "can setup and mount a tmpfs"() {
+        given:
+        String containerCreationTask = """
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImage()
+                hostConfig.tmpFs = ['/testdata': 'rw,noexec,nosuid,size=2m']
+                cmd = ['sleep','10']
+            }
+        """
+        buildFile <<
+            containerStart(containerCreationTask) <<
+            """
+            task execContainer(type: DockerExecContainer) {
+                dependsOn startContainer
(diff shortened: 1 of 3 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -82,6 +82,45 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
...
+        String containerCreationTask = """
+            task createContainer(type: DockerCreateContainer) {
+                dependsOn pullImage
+                targetImageId pullImage.getImage()
+                hostConfig.tmpFs = ['/testdata': 'rw,noexec,nosuid,size=2m']
+                cmd = ['sleep','10']
+            }
+        """
+        buildFile <<
+            containerStart(containerCreationTask) <<
```
Arzulla, line 96: tes

bmuschko:
Thanks for the PR. It has been merged.

Issue #1063, opened 2022-04-09
imageId.txt may not exists


ben-wangz:
### Expected Behavior
DockerBuildImage will write $buildDir/.docker/$taskpath-imageId.txt, but its parent directory may not exists

### Current Behavior
1. output of test-case
```text
➜  test-project ./gradlew testBuildImage
Step 1/2 : from alpine:3.15.0
 ---> 8e1d7573f448
Step 2/2 : CMD ["tail", "-f", "/etc/hosts"]
 ---> Using cache
 ---> 628949740b3b
Successfully built 628949740b3b
Successfully tagged test-image:1.0

> Configure project :
Building image using context '/Users/ben.wangz/temp/test-project'.
Using Dockerfile '/Users/ben.wangz/temp/test-project/Dockerfile'
Using images 'test-image:1.0'.

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/ben.wangz/temp/test-project/build.gradle' line: 18

* What went wrong:
A problem occurred evaluating root project 'test-project'.
> /Users/ben.wangz/temp/test-project/build/.docker/testBuild-imageId.txt (No such file or directory)
```

2. files of test case
    * tree
    ```text
    .
    ├── Dockerfile
    ├── build.gradle
    ├── gradle
     │    └── wrapper
     │        ├── gradle-wrapper.jar
     │        └── gradle-wrapper.properties
    └── gradlew
    ```
    * Dockerfile
    ```text
    from alpine:3.15.0
    
    CMD ["tail", "-f", "/etc/hosts"]
    ```
    * build.gradle
    ```text
        import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
    
    plugins {
        id "com.bmuschko.docker-remote-api" version "7.2.0"
    }
    
    task testBuildImage(){
        buildImage("test", "test-image:1.0")
    }
    
    def buildImage(String name, String imageName){
        Task buildImage = tasks.create("${name}Build", DockerBuildImage) {
            inputDir.set(project.file("."))
            dockerFile.set(project.file("Dockerfile"))
            images.add(imageName)
            remove.set(true)
        }
        buildImage.start()
    }
    ```

3. running test-case to reproduce
```shell
./gradlew testBuildImage
```

4. In addition, gradle: `distributionUrl=https\://services.gradle.org/distributions/gradle-7.1.1-bin.zip`

### Context
1. Why I need dynamically register DockerBuildImage
[...]

PR #1064, opened 2022-04-09
create parent of $buildDir/.docker/$taskpath-imageId.txt

ben-wangz:
to solve #1063 which explains the parent of `$buildDir/.docker/$taskpath-imageId.txt` may not exists


bmuschko:

PR #1065, opened 2022-04-19
Fix if-missing default auth lookup logic

### LINKED ISSUE #985: plugin overwrites auth details for https://index.docker.io/v1/ if no registryCredentials are defined
### Expected Behavior
Dockerhub credentials in `~/.docker/config.json` should be used when interacting with dockerhub without additional configuration.

### Current Behavior
Dockerhub credentials in `~/.docker/config.json` are read but discarded if there is no `registryCredentials` defined in the gradle build. This happens as the plugin generates a default `registryCredentials` configuration which has no username or password for dockerhub. This blank config then overwrites the config read from disk.

### Context
We thought we were authenticated against dockerhub, however we started getting rate limit errors which caused our builds to fail.

execute `docker login`
have a gradle project with no `registryCredentials` defined
execute `gradle dockerImage`

### Workaround
define fake registryCredentials

```
docker {
    registryCredentials {
        url = 'fake'
    }
}
```

I had to debug the plugin to discover the precise behaviour.

Sineaggi:
Fixes the issue where having credentials in the config.json are overwritten by default (but unset) auth-config passed in.

The javadoc of the function explicitly states
```
If missing, an AuthConfigurations object containing only the passed additionalAuthConfig is returned
```
But no if-missing logic was ever implemented. The `addConfig` function doesn't check to see if an existing registryAddress has already been registered, which causes it to stomp on the previously-loaded config.json loaded login details.

closes #985

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/internal/RegistryAuthLocator.groovy
@@ -243,7 +243,9 @@ class RegistryAuthLocator {
      */
     AuthConfigurations lookupAllAuthConfigs(AuthConfig additionalAuthConfig) {
         AuthConfigurations allAuthConfigs = lookupAllAuthConfigs()
-        allAuthConfigs.addConfig(additionalAuthConfig)
+        if (allAuthConfigs.configs.isEmpty()) {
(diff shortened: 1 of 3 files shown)


Sineaggi:
@bmuschko Any chance you'll be able to get to this pr?

bmuschko:
Thanks so much for the PR. It has been merged.

PR #1069, opened 2022-05-16
Support multiple images for DockerSaveImage (#1067)

### LINKED ISSUE #1067: Support multiple images for DockerSaveImage
### Expected Behavior
It would be nice if the `DockerSaveImage` task would support multiple images to be saved into the target tar.
This could for example also be used to export an image with two tags like `1.1.0` and `latest`.
The CLI is able to do that, as well as `docker-java`, it is just not exposed by this plugin.

### Current Behavior
You can only supply one image to the task

### Your Environment
Gradle 7.4.2, Plugin 7.3.0

Vampire:
Fixes #1067

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -56,6 +56,7 @@ abstract class AbstractFunctionalTest extends Specification {
             .withArguments(arguments + '-s' as List<String>)
             .withPluginClasspath()
             .withEnvironment(envVars)
+            .forwardOutput()
     }
 
     protected File file(String relativePath) {
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
@@ -1,23 +1,31 @@
 package com.bmuschko.gradle.docker.tasks.image
 
 import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
+import groovy.json.JsonSlurper
+import org.apache.commons.vfs2.VFS
+import spock.lang.Unroll
+
+import static org.gradle.testkit.runner.TaskOutcome.SKIPPED
 
 class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
 
-    private static final String IMAGE = 'alpine:3.4'
+    private static final String IMAGE_BASE = 'alpine'
+    private static final String IMAGE_3_2 = "$IMAGE_BASE:3.2"
+    private static final String IMAGE_3_3 = "$IMAGE_BASE:3.3"
+    private static final String IMAGE_3_4 = "$IMAGE_BASE:3.4"
     private static final String BUILD_DIR = 'build/docker'
-    private static final String CONTROL_SAVED_IMAGE = "$BUILD_DIR/alpine-3.4-docker-image-control.tar"
-    private static final String IMAGE_FILE = "$BUILD_DIR/alpine-3.4-docker-image.tar"
-    private static final String COMPRESSED_IMAGE_FILE = "$BUILD_DIR/alpine-3.4-compressed-docker-image.tar.gz"
+    private static final String CONTROL_SAVED_IMAGE = "$BUILD_DIR/alpine-docker-image-control.tar"
+    private static final String IMAGE_FILE = "$BUILD_DIR/alpine-docker-image.tar"
+    private static final String COMPRESSED_IMAGE_FILE = "$BUILD_DIR/alpine-compressed-docker-image.tar.gz"
 
     def "can save Docker image without compression"() {
-        buildFile << pullImageTask()
+        buildFile << pullImageTask('pullImage', IMAGE_3_4)
         buildFile << """
             import com.bmuschko.gradle.docker.tasks.image.DockerSaveImage
 
             task saveImage(type: DockerSaveImage) {
                 dependsOn pullImage
-                image = "${IMAGE}"
+                images.add("${IMAGE_3_4}")
                 destFile = file("${IMAGE_FILE}")
             }
         """
@@ -30,20 +38,20 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
     }
 
     def "can save Docker image with compression"() {
-        buildFile << pullImageTask()
+        buildFile << pullImageTask('pullImage', IMAGE_3_4)
         buildFile << """
             import com.bmuschko.gradle.docker.tasks.image.DockerSaveImage
 
             task saveImageControl(type: DockerSaveImage) {
                 dependsOn pullImage
-                image = "${IMAGE}"
+                images.add("${IMAGE_3_4}")
                 destFile = file("${CONTROL_SAVED_IMAGE}")
             }
 
             task saveImage(type: DockerSaveImage) {
                 dependsOn saveImageControl
                 useCompression = true
-                image = "${IMAGE}"
+                images.add("${IMAGE_3_4}")
                 destFile = file("${COMPRESSED_IMAGE_FILE}")
             }
         """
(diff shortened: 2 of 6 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -54,12 +62,130 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+            }
+        """
+
+        when:
+            build('saveImage')
+
+        then:
+            def repoTags = getRepoTags(IMAGE_FILE)
+            repoTags.containsAll(IMAGE_3_2, IMAGE_3_3, IMAGE_3_4)
+            repoTags.every { it.startsWith("${IMAGE_BASE}:") }
```
bmuschko, line 109: I feel like you already tested this with the previous assertion. What I would probably do it is see if the size of `repoTags` is 3 and if the collection contains the expected images.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -54,12 +62,130 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+
+            task saveImage(type: DockerSaveImage) {
+                dependsOn tasks.withType(DockerPullImage)
+                image = "${image}"
+                destFile = file("${IMAGE_FILE}")
+            }
+        """
+
+        when:
+            build('saveImage')
```
bmuschko, line 80: Can you align the indentation with `when`/`then`/`where` across the board? We don't add a tab to the next line in this code base.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
```diff
@@ -54,12 +62,130 @@ class DockerSaveImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
...
+
+        when:
+            def result = build('saveImage')
+
+        then:
+            result.task(":saveImage").outcome == SKIPPED
+            !new File(projectDir, IMAGE_FILE).exists()
+    }
+
+    def getRepoTags(imageFile) {
```
bmuschko, line 173: Would it have been easier/less brittle to use Docker Java here to inspect the repo tags?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
```diff
@@ -20,10 +24,21 @@ class DockerSaveImage extends AbstractDockerRemoteApiTask {
      * The image including repository, image name and tag to be saved e.g. {@code vieux/apache:2.0}.
      *
      * @since 6.0.0
+     * @deprecated Use {@link #images} instead
      */
     @Input
+    @Optional
+    @Deprecated(forRemoval = true)
```
bmuschko, line 31: The next release will likely be a major version upgrade. While it's usually better to deprecate first, I might just remove this field right away as a breaking change.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
```diff
@@ -36,17 +51,42 @@ class DockerSaveImage extends AbstractDockerRemoteApiTask {
...
-        InputStream image = saveImageCmd.exec()
+        Set<String> images = images.getOrElse([] as Set)
+        if (image.present) {
+            images = images + image.get()
+        }
+        // part of work-around for https://github.com/docker-java/docker-java/issues/1872
+        SaveImagesCmd saveImagesCmd = new SaveImagesCmdImpl(execution) {
+            @Override
+            List<TaggedImage> getImages() {
+                images.collect {
```
bmuschko, line 75: Would it work to use `images as List<TaggedImage>` instead?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
```diff
@@ -36,17 +51,42 @@ class DockerSaveImage extends AbstractDockerRemoteApiTask {
...
+                    { -> it } as TaggedImage
+                }
+            }
+        }
+        //SaveImagesCmd saveImagesCmd = dockerClient.saveImagesCmd()
```
bmuschko, line 80: Can you remove the commented out code?

bmuschko (changes requested):

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
Vampire: Of course, if you wish.
I just thought I leave it in, so once that mentioned bug is resolved, this can just be commented in and the work-around removed.

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
Vampire: No:
```
class java.lang.String cannot be cast to class com.github.dockerjava.api.command.SaveImagesCmd$TaggedImage (java.lang.String is in module java.base of loader 'bootstrap'; com.github.dockerjava.api.command.SaveImagesCmd$TaggedImage is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @3b15ef58)
```

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImage.groovy
Vampire: Sure, no problem, makes it slightly easier of course.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
Vampire: Can Docker Java look inside the exported image file?
If it only can ask the Docker daemon what I suspect, then it wouldn't help.
The question is not whether the tags are present in the repo, the check is whether the correct tags are exported to the image file.
Thus the content of the image file needs to be checked.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
Vampire: Whoops, sorry.
Usually I adapt to the surrounding style.
I'm just too used to it and my formatter settings do it automatically as it imho majorly improves readability. :-)

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
Vampire: Not really, that would make the test flaky.
If there are other alpine images in your local Docker daemon, they will be exported alongside as the command instructs to export all alpine tags.
So the last line ensures that as expected all exported tags are alpine tags and the line before that at least the three that are guaranteed by the test code are included, but there can be more.
Unless some code is added to delete all other alpine tags from the Docker daemon, but I didn't want to do that to anyone just running those tests.

Vampire:
Just in case you missed it, I addressed the comments and requested a re-review :-)

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
```diff
@@ -56,6 +56,7 @@ abstract class AbstractFunctionalTest extends Specification {
             .withArguments(arguments + '-s' as List<String>)
             .withPluginClasspath()
             .withEnvironment(envVars)
+            .forwardOutput()
```
bmuschko, line 59: This was probably only added temporarily for troubleshooting purposes. It usually produces a lot of output. I will remove it later.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerSaveImageFunctionalTest.groovy
bmuschko: Good point. I think Docker Java can't do this. Thanks for clarifying.

bmuschko:
Thanks, this has been merged.

src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
Vampire: Actually it was intended.
I was majorly confused by not getting any helpful output.
But as you like. :-)

PR #1072, opened 2022-05-21
Make build image task out-of-date if configured tags are missing

### LINKED ISSUE #1008: DockerBuildImage task doesn't re-run if the previous output still exists, but is no longer tagged
Create a project with a `DockerBuildImage` task, with the `images` property set to include at least one tag. Check out two copies of this project (gradle will try to rebuild each on its own, noticing changes, etc), with different contents in the `inputDir`. Build one project, then the second, then go back to the first and build its task again.

### Expected Behavior
Rebuilding the first project should result in the `DockerBuildImage` task running and tagging the resulting image (though the image was almost certainly still cached), so minimal actual work was done.

### Current Behavior
Currently, the DockerBuildImage notices that the `build/.docker/task-imageId.txt` exists, and that an image exists in docker with that ID, so it assumes that it is up to date. This means that the DockerBuildImage task does not run at all (if _within that gradle project_ there were no changes that influence the task). 

### Context
This is surprising in any case where you actually used the tag - for example, a downstream build image task in the same project that references that image by tag (with a `myTask.inputs.files(baseImage.outputs.files())` to correctly notice when the upstream image changed).

I could reasonably imagine that this is expected behavior, and that a case like this should instead be managed through DockerTagImage, but if that is the case, DockerBuildImage probably shouldn't offer to tag the image at all, since it will not reliably do so. An alternative could be to change the `upToDateWhen` for the DockerBuildImage task to instead/additionally invoke inspectImageCmd with each provided imageId, and verify that the resulting `InspectImageResponse`'s ID property, and confirm that it matches the id that was in the imageIdFile. Alternatively, check `repoTags` and ensure it contains all `image` values.

Sample projects which demonstrate the issue:
https://github.com/niloc132/docker-gradle-build-image-tag-bug/
Two ways to use this - check it out twice, once at `main` and once at `alt-branch`, or check out the https://github.com/niloc132/docker-gradle-build-image-tag-bug/tree/both-projects branch and move between `proj1/` and `proj2/` - the idea is the same. Both projects are identical, except for a file that ends up in the docker image that is built.

To reproduce with this setup, cd into one project, invoke `./gradlew showLogs`, and confirm it says "Hello, world" (main or proj1/), then switch to the other, and confirm it says "Greetings, planet". Then, switch back to the first - the expected result is that you see "Hello, world" again, but instead you incorrectly see "Greetings, planet".

Please note that this project is using version 6.7.0 of the plugin, though the code is the same in 7.0.0 - the project I'm working on is still using Java 8, so as I understand it, we will keep using 6.7 even if a fix is made.

---

### Proposed fix/workaround:
As discussed above, there are probably two reasonable ways to make this behave the way it looks like it should:
 * Remove the `images` field from DockerBuildImage, if it isn't expected that this will be built and the tag applied
 * Modify DockerBuildImage's `upToDateWhen` to iterate over the given `images` values and perform a `inspectImageCmd` on each one, confirming that the resulting `id` field matches the current `fileImageId`, as currently read in the current task.

The first option would mean adding a new task whenever an image is expected to be tagged - not the end of the world, but perhaps surprising to users. The second could look like this as a workaround in any DockerBuildImage task, or adapted to be part of the existing upToDateWhen closure:
```
    outputs.upToDateWhen {
        File file = imageIdFile.get().asFile
        if(file.exists()) {
            try {
                // get the last imageId we used
                def fileImageId = file.text
                // check if that image still exists
                for (String image : images.get()) {
                    def inspect = getDockerClient().inspectImageCmd(image).exec();
                    // see if that image is tagged the way we expectif not, re-run
                    def sha = inspect.id.substring("sha:256".length());
                    if (sha != fileImageId && !sha.startsWith(fileImageId)) {
                        return false;
                    }
                }
                return true;
            } catch (DockerException e) {
                // if we fail, it must not have existed, re-run
                return false
            }
        }
        return false
    }

```

Vampire:
Fixes #1008

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImageFunctionalTest.groovy
@@ -58,26 +58,26 @@ USER \$user"""
                 dockerFile = file('Dockerfile')
                 buildArgs = ['user': 'what_user']
             }
-            
+
             task inspectImage(type: DockerInspectImageUser) {
                 dependsOn buildImage
                 imageId = buildImage.imageId
             }
-            
+
             task removeImage(type: DockerRemoveImage) {
                 force = true
                 imageId = buildImage.imageId
             }
 
             inspectImage.finalizedBy tasks.removeImage
-            
+
             class DockerInspectImageUser extends DockerExistingImage {
                 DockerInspectImageUser() {
                     onNext({ image ->
                         project.logger.quiet "user: \$image.containerConfig.user"
                     })
                 }
-            
+
                 @Override
                 void runRemoteCommand() {
                     def image = dockerClient.inspectImageCmd(imageId.get()).exec()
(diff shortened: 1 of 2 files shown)


bmuschko:
I am just getting back to this. LGTM! Sounds like your PR would address the problem in https://github.com/bmuschko/gradle-docker-plugin/issues/1008 AFAIKT. Can you confirm?

Vampire:
I think so too, yes

bmuschko:
Thanks for the pull request. This has been merged.

PR #1078, opened 2022-08-02
Configure test tasks to use JUnit Platform which is necessary for Spock 2+

Vampire:
When updating to Spock 2 you disabled each and every test.
You probably missed the important information in the release notes,
that Spock 2+ is now based on JUnit Platform, so it is necessary
to tell Gradle to not use JUnit 4, but JUnit Platform.


bmuschko:
Thanks, I noticed that as well. It has already been fixed on the `master` branch.

Vampire:
Ah, sorry, updated day before yesterday and didn't notice there was another push, after brooding a day over why the hell tests do not work anymore for me. :-D

bmuschko:
It took me a while too after getting back to it. I would have let you know long-term but was working on finishing another PR right now.

PR #1082, opened 2022-08-06
Use Java toolchain

bmuschko:
Standardize JDK across developer machines and CI agents.


bmuschko:
From the Gradle documentation:

> Toolchain support is available in the Java plugins and for the tasks they define. For the Groovy plugin, compilation is supported but not yet Groovydoc generation. For the Scala plugin, compilation and Scaladoc generation are supported.

PR #1088, opened 2022-08-13
Tweak DockerCreateContainerFunctionalTest

denuno:
Did not create an issue for this, as it's just a bit of cleanup basically, but it fixes the failing tmpfs test

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
@@ -118,7 +118,7 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
 
         expect:
         def result = build('inspectContainer')
-        result.output.contains("tmpfs /testdata tmpfs rw,nosuid,nodev,noexec,relatime,size=2048k 0 0")
+        result.output.contains("tmpfs /testdata tmpfs rw,nosuid,nodev,noexec,relatime,size=2048k,inode64 0 0")
     }
 
     def "can override default MAC address"() {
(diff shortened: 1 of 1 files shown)


denuno:
It might be more future-proof to just truncate that string before the next segment (so cut off the ` 0 0` at the end) but YMMV

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainerFunctionalTest.groovy
```diff
@@ -118,7 +118,7 @@ class DockerCreateContainerFunctionalTest extends AbstractGroovyDslFunctionalTes
 
         expect:
         def result = build('inspectContainer')
-        result.output.contains("tmpfs /testdata tmpfs rw,nosuid,nodev,noexec,relatime,size=2048k 0 0")
+        result.output.contains("tmpfs /testdata tmpfs rw,nosuid,nodev,noexec,relatime,size=2048k,inode64 0 0")
```
bmuschko, line 121: Can you explain why this is needed? The test is passing right now on the master branch?

denuno: It's needed because the output is different, but apparently we'll get different values for it depending on the host system, so it's not a very good fix.

If the order is always the same, then my "future-proofing" idea should work — just truncating it before it gets to the next section of output, basically right after the size — but if the order is not reliable maybe just looking for specific output, like the size, would be better.

I don't know if it listed the specific difference when it failed on the server the first time, but that's the difference I saw when I ran it locally.

bmuschko: Honestly, I think checking the console output doesn't make sense and is brittle. The `inspectContainer` task already verifies the relevant configuration.

```
onNext { info ->
    if (info.hostConfig.tmpFs['/testdata'] != 'rw,noexec,nosuid,size=2m') {
        throw new GradleException("'testdata' tmpfs not found")
    }
}
```

I'd say, let's remove the `result.output.contains` line.

PR #1094, opened 2022-09-09
Cleanup close calls

Sineaggi:
Not sure if this needs a related issue, but there are a few calls to `close()` methods that should no longer be necessary since the groovy 3 try-with-resources changes.

Removes now-unused `closeQuietly()` method.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/internal/IOUtils.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/internal/IOUtils.groovy
@@ -22,16 +22,6 @@ final class IOUtils {
         }
     }
 
-    static void closeQuietly(Closeable toClose) {
-        try {
-            if (toClose != null) {
-                toClose.close()
-            }
-        } catch (IOException ignored) {
-            // ignore
-        }
-    }
-
     /**
      * Create a progress logger for an arbitrary project and class.
      *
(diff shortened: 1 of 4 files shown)


bmuschko:
This change would make the plugin incompatible automatically with all Gradle versions that still use Groovy 2. I'd like to avoid making the change if I can for backward-compatibility reasons. I am not sure which version is really the last compatible Gradle version as I am not testing against different Gradle version. It would take far too long to run those tests.

Sineaggi:
With the release of 8.0, the user guide https://bmuschko.github.io/gradle-docker-plugin/current/user-guide/ says
> This plugin requires Gradle >= 7.4.0 to work properly.

and gradle >= 7.0 has a baseline of groovy 3.

Sineaggi:
I've added an alternate commit that uses the `withClosable(Closure)` which is groovy 2 compatible.

bmuschko:
You can only really make a determination about compatibility for the Gradle version used to build the plugin. All other versions can potentially have backward- or forward-compatibility issues if you don't test for it aka running all test cases with the Gradle distribution you are planning to support. That's why the user guide of the plugin makes this statement. Long-term, I would like to change all production source code of the plugin to Java to element the Groovy variable here.

bmuschko:
The CI build is currently failing. I believe this is due to scoping of a method call from a Closure. Honestly, I don't quite see the benefit for making the changes in this PR given that I'd like to keep the code as Java-esque as a possible to easy a transition. Let's close this PR. Thanks again.

Sineaggi:
> Long-term, I would like to change all production source code of the plugin to Java to element the Groovy variable here.
> Honestly, I don't quite see the benefit for making the changes in this PR given that I'd like to keep the code as Java-esque as a possible to easy a transition.

This seems reasonable, in that case we could still merge the first commit as it would be closer to the Java way to auto-close reasource.

Issue #954, opened 2020-07-22
Docker Java dependency clashes with Testcontainers


frankkoornstra:
When I combine this plugin with [Testcontainers](https://www.testcontainers.org/) to test the built image, Testcontainers runs into an NPE because of deserialization issues. Testcontainer's author has [tracked it down to](https://testcontainers.slack.com/archives/C1SUBPZK6/p1595329713276000?thread_ts=1595256124.253800&cid=C1SUBPZK6) the following:
>it looks like gradle-docker-plugin “embeds” docker-java, both api and core, and then shades Jackson [...] you should report to them because it breaks any project that uses docker-java

### Expected Behavior
Be able to combine this plugin with any other plugin that relies on Docker Java and Jackson

### Current Behavior
Deserialization starts to fail. I've created a [reproducer](https://github.com/frankkoornstra/testcontainers-reproducer) that shows the issue.

### Context
(none)

### Steps to Reproduce (for bugs)
Take a look at the [reproducer](https://github.com/frankkoornstra/testcontainers-reproducer)

### Your Environment
Ran the reproducer on several systems, all with the same result

bmuschko:
Thanks for reporting. Are you interested in sending a pull request?

frankkoornstra:
To be honest: I don't have that much experience yet with the dependency system in the JVM world yet so I don't know if I can fix it. I can only understand the suggestions that were given in the Slack chat partly and was already happy that I can show what was wrong :D

bmuschko:
@frankkoornstra Why do you define the plugin as a dependency for the project source code and not for the build script? Can you explain what you are trying to achieve? Are you trying to build a Gradle plugin and then reference a class of the plugin from your plugin code? https://github.com/frankkoornstra/testcontainers-reproducer/blob/master/build.gradle.kts#L14

frankkoornstra:
I'm building a Gradle plugin (let's call it `pA`) and in there applying your plugin (`pB`) to the project that `pA` is applied to. That way `pA` can define new tasks based on `pB` tasks in the project `pA` is applied to.

This might be a bit of an edge case and I already found another solution: simply using the Docker client because it was less hassle than I thought it would be.

bmuschko:
[...]

PR #1095, opened 2022-09-13
Add smoke tests support

Sineaggi:
This is based off of the `integrationTest.gradle` file in @michel-kraemer's [gradle-download-task](https://github.com/michel-kraemer/gradle-download-task/blob/master/gradle/integrationTest.gradle).

In order to validate against multiple versions of gradle, this integration test will create a project directory with a simple source file and build the docker image. I'm not sure about running these tests in CI, as they would require a docker socket but at least they can be run locally to validate the plugin works on multiple gradle versions.

I've also removed shading as it's broken when attempting to publish locally (and has been causing other issues such as https://github.com/bmuschko/gradle-docker-plugin/issues/954)


michel-kraemer:
@Sineaggi Thanks for referring to my repo.

Just a quick note from my side: Be aware that Gradle also offers a more modern (and more idiomatic) approach to testing plugins against various Gradle versions:
https://docs.gradle.org/current/userguide/testing_gradle_plugins.html

We've made use of this in gradle-download-task here:
https://github.com/michel-kraemer/gradle-download-task/blob/master/src/test/java/de/undercouch/gradle/tasks/download/FunctionalTestBase.java

bmuschko:
We'd definitely would want to go with the approach described by @michel-kraemer and not with a custom solution. Keep in mind that we'd also have run _all_ test cases with all supported Gradle version to make a definitive claim that the plugin is compatible with a given Gradle version. Doing so would highly increase the build-time for any change to the plugin which I am not willing to swallow right now. I am going to close this PR for now. Thanks for looking into this but the proper solution would be more involved.

Sineaggi:
I wasn't aware the `GradleRunner` api supported running with different versions of gradle, that's significantly simpler than the approach I was trying to replicate.

> Doing so would highly increase the build-time for any change to the plugin which I am not willing to swallow right now.

I would imagine it should be possible to do something similar to what @michel-kraemer does where each ci runs the tests with a specific subset of versions right?

PR #1108, opened 2022-10-20
Use build service for Docker client to better support Gradle's configuration cache feature

Sineaggi:
The main fix here is the removal of the buildFinished call
```
- Unknown location: registration of listener on 'Gradle.buildFinished' is unsupported
  See https://docs.gradle.org/7.4.2/userguide/configuration_cache.html#config_cache:requirements:build_listeners
```
Now we register a build service that implements AutoClosable. This build service contains the same code that previously existed in the base abstract class moved to its own class. I also implemented a simple 'caching' system where already requested clients will simply be cached in a ConcurrentHashMap. I expect most users will have one docker client configuration. The caching system was implemented to mimic the `@Memoized` used previously.

We had to remove the `@Memoized` annotations since those have issues with gradle's build cache.
```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':buildImage'.
...
Caused by: java.lang.ClassCastException: class java.lang.Object cannot be cast to class com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask (java.lang.Object is in module java.base of loader 'bootstrap'; com.bmuschko.gradle.docker.tasks.AbstractDockerRemoteApiTask is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @f540b62)
```

The upToDateWhen was changed from a Closure to a Spec to avoid an issue where gradle's configuration cache would capture too much of the surrounding scope.
```
* Exception is:
...
Caused by: org.codehaus.groovy.runtime.typehandling.GroovyCastException: Cannot cast object 'java.lang.Object@7fed6142' with class 'java.lang.Object' to class 'com.bmuschko.gradle.docker.tasks.image.DockerBuildImage'
...
```

The `imageId` property was changed from Internal + setter to based off the output file. I'm not sure if this is the best fix, but task outputs can only be files so basing the value of `imageId` off of the value of the output file seemed the most sane.
```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':assertImageIdForBuildImage'.
...
Caused by: org.gradle.api.GradleException: The imageId property was not set from task buildImage
...
```

To fix the imageIdValidation task, which had this issue
```
1 problem was found storing the configuration cache.
- Task `:assertImageIdForBuildImage` of type `org.gradle.api.DefaultTask`: cannot serialize object of type 'com.bmuschko.gradle.docker.tasks.image.DockerBuildImage', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.4.2/userguide/configuration_cache.html#config_cache:requirements:task_access
```
I've replaced the calls to dependantTask with the values outside of the doLast closure.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
+++ b/src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
@@ -1,0 +1,108 @@
+package com.bmuschko.gradle.docker.services;
+
+import com.bmuschko.gradle.docker.tasks.DockerClientConfiguration;
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.core.DefaultDockerClientConfig;
+import com.github.dockerjava.core.DockerClientImpl;
+import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
+import com.google.common.io.Closer;
+import org.gradle.api.file.Directory;
+import org.gradle.api.file.DirectoryProperty;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.services.BuildService;
+import org.gradle.api.services.BuildServiceParameters;
+
+import javax.inject.Inject;
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public abstract class DockerClientService implements BuildService<DockerClientService.Params>, AutoCloseable {
+    private final Map<DefaultDockerClientConfig, DockerClient> dockerClients;
+
+    private final ObjectFactory objects;
+
+    public interface Params extends BuildServiceParameters {
+        Property<String> getUrl();
+        DirectoryProperty getCertPath();
+        Property<String> getApiVersion();
+    }
+
+    @Inject
+    public DockerClientService(ObjectFactory objects) {
+        this.objects = objects;
+        dockerClients = new ConcurrentHashMap<>();
+    }
+
+    public DockerClient getDockerClient(DockerClientConfiguration dockerClientConfiguration) {
+        String dockerUrl = getDockerHostUrl(dockerClientConfiguration);
+        File dockerCertPath = thingOrProperty(objects.directoryProperty(), dockerClientConfiguration.getCertPath(), getParameters().getCertPath()).map(Directory::getAsFile).getOrNull();
+        String apiVersion = thingOrProperty(objects.property(String.class), dockerClientConfiguration.getApiVersion(), getParameters().getApiVersion()).getOrNull();
+
+        // Create configuration
+        DefaultDockerClientConfig.Builder dockerClientConfigBuilder = DefaultDockerClientConfig.createDefaultConfigBuilder();
+        dockerClientConfigBuilder.withDockerHost(dockerUrl);
+
+        if (dockerCertPath != null) {
+            String caninicalCertPath;
+            try {
+                caninicalCertPath = dockerCertPath.getCanonicalPath();
+            } catch (IOException e) {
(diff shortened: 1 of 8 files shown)


src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
```diff
@@ -221,18 +226,31 @@ class DockerBuildImage extends AbstractDockerRemoteApiTask implements RegistryCr
         quiet.set(false)
         pull.set(false)
         cacheFrom.empty()
+
+        imageId.set(imageIdFile.map { RegularFile it ->
```
bmuschko, line 230: We'll want to have similar handling in DockerCommitImage I'd think. Can you add that?

bmuschko (changes requested):
Looks pretty good. Thanks! Could you have a look at the one comment I had for the change?

src/main/groovy/com/bmuschko/gradle/docker/tasks/image/DockerBuildImage.groovy
Sineaggi: I've added a fix for DockerCommitImage

Sineaggi: DockerCreateImage is also similar

Sineaggi:
Failed build appears to be a due to a repository blip.

src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
```diff
@@ -0,0 +1,108 @@
+package com.bmuschko.gradle.docker.services;
+
+import com.bmuschko.gradle.docker.tasks.DockerClientConfiguration;
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.core.DefaultDockerClientConfig;
+import com.github.dockerjava.core.DockerClientImpl;
+import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
+import com.google.common.io.Closer;
```
bmuschko, line 8: Can you remove the explicit dependency on Guava and handle this differently?

src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
```diff
@@ -0,0 +1,108 @@
+package com.bmuschko.gradle.docker.services;
```
bmuschko, line 1: We'll want to give an indication that it is not meant to be part of the plugin's public API. Can you move this into `com.bmuschko.gradle.docker.internal.services`?

bmuschko (changes requested):

bmuschko:
The CI build is passing now. The plugin portal had an issue over the weekend. Can you address the two comments I just added? After that we are ready to merge. Thanks again for your work.

src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
Sineaggi: Removed and replaced with something similar.

src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
Sineaggi: Moved

bmuschko:
Looks like you missed changing the imports. CI is reporting a compilation error.

src/main/groovy/com/bmuschko/gradle/docker/services/DockerClientService.java
Sineaggi: > Looks like you missed changing the imports. CI is reporting a compilation error.

I was wondering, tests were running locally and not in CI. Fixed

bmuschko:
Thanks a lot. It has been merged.

Issue #1109, opened 2022-10-27
Support auto-detection DOCKER_HOST on macOS Ventura with latest Docker Desktop


donbeave:
Most of my builds suddenly failed on the latest macOS version with the latest Docker Desktop app, the error is:

```
com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.HttpHostConnectException: Connect to http://127.0.0.1:2375 [/127.0.0.1] failed: Connection refused
```

The issue is because the unix socket file location was changed to user dir, in my case it is: `unix:///Users/donbeave/.docker/run/docker.sock`.

It can be fixed with the fix:

```
export DOCKER_HOST="unix://${HOME}/.docker/run/docker.sock"
```

But I think it's better to do such checking inside the plugin directly and autodetect the proper location.

### Your Environment

Apple M1 Max
macOS Ventura 13.0 (22A380)
Docker Desktop 4.13.0 (89412) for ARM64

bmuschko:
Thanks for reporting. The logic that would have to be changed is [here](https://github.com/bmuschko/gradle-docker-plugin/blob/master/src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy#L94-L110). Would you be interested in providing a pull request for this?

gesellix:
This seems to be related with https://github.com/docker/for-mac/issues/6529. Please note that the change has been rolled back with [Docker Desktop 4.13.1](https://docs.docker.com/desktop/release-notes/#docker-desktop-4131) and they are going to retry with a better upgrade path and docs.

bmuschko:
@gesellix Thanks for the pointer.
@donbeave Despite the fix in Docker Desktop I think it still makes sense to integrate your changes. I am going to have another look at your PR soon.

PR #1111, opened 2022-10-29
Support auto-detection DOCKER_HOST on macOS Ventura with latest Docker Desktop

donbeave:
This is a fix for the problem described in this issue: https://github.com/bmuschko/gradle-docker-plugin/issues/1109.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
```
bmuschko, line 3: Let's not bump up the version of Gradle ensure better backward compatibility. Can remove the Wrapper update from the change?

src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
```diff
@@ -97,6 +97,8 @@ class DockerExtension {
             boolean isWindows = isWindows()
             if (!isWindows && new File('/var/run/docker.sock').exists()) {
                 dockerUrl = 'unix:///var/run/docker.sock'
+            } else if (!isWindows && new File("${System.getProperty("user.home")}/.docker/run/docker.sock").exists()) {
```
bmuschko, line 100: I'd have a nested if conditional here. That'd make it a bit more readable I think.

Something like:

```
if (!isWindows) {
   if(new File('/var/run/docker.sock').exists()) {
      // ...
   }
   else if(new File("${System.getProperty("user.home")}/.docker/run/docker.sock").exists()) {
       // ...
   }
}
```

Long term, I think we should break out this logic into a different class to make it testable. I'd really want some basic unit tests for this logic. Do you think you could do that? If not, I can work on it later.

bmuschko (changes requested):

gradle/wrapper/gradle-wrapper.properties
donbeave: Ok, I'll revert it back.

src/main/groovy/com/bmuschko/gradle/docker/DockerExtension.groovy
donbeave: I can move this out and cover it with some tests.

donbeave:
@bmuschko I did some refactoring and added some unit tests for this logic, I hope you like this approach, in case you need to change something please let me know.

src/test/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolverTest.groovy
```diff
@@ -0,0 +1,115 @@
...
+    private static String fallbackDockerUrl = 'tcp://127.0.0.1:2375'
+
+    @TempDir
+    Path temporaryFolder
+
+    void setup() throws IOException {
+        SystemConfig.clear()
+
+        // override system envs with null values, to avoid flaky side effects
+        SystemConfig.setEnv('DOCKER_HOST', null)
```
bmuschko, line 21: The relevant test methods should use the Spock annotation `@RestoreSystemProperties` which will restore the original values after a test.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
```diff
@@ -0,0 +1,66 @@
+package com.bmuschko.gradle.docker.internal
+
+import com.google.common.annotations.VisibleForTesting
```
bmuschko, line 3: Let's not explicitly use a Guava class. It may cause issues down the road as versions are often times breaking + this is just a transitive dependency of something else. We didn't declare Guava to be use in the plugin.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
```diff
@@ -0,0 +1,66 @@
...
+import groovy.transform.CompileStatic
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.Logging
+
+import javax.annotation.Nullable
+
+import static com.bmuschko.gradle.docker.internal.OsUtils.isWindows
+
+@CompileStatic
+class DefaultDockerConfigResolver {
```
bmuschko, line 13: This should implement an interface so that it becomes mockable. Please remove the usable of `static` methods.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
```diff
@@ -0,0 +1,66 @@
...
+    static String getDefaultDockerUrl() {
+        getDefaultDockerUrl(
+                { new File('\\\\.\\pipe\\docker_engine').exists() },
+                { new File('/var/run/docker.sock').exists() },
+                { new File("${SystemConfig.getProperty("user.home")}/.docker/run/docker.sock").exists() }
+        )
+    }
+
+    @VisibleForTesting
+    protected static String getDefaultDockerUrl(Closure<Boolean> winPipeDockerEngineExists,
```
bmuschko, line 26: Can you explain why those parameters are Closures? Why not just use Boolean values?

src/main/groovy/com/bmuschko/gradle/docker/internal/SystemConfig.groovy
```diff
@@ -0,0 +1,50 @@
...
+import groovy.transform.CompileStatic
+
+/**
+ * The internal abstraction on top of JVM class {@link System}. The main purpose is to provide a way to manipulate
+ * these properties and environments in the unit test environment.
+ */
+@CompileStatic
+class SystemConfig {
+
+    private static Map<String, String> OVERRIDE_PROPERTIES = [:]
```
bmuschko, line 13: I think this internal "cache" makes things unnecessarily brittle. I don't think there's a big impact to querying the info multiple times. Please also hidden this behind an interface and make it non-static.

bmuschko (changes requested):

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
donbeave: Can you give a little bit snippet of how you see it?

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
donbeave: Originally it was because don't want to make an IO call to verify are all of these files are exists or not, even if this check is not used in that particular case, with closure I can check file existence only if the method goes into that scope.

donbeave: I think maybe move these check files existing calls to dedicated methods, and these methods can be overridden inside the unit tests. I'll add these changes in a few minutes.

src/main/groovy/com/bmuschko/gradle/docker/internal/SystemConfig.groovy
donbeave: Removed `OVERRIDE_PROPERTIES`, but still keep `OVERRIDE_ENVS`, as there is no similar functionality for system environments. I think `SystemConfig` can be an interface with default implementation which use System class to get system environments. For unit tests, I can pass my own implementation which can override some envs. What do you think?

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
donbeave: Done.

donbeave:
@bmuschko I did some small refactoring and completely removed SystemConfig class, please check the current implementation.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
bmuschko: Something like:

```
interface DockerConfigResolver {
   String getDefaultDockerUrl()
    ...
}

class DefaultDockerConfigResolver implements DockerConfigResolver {
    @Override
    String getDefaultDockerUrl() {
        ...
    }
    ...
}
```

src/test/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolverTest.groovy
bmuschko: My bet. You are right. Looks like you remove the implementation.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
```diff
@@ -0,0 +1,63 @@
...
+            File certFile = new File(dockerCertPath)
+            if (certFile.exists()) {
+                return certFile
+            }
+        }
+        return null
+    }
+
+    @Nullable
+    protected String getEnv(String name) {
```
bmuschko, line 55: I am not sure it's worth exposing those two methods here: getEnv and isFileExists. I'd rather not make it part of the API of this class.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
donbeave: Hmm, not very understand what it can bring to me, as I need to test the body of the `getDefaultDockerUrl()` method.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
donbeave: What's your suggestion? I can see another way is back to some abstraction around `System.getenv()` and File utilities, like an interface, with methods:


```java
interface SystemSettings {
   @Nullable
   String getEnv(String name)
   
   boolean isFileExists(String path)
   
   // etc
}
```

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
bmuschko: It will help with testing where the class is being used, in the extension. I can touch that up after merging the PR.

src/main/groovy/com/bmuschko/gradle/docker/internal/DefaultDockerConfigResolver.groovy
bmuschko: Basically, like a class with no constructor and static methods. I'll touch it up after merging the PR.

bmuschko (approved):
Thanks for the PR.

bmuschko:
The unit tests are currently failing. Can you add the relevant Spock annotations to the method that should only run on specific operating systems?

For example: `@Requires({ os.windows })`

bmuschko (changes requested):
Reopened PR as unit tests are failing.

donbeave:
This test fixed:

```
DefaultDockerConfigResolverTest > returns default fallback docker url for all operation systems: #osName > com.bmuschko.gradle.docker.internal.DefaultDockerConfigResolverTest.returns default fallback docker url for all operation systems: Windows 10 FAILED
[59](https://github.com/bmuschko/gradle-docker-plugin/pull/1111/checks?sha=41b488b2e839c7b7a6e7390dc74bdc7d52299c4f#step:5:60)
    org.spockframework.runtime.SpockComparisonFailure at DefaultDockerConfigResolverTest.groovy:38
```

donbeave:
@bmuschko build fixed

bmuschko (approved):

bmuschko:
Thanks! This change has been merged.

bmuschko:
I see the following issue when running some of the functional tests with Docker Desktop 4.13.1 (90346).

```
com.github.dockerjava.api.exception.InternalServerErrorException: Status 500: unknown image in /images/create?fromImage=alpine%3A3.4

	at com.github.dockerjava.core.DefaultInvocationBuilder.execute(DefaultInvocationBuilder.java:247)
	at com.github.dockerjava.core.DefaultInvocationBuilder.lambda$executeAndStream$1(DefaultInvocationBuilder.java:269)
```

Just checking to see if you tried running those tests on your local machine. I am wondering if this is an issue with the latest Docker Desktop version or something we changed recently. Would you mind checking?

gesellix:
This looks like https://github.com/docker/for-mac/issues/6540 :)

bmuschko:
@gesellix Thanks for the pointer. Super helpful!

PR #1113, opened 2022-10-31
Enable configuration cache for all functional tests

Sineaggi:
Mostly the same fixes we've seen before.

Fixed DockerCreateContainer to have the same file backing code as in https://github.com/bmuschko/gradle-docker-plugin/pull/1108/files#diff-be81629192d1cfcf7ebdbfb002ee9f04492909eb2511895b4d39e9eb2fc8d640.

Exposed port needs Serializable.

DockerExecContainer needs a logger which we can get from `getServices`, avoiding the call to project.

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -54,7 +54,7 @@ abstract class AbstractFunctionalTest extends Specification {
     private GradleRunner createAndConfigureGradleRunner(String... arguments) {
         GradleRunner.create()
             .withProjectDir(projectDir)
-            .withArguments(arguments + '-s' as List<String>)
+            .withArguments(arguments + '-s' + CONFIGURATION_CACHE as List<String>)
             .withPluginClasspath()
             .withEnvironment(envVars)
     }
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/AbstractFunctionalTest.groovy
@@ -12,7 +12,6 @@ abstract class AbstractFunctionalTest extends Specification {
     public static final String TEST_IMAGE = 'alpine'
     public static final String TEST_IMAGE_TAG = '3.4'
     public static final String TEST_IMAGE_WITH_TAG = "${TEST_IMAGE}:${TEST_IMAGE_TAG}"
-    public static final String CONFIGURATION_CACHE = '--configuration-cache'
 
     @TempDir
     File temporaryFolder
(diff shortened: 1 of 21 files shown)


Sineaggi:
For the classes that aren't serializable, I'm seeing this exception
```
Caused by: java.io.NotSerializableException: com.bmuschko.gradle.docker.tasks.container.DockerCreateContainer$HostConfig$LogConfig |  
-- | --
  | at org.gradle.internal.snapshot.impl.AbstractValueProcessor.javaSerialization(AbstractValueProcessor.java:171) |  
  | ... 100 more
```

Sineaggi:
Fix copy containers call to project at task time
```
- Task `:copyFileIntoContainer` of type `com.bmuschko.gradle.docker.tasks.container.DockerCopyFileToContainer`: invocation of 'Task.project' at execution time is unsupported.
```

Sineaggi:
Not sure how to fix the `getApplicationMainClassName` function, it needs the outputs of the compile task before they're ready.

```
1 problem was found storing the configuration cache.
- Task `:dockerCreateDockerfile` of type `com.bmuschko.gradle.docker.tasks.image.Dockerfile`: value 'provider(?)' failed to unpack provider
```

Current workaround is to manually specify the `mainClassName`.
```
docker {
    javaApplication {
        mainClassName = 'com.bmuschko.gradle.docker.application.JettyMain'
    }
}
```

Sineaggi:
All functional tests now have `--configuration-cache` enabled.

Sineaggi:
I had to bump the spring boot gradle plugin version because older versions didn't support configuration cache.

Sineaggi:
Values in maps like
```
labels = ["project.name": "\$project.name"]
```
should be replaced with
```
labels = ["project.name": project.name]
```
as otherwise you'd see cast errors.
```
* What went wrong:
Execution failed for task ':createContainer'.
> com.fasterxml.jackson.databind.JsonMappingException: class org.codehaus.groovy.runtime.GStringImpl cannot be cast to class java.lang.String (org.codehaus.groovy.runtime.GStringImpl is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @685cb137; java.lang.String is in module java.base of loader 'bootstrap') (through reference chain: com.github.dockerjava.core.command.CreateContainerCmdImpl["Labels"]->com.google.common.collect.SingletonImmutableBiMap["project.name"])
```

Sineaggi:
I don't know how to fix the `Writer sink` issue
```
2 problems were found storing the configuration cache, 1 of which seems unique.
- Task `:logContainer` of type `com.bmuschko.gradle.docker.tasks.container.DockerLogsContainer`: cannot serialize object of type 'java.io.FileOutputStream', a subtype of 'java.io.OutputStream', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.4.2/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```
Not sure how this can be evolved in the future, so instead if a sink is set, configuration cache is disabled for the task.

bmuschko (approved):
Awesome work. Thanks!

PR #1114, opened 2022-11-01
Convert instructions to only store providers in Dockerfile task

Sineaggi:
This is something I came across while working on the `Dockerfile` task. Internally each instruction has either `T` or `Provider<T>`. This means each instruction will have fields for both. To make the code cleaner and simplify implementation, I've replaced the raw `T` field with a provider constructed with the `Providers.ofNullable` method.


bmuschko (approved):
Perfect, that makes the code much simpler.

PR #1116, opened 2022-11-06
Replace copy function with transferTo

Sineaggi:
Since we're on java 11 already, let's use the new `transferTo` method instead of continuing to use the custom copy.


bmuschko:
Thanks. This has been merged.

PR #1117, opened 2022-11-06
Convert classes to java

Sineaggi:
I've started the conversion with a set of classes that were pretty easy to convert, using Intellij's automatic conversion from groovy to java. This is a good first step as it'll make sure all semicolons are inserted and all access modifiers are correct.

I've split the rewrite into two commits, the first to make it java-valid groovy, then the final rename step. That way as much history in the files is retained as possible.


src/main/groovy/com/bmuschko/gradle/docker/DockerSpringBootApplicationPlugin.groovy
```diff
@@ -1,34 +0,0 @@
...
-
-/**
- * Opinionated Gradle plugin for creating and pushing a Docker image for a Spring Boot application.
- * <p>
- * This plugin can be configured with the help of {@link DockerSpringBootApplication}.
- *
- * @since 3.4.5
- */
-@CompileStatic
-class DockerSpringBootApplicationPlugin extends DockerConventionJvmApplicationPlugin<DockerSpringBootApplication> {
```
donbeave, line 16: I suggest don't delete original file. I know git sometimes do it because you change the content, in this case you should rename file first, commit it, and later change the content, it will save the history of the file.

donbeave (changes requested):

bmuschko (changes requested):
A good start. Can you add the build logic for generating Javadocs, and integrating it into the release process? Having Groovydoc and Javadoc at the same time will be pretty confusing especially to end users so we should complete the migration process asap.

donbeave:
@Sineaggi do you need any help here?

bmuschko:
Could you please also move Java classes under `src/main/java`?

bmuschko:
FYI: I released v9.0.0.

Sineaggi:
It looks like the java source is being properly included in the groovy output, so as long as the main source in considered to be "groovy first" we should be fine until a full switch from `groovydoc` to `javadoc`. See the existing docs for [DockerClientService](https://bmuschko.github.io/gradle-docker-plugin/current/api/com/bmuschko/gradle/docker/internal/services/DockerClientService.html) currently written in java

bmuschko:
@Sineaggi I see you are making great progress. Can you give me a rough idea on where you want to take this PR? We just got a PR that tries to change files you already converted: https://github.com/bmuschko/gradle-docker-plugin/pull/1124. So we'll have to make a call which one goes first. I'd prefer yours but would like to know where and when you are planning to land.

Sineaggi:
@bmuschko I was going to finish out with the root package (`com.bmushko.gradle.docker`) and then migrate the tasks in another mr. Then we could merge the set/convention and I can follow up with another set of conversions.

bmuschko:
@Sineaggi OK, thanks. Can you send me a message when you are done with this PR?

Sineaggi:
@bmuschko Finished the conversion of the basic classes, ready to review and merge now.

bmuschko (approved):
Great, LGTM

PR #1124, opened 2022-11-21
Don't use set to define default property values

### LINKED ISSUE #1119: Tasks should not set values of properties
When tasks are instantiated they should not `set` values of properties because setting values breaks behavior with conventions.

### Expected Behavior

It should be able to configure a conventions when using tasks from this plugin.

### Current Behavior

It's not possible to set a convention e.g. for `DockerCreateContainer.entrypoint` because `empty()` is called in the constructor which sets the value overriding any convention.

### Context

I'm writing a custom plugin on top of this plugin that configures several tasks provided by this plugin. My custom plugin should set some conventions for task values that can be overridden by build script authors. However setting conventions it in some cases not possible due to calls to `empty` and `set` in some of the tasks constructors.

### Reproducer

See https://github.com/britter/gradle-sandbox/tree/main/docker-plugin-conventions-issue

britter:
Changes all occurences where a property default value is defined using
the set method to use convention instead, which is the idiomatic way of
setting a default value.
Furthermore all calls to `empty` on List-, Set-, and MapProperties are
dropped without replacement. These containers default to an empty
collection anyway but calling `empty` will _set_ their value to an empty
collection.

Resolves #1119

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
@@ -61,8 +61,8 @@ class DockerCopyFileFromContainer extends DockerExistingContainer {
     final DirectoryProperty buildDirectory = project.layout.buildDirectory
 
     DockerCopyFileFromContainer() {
-        hostPath.set(project.projectDir.path)
-        compressed.set(false)
+        hostPath.convention(project.projectDir.path)
+        compressed.convention(false)
     }
 
     @Override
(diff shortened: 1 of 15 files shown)


britter:
@bmuschko I have some failures locally that seem to be related to my docker installation. Can you please run this on CI? Thank you!

bmuschko:
Thanks, we are currently in process of turning the production source code from Groovy to Java: https://github.com/bmuschko/gradle-docker-plugin/pull/1117. I want to merge that PR first and then adept your changes to the Java classes if possible. I will let you know when the relevant files have been merged to `master`.

britter:
@bmuschko just let me know. I can make the necessary changes.

bmuschko:
Alright, I merged the other PR. Would you mind resolving the merge conflicts? I will merge your PR next.

britter:
@bmuschko done!

PR #1127, opened 2022-11-29
Convert more image tasks to java

Sineaggi:
Convert a few more tasks to java. More of the easy ones that can be auto-converted then cleaned up.


bmuschko:
Thanks! It has been merged.

PR #1128, opened 2022-11-30
Convert domain objects to java

Sineaggi:
Seems like LivenessProbe is unused, ExecProbe is used as a `@Nested` input, and CopyFileToContainer is used as an input list.


bmuschko:
Thanks, changes have been merged.

PR #1129, opened 2022-12-02
Upgrade to gradle 7.6

Sineaggi:
Also used this time to move `buildSrc` to `build-logic` via an `includeBuild` (as per https://developer.squareup.com/blog/herding-elephants/)

Thoughts?


bmuschko:
I'd rather not upgrade to Gradle 7.6 yet. It will potentially force consumers to use that version to build the project as the plugin was built with it. Is there a specific feature in 7.6 you are planning to use with the plugin that requires the upgrade?

Sineaggi:
> It will potentially force consumers to use that version to build the project as the plugin was built with it.
Is this really the case? Seems insane that a plugin would be unable to upgrade the version of Gradle it's built with.
> Is there a specific feature in 7.6
Nothing in particular, but I try to keep libraries/build tools up-to-date.

bmuschko:
Let's wait with this until everything has been converted to Java and released. We can take additional upgrade steps after.

I am going to close this PR for now and can revive it at a later time.

PR #1133, opened 2022-12-17
Migrate image tasks

Sineaggi:
Migrate the rest of the image tasks to java.

The nasty reflection code is necessary to keep parity with the existing behavior (added in https://github.com/bmuschko/gradle-docker-plugin/issues/1067).


src/main/java/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.java
```diff
@@ -0,0 +1,167 @@
...
+     * The repository and image name e.g. {@code vieux/apache}.
+     *
+     * @since 9.0.0
+     */
+    @Input
+    public final Property<String> getRepository() {
+        return repository;
+    }
+
+    private final Property<String> repository = getProject().getObjects().property(String.class);
```
bmuschko, line 44: For future reference, can you organize all fields together and then the getter methods together?

bmuschko (approved):
LGTM

src/main/java/com/bmuschko/gradle/docker/tasks/image/DockerCommitImage.java
Sineaggi: Yes, will do going forward.

PR #1135, opened 2022-12-18
Migrate networks tasks

Sineaggi:
Migrate the network tasks to java.

The only question I have is about
```java
@Nested
public final ListProperty<Config> getConfigs() {
    return configs;
```
Should a ListProperty be `@Nested`, or `@Input`? I've left it as it was before.


bmuschko:
The `@Nested` is required for a List of POJOs. `@Input` will not provide the correct runtime behavior.

bmuschko (approved):
LGTM

PR #1136, opened 2022-12-18
Replace internal class with public version

Sineaggi:
The main goal of this change was to replace the use of `ProjectInternal#fileOperations` with the public `ArchiveOperations` and `FileSystemOperations` classes which was an oversight of mine while working on the configuration cache support.

This file in particular is probably one of the hairiest in the plugin. It uses io instead of nio where every operation returns a boolean instead of throwing, and the temp file cleanup is difficult to follow.

### DIFF
--- a/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
+++ b/src/main/groovy/com/bmuschko/gradle/docker/tasks/container/DockerCopyFileFromContainer.groovy
@@ -17,17 +17,22 @@ package com.bmuschko.gradle.docker.tasks.container
 
 import com.github.dockerjava.api.command.CopyArchiveFromContainerCmd
 import groovy.io.FileType
+import groovy.transform.CompileStatic
 import org.gradle.api.Action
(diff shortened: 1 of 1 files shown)


bmuschko:
Looks like this has a CI issue.

```
Caused by: java.lang.NullPointerException: Cannot get property 'path' on null object |  
-- | --
  | at com.bmuschko.gradle.docker.tasks.container.DockerCopyFileFromContainer.copyFile(DockerCopyFileFromContainer.groovy:166)
```

Sineaggi:
So I think that's a pre-existing bug, where the logger will attempt to log the path of a file that hasn't been initialized. I added a `@CompileStatic` and that found a few other issues.
```groovy
(fileName.endsWith(".tar") ?: fileName + ".tar") :
```
needs to change to
```groovy
(fileName.endsWith(".tar") ? fileName : fileName + ".tar") :
```
otherwise compressedFileName would be set to true.

Sineaggi:
Closing this in favor of https://github.com/bmuschko/gradle-docker-plugin/pull/1138

PR #1140, opened 2022-12-19
Convert auth locator to java

Sineaggi:
Replaced use of groovy's process api with Gradle's `ExecOperations`. This required injecting the execOperations instance via a constructor, so I added a thin factory type to help.

While testing on windows, the difference between command being a string and a list ended up being difficult, so I rewrote the testing cred scripts to windows batch files.

Finally I added an optional suffix to help the tests differentiate the bash scripts from the batch scripts.


bmuschko:
Great, thanks.

PR #1144, opened 2022-12-19
Fix shading

Sineaggi:
A quick look through the packaged jar shows that `com.github.dockerjava` is still in the root directory. If we relocate this package, other plugins pulling in that library should work.

I've also excluded slf4j from the shaded configuration since that will be pulled in from other plugins, but replaced by gradle at runtime.

The producer https://github.com/bmuschko/gradle-docker-plugin/issues/954#issuecomment-1270100521 in Intellij shows two classes loaded in the same location.
Screen Shot 2022-12-19 at 4 31 20 PM

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,6 +23,10 @@ repositories {
     mavenCentral()
 }
 
+configurations.shaded {
+    exclude("org.slf4j")
+}
+
 dependencies {
     shaded(libs.bundles.docker.java)
     shaded(libs.activation)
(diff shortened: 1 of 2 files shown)


Sineaggi:
It looked like there were still packages from `javax.activation` and `jna` that were missing. I would hope that at least some of these could be considered stable enough to exclude by default.

bmuschko:
I looked at the commit history and it looks like the Docker Java package itself was never relocated. I was 100% convinced it was the case but it doesn't seem to have been the case.

PR #1146, opened 2022-12-27
Create parent directory for container id

napstr:
> Please do not create a Pull Request without first creating an ISSUE.
Any change needs to be discussed before proceeding.
Failure to do so may result in the rejection of the Pull Request.

Discussion can happen here. This wasn't a big up-front investment so if it is the wrong approach please do close =)

> Explain the **details** for making this change: What existing problem does the Pull Request solve? Why is this feature beneficial?

Noticed this problem after updating the docker plugin from 8.1.0 to 9.1.0:

```
> Task :flywayMigrate FAILED
Created container with ID 'codegen-postgres'.

FAILURE: Build failed with an exception.

...

* What went wrong:
Execution failed for task ':flywayMigrate'.
> java.nio.file.NoSuchFileException: /Users/foo/bar/build/.docker/createCodegenDbContainer-containerId.txt
```

This is fixed by executing in the project directory:
```
mkdir build/.docker
```

It seems to be the Groovy migration of `new File(".docker/foo").text="bar"` to `Files.writeString(new File(".docker/foo"), "bar"` might have introduced a behaviour difference where it does not create missing directories. Not easily noticable unless you clean the build environment.

My best guess is that explicitly creating the parent directory should avoid this problem, as suggested in the PR.

### DIFF
--- a/src/main/java/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.java
+++ b/src/main/java/com/bmuschko/gradle/docker/tasks/container/DockerCreateContainer.java
@@ -287,7 +287,9 @@ public class DockerCreateContainer extends DockerExistingImage {
         final String orNull = containerName.getOrNull();
         final String localContainerName = orNull != null ? orNull : container.getId();
         getLogger().quiet("Created container with ID '" + localContainerName + "'.");
-        Files.writeString(containerIdFile.get().getAsFile().toPath(), container.getId());
+        File containerIdFile = this.containerIdFile.get().getAsFile();
+        containerIdFile.getParentFile().mkdirs();
+        Files.writeString(containerIdFile.toPath(), container.getId());
         if (getNextHandler() != null) {
             getNextHandler().execute(container);
         }


Sineaggi:
I was under the impression that by the time a task's `@TaskAction` annotated method is called all `@Output*` annotations will have had their parent directories created.

Sineaggi:
[These docs](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:working_with_files_in_custom_tasks_and_plugins) create the parent dirs themselves.

napstr:
I see the problem on both 7.5.1 and 7.6.

For reproduction, you can use this branch of this project: https://github.com/wolfiabot/Wolfia/commits/docker-plugin-update, simpy run a `./gradlew clean assemble`. For multiple executions you might need to run `docker rm wolfia-codegen-postgres` in between, the script doesn't deal well with a container having been created but not started due to the error.

There is a custom task based on your old [DockerWaitHealthyContainer task](https://github.com/wolfiabot/Wolfia/blob/5dcb0b5c71dcd7bea16d2e78e6c362c475f2e2a4/buildSrc/src/main/groovy/space/npstr/gradle/docker/DockerWaitHealthyContainer.groovy), however the build fails before reaching it during the `DockerCreateContainer`.
Maybe I'm doing something wrong with my the Gradle config? 🤔 I use a weird way to run these tasks: https://github.com/wolfiabot/Wolfia/blob/5dcb0b5c71dcd7bea16d2e78e6c362c475f2e2a4/build.gradle#L271 because I had trouble forcing them to run only when there are relevant changes for flyway.

bmuschko:
@napstr You should never call the task action programmatically in your Gradle script aka the call to the `start()` method. That's will lead to all kinds of unexpected side effects. You should define a `dependsOn` here.

bmuschko:
I am closing this PR as the mechanism for automatically creating the directory via `@OutputDirectory` has been defined properly.

Issue #1123, opened 2022-11-21
DockerRemoteApiPlugin broken in multi project build since 9.0.0


mattlong-finocomp:
I get the following error when attempting to upgrade to version 9.0.0 or 9.0.1 of this plugin.
`A problem occurred evaluating project ':javascript-app'.
> Could not create task ':javascript-app:buildImage'.
   > Cannot set the value of task ':javascript-app:buildImage' property 'dockerClientService' of type com.bmuschko.gradle.docker.internal.services.DockerClientService using a provider of type com.bmuschko.gradle.docker.internal.services.DockerClientService.
`

I suspect this issue was introduced in https://github.com/bmuschko/gradle-docker-plugin/pull/1108

### Expected Behavior
Project with configuration that works in 8.1.0 should continue to work

### Current Behavior
Above error appears when loading the Gradle project in IntelliJ.
See context below for details on project structure, but when upgrading the plugin version  from 8.1.0 in either subproject in isolation the Gradle project cleanly loads. When both are upgraded the above error occurs.

### Context
I have a Gradle multi-project build.
There is a root project which mainly contains no code, but just configures repositories for all projects and configures some unrelated gradle plugins.
There is a java application in a subproject which is using the docker-java-application plugin.
There is a static web application in a subproject which is using the docker-remote-api plugin.
In my real application I use the openapi generator gradle plugin to generate typescript definitions based on an Open API Specification document. I've omitted this from the sample reproduction repo as applying the plugin is enough to trigger the issue.

### Steps to Reproduce (for bugs)
See https://github.com/mattlong-finocomp/gradle-docker-plugin-error
Look at javascript-app/build.gradle and uncomment indicated problematic plugins.
Subsequently downgrading the docker-java-application plugin in the :java-app project will cause the error to not appear even though javascript-app is still using 9.0.1.

### Your Environment
Linux Pop!_OS 20.04 LTS
Java Temurin 17.0.1

Sineaggi:
Does it work if you if you add the plugin with an `apply false` in the root project?

```groovy
plugins {
    id 'base'
    id 'com.bmuschko.docker-remote-api' version '9.0.1' apply false
}
```

bmuschko:
@Sineaggi Thanks for the pointer. Seems like that's the issue. I am going to close the issue.

twarner-sofi:
[...]

PR #1149, opened 2023-01-10
Add checking for classpath issues

Sineaggi:
To fix https://github.com/bmuschko/gradle-docker-plugin/issues/1123

I took a cue from the [kotlin gradle plugin](https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPluginInMultipleProjectsHolder.kt) and added a bit of validation around shared build services.

We can register a new unique build service on failure, and for projects that have the plugin defined at the root (or if gradle fixes the bug) the code will be forwards compatible.

### DIFF
--- a/src/main/java/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.java
+++ b/src/main/java/com/bmuschko/gradle/docker/DockerRemoteApiPlugin.java
@@ -6,8 +6,11 @@ import com.bmuschko.gradle.docker.tasks.RegistryCredentialsAware;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.provider.Property;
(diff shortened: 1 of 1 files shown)


bmuschko:
If I am reading the comment here correctly, the validation will be removed in Gradle core with v8.0: https://github.com/gradle/gradle/issues/23204#issuecomment-1360075575. While your PR contains a "hint" to end users on what to do, I fear that the code will stick around forever and will add to the maintenance cost of the plugin. It's really a Gradle core issue and it feels wrong that plugin developers will have put in custom code because it wasn't handled correctly by the product itself.

Sineaggi:
> it feels wrong

Frankly, I agree. I think since people can work around it for now (placing the dependency in a buildSrc/includeBuild or adding it to their root project) we should be fine skipping this.

Issue #1147, opened 2023-01-04
Shading for Docker Java causes an issue with no being able to resolve JNA classes


m-raab:
I created a plugin to build Docker images based on your plugin. I call DockerBuildImage, but after the update to version 9.1.0 I get an exception:
```
Exception in thread "docker-java-stream--358492054" java.lang.UnsatisfiedLinkError: 'java.lang.String com.bmuschko.gradle.docker.shaded.com.sun.jna.Native.getNativeVersion()'
	at com.bmuschko.gradle.docker.shaded.com.sun.jna.Native.getNativeVersion(Native Method)
	at com.bmuschko.gradle.docker.shaded.com.sun.jna.Native.<clinit>(Native.java:223)
	at com.bmuschko.gradle.docker.shaded.com.github.dockerjava.transport.DomainSocket.<clinit>(DomainSocket.java:54)
	at com.bmuschko.gradle.docker.shaded.com.github.dockerjava.transport.UnixSocket.get(UnixSocket.java:29)
	at com.bmuschko.gradle.docker.shaded.com.github.dockerjava.httpclient5.ApacheDockerHttpClientImpl$2.createSocket(ApacheDockerHttpClientImpl.java:153)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.io.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:125)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.connect(PoolingHttpClientConnectionManager.java:409)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:164)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.classic.InternalExecRuntime.connectEndpoint(InternalExecRuntime.java:174)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.classic.ConnectExec.execute(ConnectExec.java:135)
	at com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.impl.classic.ExecChainElement.execute(ExecChainElement.java:51)
[...]

PR #1154, opened 2023-01-21
Revert shading of Docker Java

bmuschko:
Fix for https://github.com/bmuschko/gradle-docker-plugin/issues/1147.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,6 +25,7 @@ repositories {
 
 configurations.shaded {
     exclude("org.slf4j")
+    exclude("net.java.dev.jna")
 }
 
 dependencies {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -25,7 +25,6 @@ repositories {
 
 configurations.shaded {
     exclude("org.slf4j")
-    exclude("net.java.dev.jna")
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


Sineaggi:
Should the dependency also be excluded from repackaging?

bmuschko:
I tried. It seemed to cause other issues. Looks like shading Docker Java won't work. I reverted those changes here. Can you try with this?

Sineaggi:
I'd simply much prefer not shading at all. Relocation has problems with jna (jni) and not shading has problems with two versions of the same jar being available on the classpath.

bmuschko:
@Sineaggi Yeah, that's unfortunate. Long term, Gradle core should run each plugin in separate classloader to ensure that dependencies don't conflict with each other. At this time, there's nothing else we could have tried from my perspective than shading.

> It’s important to understand that a Gradle plugin does not run in its own, isolated classloader. In turn those dependencies might conflict with other versions of the same library being resolved from other plugins and might lead to unexpected runtime behavior. When writing Gradle plugins consider if you really need a specific library or if you could just implement a simple method yourself. A future version of Gradle will introduce proper classpath isolation for plugins.

https://docs.gradle.org/current/userguide/designing_gradle_plugins.html

big-andy-coates:
I've tested this locally on my M1 Mac and images built perfectly 👍

Sineaggi:
@bmuschko Long term I agree, but I honestly think that for now just having no shading has the least amount of problems.

PR #1162, opened 2023-01-29
Refactor DockerExecContainer for easier extension

### LINKED ISSUE #1159: How to lazily initialize commands for DockerExecContainer?
Hey all,

I'm just looking for a bit of help on how to lazily provide the commands for a `DockerExecContainer` task. 

My project builds a docker image using the gradle-docker-plugin, however I now also want to scan that image for vulnerabilities using [Trivy](https://github.com/aquasecurity/trivy). They don't have a gradle plugin but do have a docker image, so I'm trying to use the `DockerCreateContainer` and `DockerStartContainer` tasks to get a container of it up and running and then run the necessary commands inside of it with a `DockerExecContainer`. The relevant part currently looks like this:
```
        def trivy = project.tasks.register('trivy', com.bmuschko.gradle.docker.tasks.container.DockerExecContainer) {
            group = "Trivy"
            description = "Identifies and reports on known vulnerabilities (CVEs) in the generated docker image for this project"
            dependsOn buildImage, trivyStart
            targetContainerId trivyCreate.get().containerId

            withCommand(['mkdir', '/trivy/'])  // create the directory we'll be saving reports too (trivy can't do it)
            withCommand(['trivy', 'version'])  // print the version for debugging
            withCommand(["trivy", "image", "${-> buildImage.get().imageId.get()}", "-f", "json", "-o", "/trivy/trivy.json", "--security-checks", "vuln", "--timeout", "15m"])
            withCommand(["trivy", "image", "${-> buildImage.get().imageId.get()}", "-f", "table", "-o", "/trivy/trivy.txt", "--security-checks", "vuln", "--timeout", "15m"])

            finalizedBy trivyLog, trivyCopy
        }
```

**The problem I'm having is that the `commands` for the `DockerExecContainer` are eagerly evaluated, but I need them to be lazily evaluated so I can pass in the ID for the image I've just built.** 

Has anybody ever had to do anything similar and found a nice way of doing it? I've tried a couple of different things, most notably trying to add the commands at execution time when the image ID is known, however the `commands` property is already finalized by that point and can't be modified.

The only solution I can see to my problem would be extending the `DockerExecContainer` and overriding the `doRunRemoteCommand` to instead build the commands on the fly using an input containing just the image ID.

ThomGeG:
Explain the **details** for making this change: What existing problem does the Pull Request solve? Why is this feature beneficial?

Just a small refactor to allow for easier extension of the `DockerExecContainer` with more sophisticated behavior to generate commands. 

Having the `doRunRemoteCommand` instead *just* execute single commands while the `runRemoteCommand` calls to it should allow for people to more easily override the task and create their own logic to generate the command list.

Closes #1159

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainerFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainerFunctionalTest.groovy
@@ -40,6 +40,34 @@ class DockerExecContainerFunctionalTest extends AbstractGroovyDslFunctionalTest 
         result.output.contains('Hello World')
     }
 
+    def "Executes overriden commands within running container"() {
+        given:
+        String containerExecutionTask = """
+            import com.github.dockerjava.api.DockerClient
+            abstract class OverridenExecTask extends DockerExecContainer {
+                @Override
+                public void runRemoteCommand() throws InterruptedException {
+                    DockerClient dockerClient = getDockerClient()
+                    doRunRemoteCommand(dockerClient, ['echo', 'Hello World'] as String[])
+                }
+            }
+            task execContainer(type: OverridenExecTask) {
+                dependsOn startContainer
+                targetContainerId startContainer.getContainerId()
+                commands.add(['echo', 'Goodbye World'] as String[])
+            }
+        """
+        buildFile << containerUsage(containerExecutionTask)
+
+        when:
+        BuildResult result = build('logContainer')
+
+        then:
+        // ran the overriden method, not the regular one w/ normal commands
+        result.output.contains('Hello World')
+        !result.output.contains('Goodbye World')
+    }
+
     @Ignore("Flaky test that fails with org.apache.hc.core5.http.StreamClosedException: Stream already closed")
     def "Execute multiple commands within running container"() {
         given:
--- a/src/main/java/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.java
+++ b/src/main/java/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.java
@@ -121,72 +121,72 @@ public class DockerExecContainer extends DockerExistingContainer {
     @Override
     public void runRemoteCommand() throws InterruptedException {
         getLogger().quiet("Executing on container with ID '" + getContainerId().get() + "'.");
-        doRunRemoteCommand(getDockerClient());
+
+        DockerClient dockerClient = getDockerClient();
+        List<String[]> localCommands = commands.get();
+        for (int i = 0; i < localCommands.size(); i++) {
+            doRunRemoteCommand(getDockerClient(), localCommands.get(i));
+        }
     }
 
-    protected void doRunRemoteCommand(DockerClient dockerClient) throws InterruptedException {
+    protected void doRunRemoteCommand(DockerClient dockerClient, final String[] singleCommand) throws InterruptedException {
         ResultCallback.Adapter<Frame> execCallback = createCallback(getNextHandler());
 
-        List<String[]> localCommands = commands.get();
-        for (int i = 0; i < commands.get().size(); i++) {
-
-            final String[] singleCommand = localCommands.get(i);
-            ExecCreateCmd execCmd = dockerClient.execCreateCmd(getContainerId().get());
-            setContainerCommandConfig(execCmd, singleCommand);
-            String localExecId = execCmd.exec().getId();
-            dockerClient.execStartCmd(localExecId).withDetach(false).exec(execCallback).awaitCompletion();
+        ExecCreateCmd execCmd = dockerClient.execCreateCmd(getContainerId().get());
+        setContainerCommandConfig(execCmd, singleCommand);
+        String localExecId = execCmd.exec().getId();
+        dockerClient.execStartCmd(localExecId).withDetach(false).exec(execCallback).awaitCompletion();
 
 
-            // create progressLogger for pretty printing of terminal log progression.
-            final ProgressLogger progressLogger = getProgressLogger(getServices(), DockerExecContainer.class);
-            progressLogger.started();
+        // create progressLogger for pretty printing of terminal log progression.
+        final ProgressLogger progressLogger = getProgressLogger(getServices(), DockerExecContainer.class);
+        progressLogger.started();
 
-            // if no livenessProbe defined then create a default
-            final ExecProbe localProbe = execProbe != null ? execProbe : new ExecProbe(60000, 2000);
+        // if no livenessProbe defined then create a default
+        final ExecProbe localProbe = execProbe != null ? execProbe : new ExecProbe(60000, 2000);
 
-            long localPollTime = localProbe.getPollTime();
-            int pollTimes = 0;
-            boolean isRunning = true;
+        long localPollTime = localProbe.getPollTime();
+        int pollTimes = 0;
+        boolean isRunning = true;
 
-            // 3.) poll for some amount of time until container is in a non-running state.
-            InspectExecResponse lastExecResponse = null;
-            while (isRunning && localPollTime > 0) {
-                pollTimes += 1;
+        // 3.) poll for some amount of time until container is in a non-running state.
+        InspectExecResponse lastExecResponse = null;
+        while (isRunning && localPollTime > 0) {
+            pollTimes += 1;
 
-                lastExecResponse = dockerClient.inspectExecCmd(localExecId).exec();
-                isRunning = lastExecResponse.isRunning();
-                if (isRunning) {
-
-                    long totalMillis = pollTimes * localProbe.getPollInterval();
-                    final long totalMinutes = TimeUnit.MILLISECONDS.toMinutes(totalMillis);
-                    progressLogger.progress("Executing for " + totalMinutes + "m...");
-                    try {
-
-                        localPollTime -= localProbe.getPollInterval();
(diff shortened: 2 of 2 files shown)


bmuschko:
Can you please add a functional test case that demonstrates the use?

ThomGeG:
> Can you please add a functional test case that demonstrates the use?

I've added a simple test case, but haven't been able to actually run it because the integration tests just hang for me. Probably something to do with my docker setup and use of minikube rather than Docker Desktop that I'll have to tinker with.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainerFunctionalTest.groovy
```diff
@@ -40,6 +40,32 @@ class DockerExecContainerFunctionalTest extends AbstractGroovyDslFunctionalTest
...
 
+    def "Execute custom commands within running container"() {
+        given:
+        String containerExecutionTask = """
+            import com.github.dockerjava.api.DockerClient;
+            abstract class CustomExecTask extends DockerExecContainer {
+                @Override
+                public void runRemoteCommand() throws InterruptedException {
+                    DockerClient dockerClient = getDockerClient();
+                    doRunRemoteCommand(getDockerClient(), ['Hello', 'World'] as String[]);
```
bmuschko, line 51: You are calling `getDockerClient()` twice.

bmuschko, line 51: `['Hello', 'World']` is not a command.

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainerFunctionalTest.groovy
```diff
@@ -40,6 +40,32 @@ class DockerExecContainerFunctionalTest extends AbstractGroovyDslFunctionalTest
         result.output.contains('Hello World')
     }
 
+    def "Execute custom commands within running container"() {
```
bmuschko, line 43: How exactly does this fix the issue you reported?

bmuschko (changes requested):

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/container/DockerExecContainerFunctionalTest.groovy
ThomGeG: The `DockerExecContainer.doRunRemoteCommand` method is both iterating over the `commands` _and_ running them, so if you want to extend the task to write your own logic to determine what the commands are (like I want to do so I lazily wrap some around a new input property) you need to duplicate all that logic.

There's no bug or change of behavior, I've just moved the logic that iterates over the commands out so you can override it in isolation and avoid all the duplication. The test doesn't do what I'm interested in doing, its just demonstrating you can override the task to write whatever you want to determine your commands (if I can get it running and fix any more booboos I made).

bmuschko:
Sorry, but I still don't understand what we are trying to achieve with this change. What's the use case for wanting to override an existing command, as demonstrated by your test case?

ThomGeG:
> Sorry, but I still don't understand what we are trying to achieve with this change. What's the use case for wanting to override an existing command, as demonstrated by your test case?

It's not so much about overriding an existing command as it's about refactoring the exec task so it can be more easily extended to write arbitrary logic to determine your commands. The check in the test that the original command didn't run was just to further demonstrate conceptually we're running overridden behaviors, not the normal ones.

My motivation for the change is that the exec task only takes an array of strings as the commands, meaning they're eagerly evaluated and you can't lazily use outputs from other tasks in them. At present the `DockerExecContainer.doRunRemoteCommand` function alone is responsible for _both_ iterating over the commands _and_ executing them. If you wanted to write your own exec task that extended the inbuilt one and adds a new input & logic to lazily evaluate your commands instead, you'd need to rewrite all that execution logic too.

https://github.com/bmuschko/gradle-docker-plugin/blob/c0718d0cfdb7b06c77299ae4e34326b8054d6b2b/src/main/java/com/bmuschko/gradle/docker/tasks/container/DockerExecContainer.java#L127-L137

Instead if they were separate (a function for iterating over the commands and another for actually executing them), you could simply override that first function with your own logic to generate commands that calls into the second, avoiding the duplication.

My specific use case for this would be so I can add a new input property (call it `imageId`), that is itself the output property of another task (`buildImage.imageId`), and then override the function to lazily generate my command at run-time (`doRunRemoteCommand(dockerClient, ['trivy', imageId.get()] as String[])`). I didn't put that into the test I wrote sorry because it felt a bit specific, different people may be interested in doing all kinds of weird things themselves. It was more the principal of it all I was testing/demonstrating, not my specific use-case.

Obviously one could instead just write a entirely new exec task and not extend the existing one, however it felt more prudent to extend so you're capitalizing on all the great work that has previously gone into the existing task and any/all future work.

big-andy-coates:
Hi @ThomGeG,

The `DockerExecContainer` already has a way of updating the commands as an output from another task.  The `getCommands` method returns a [Gradle Property](https://docs.gradle.org/current/userguide/lazy_configuration.html#:~:text=Gradle%20provides%20lazy%20properties%2C%20which,property's%20value%20will%20be%20known.):

```java
@Input
    @Optional
    public final ListProperty<String[]> getCommands() {
        return commands;
    }
```

Gradle properties are _lazy_. You can set the value of a property to the output of another task.  All done in a standard Gradle way. 

The  [`Property#set(Provider)`](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Property.html#set-org.gradle.api.provider.Provider-) method can be used to lazily set a property to the value of another property.

For example, given a `taskA` with an output property called `op` of type `ListProperty<String[]>`, and an instance of `DockerExecContainer` called `decB`, it is possible to do:

```kotlin
decB.commands.set(taskA.op);
```

Seems to me that this is better than the change you're proposing.

ThomGeG:
Hey @big-andy-coates 

The problem here is that it's a `ListProperty` of `String[]`, so correct me if I'm wrong but I'm therefore unable to easily provide individual Strings in a lazy manner. No task generates the entire commands list for me to simply take its output and feed it back in as an input.

I suppose I could write another adhoc task myself responsible for taking the relevant inputs and generating the output `ListProperty` at runtime to be used as the commands. That certainly sounds like a better idea than what I otherwise had planned if this PR couldn't be merged, but seems very much in the same spirit of the PR. Rather than defining a new standalone task, the PR would allow you to extend the existing task to bake the behaviors directly into it instead.

The PR's approach still seems more wholesome and versatile to me (but I'm still fairly new to the Gradle-way of things). There's a lot of great logic in the `DockerExecContainer` task around running arbitrary commands inside the container, it's just that iterating over the commands and running them is done in the same 60 line method. It'd be logical to me to want to split those behaviors up so people can extend the task and write their own logic for determining the commands by doing whatever fancy footwork they need within the same task. Otherwise people would have to define two tasks whenever they need to do a single `DockerExecContainer` with lazy behaviors - one to generate your commands and another to actually run them.

bmuschko:
@ThomGeG Don't spend too much time on the PR. I wanted to have a look at again soon but I don't think we'll want to change the code as proposed.

For two reasons:

* It doesn't solve the original issue you posted in an elegant way or "Gradle-way".
* The previously-existing implementation should be changed in general. Right now you are supposed to pass 1 or more commands but the functional tests for executing more than 1 command fails. This needs to be investigated.
* IMHO the field should be changed to `Property<String>` which would allow you provide a lazily-evaluated command or more than one if you combine with `&&`.

big-andy-coates:
> The problem here is that it's a `ListProperty` of `String[]`, so correct me if I'm wrong but I'm therefore unable to easily provide individual Strings in a lazy manner. No task generates the entire commands list for me to simply take its output and feed it back in as an input.

You can totally build the list of arrays from multiple sources.  The `Provider` interface, which the `Property` interface extends from, is very powerful and flexible.  

In the below example the value of `taskA.op` is _lazily_ evaluated and _mutated_ when `commands` is read:

```kotlin
decB.commands.set(  taskA.op.map { it.subList(0, 1) });
```

The `map` call could also _lazily_ pull in values from other `Property` or `Provider` instances; as many as you like.

> The PR's approach still seems more wholesome and versatile

I'd have to politely disagree. The `Provider` interface is both more flexible and a _standard_ pattern in Gradle, so well known to others.  I'd say it therefore trumps any _custom_ solution IMHO.

That all said, I take your point about splitting the _iteration_ from the _execution_.  My point it more that you can already achieve what you want _without any code change_.

ThomGeG:
Ah, cheers @big-andy-coates 

I've taken what you've mentioned and updated my task to just create a new `ListProperty` to achieve my ends;

```
commands = project.getObjects().listProperty(String[].class).map {[
  ['mkdir', '/trivy/'] as String[],  // create the directory we'll be saving reports too (trivy won't do it)
  ["trivy", "image", buildImage.get().imageId.get(), "-f", "json", "-o", "/trivy/trivy.json", "--security-checks", "vuln", "--timeout", "15m"] as String[],
  ["trivy", "image", buildImage.get().imageId.get(), "-f", "table", "-o", "/trivy/trivy.txt", "--security-checks", "vuln", "--timeout", "15m"] as String[]
]}
```

Definitely agree then, this is a far better way of resolving my problem. Evidently I hadn't dived deep enough into [the Gradle documentation](https://docs.gradle.org/current/userguide/lazy_configuration.html) to see all the references to the factories and the ability to `map`. I'll close this PR off then as its all moot.

Thanks for the help everybody

PR #1164, opened 2023-02-01
Fix DockerLoadImage input

Sineaggi:
Fixes `DockerLoadImage` task to have `@InputFile` and not `@Input`, found while running `check`. It doesn't look like this project calls check during ci, so instead I've added the `validatePlugins` task to our the list of tasks run in github.


bmuschko:
Thanks, this has been merged.

PR #1183, opened 2023-07-11
Fix docker push errors not failing build

### LINKED ISSUE #1178: DockerPushImage does not fail the task / build if the upload fails
### Expected Behavior
If a DockerPushImage task is executed but fails (here it fails because of missing privileges of the user) the build should fail and not be marked as successful.

### Current Behavior
If you run a DockerPushImage task it will finish at the moment with:
```

> Task :pushDockerImage
Pushing image '$registry/image:version' to registry.

BUILD SUCCESSFUL in 6s
```

In the debug logs however you will find this:

`2023-04-24T14:42:02.971+0200 [DEBUG] [com.bmuschko.gradle.docker.shaded.org.apache.hc.client5.http.wire] http-outgoing-2 << "{"errorDetail":{"message":"unauthorized: User is unauthorized to upload to $registry/image/_uploads"},"error":"unauthorized: User is unauthorized to upload to $registry/image/_uploads"}[\r][\n]"`

The image failed to upload because the user does not have the rights to do so.

### Context
Push an image and expect the build to fail if the push fails.

1. Call a DockerPushImage to a registry where you don't have the privilege to upload anything.

### Your Environment
Plugin Version 9.3.1 with Gradle 8.1 and JDK 17.

MarkRx:
Fixes DockerPushImage task not failing when pushing fails

closes #1178

### DIFF
--- a/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImageFunctionalTest.groovy
+++ b/src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImageFunctionalTest.groovy
@@ -1,0 +1,49 @@
+package com.bmuschko.gradle.docker.tasks.image
+
+import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.TaskOutcome
+
+class DockerPushImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
+
+    def "fails on push error"() {
+        def image = createUniqueImageId()
+
+        buildFile << """
+            import com.bmuschko.gradle.docker.tasks.image.Dockerfile
+            import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+            import com.bmuschko.gradle.docker.tasks.image.DockerPushImage
+            import com.bmuschko.gradle.docker.tasks.image.DockerRemoveImage
+
+            task dockerfile(type: Dockerfile) {
+                from '$TEST_IMAGE_WITH_TAG'
+                label(['maintainer': 'jane.doe@example.com'])
+            }
+
+            task buildImage(type: DockerBuildImage) {
+                dependsOn dockerfile
+                inputDir = file("build/docker")
+                images.add("${image}")
+            }
+
+            task removeImage(type: DockerRemoveImage) {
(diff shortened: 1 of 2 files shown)


src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImageFunctionalTest.groovy
```diff
@@ -0,0 +1,40 @@
+package com.bmuschko.gradle.docker.tasks.image
+
+import com.bmuschko.gradle.docker.AbstractGroovyDslFunctionalTest
+import org.gradle.testkit.runner.BuildResult
+
+class DockerPushImageFunctionalTest extends AbstractGroovyDslFunctionalTest {
+
+    def "fails on push error"() {
```
bmuschko, line 8: Can you add a task for add a task for removing the container image after the push. You will also need to add a pull task to ensure that the container image is available before use. It's possible that your test case runs first and the image may not exist yet.

```
task removeImage(type: DockerRemoveImage) {
     dependsOn buildImage
     mustRunAfter pushImage, pullImage
     targetImageId buildImage.getImageId()
     force = true
}
```

src/main/java/com/bmuschko/gradle/docker/tasks/image/DockerPushImage.java
```diff
@@ -80,14 +83,30 @@ public void registryCredentials(Action<? super DockerRegistryCredentials> action
...
                     nextHandler.execute(item);
                 }
             }
 
+            @Override
+            protected void throwFirstError() {
+                super.throwFirstError();
+
+                if (latestItem == null) {
+                    throw new DockerClientException("Could not push image");
```
bmuschko, line 105: Is there a way to get the root cause of this failure e.g. one that indicates that incorrect credentials were used?

bmuschko (changes requested):

src/functTest/groovy/com/bmuschko/gradle/docker/tasks/image/DockerPushImageFunctionalTest.groovy
MarkRx: 👍

bmuschko (approved):
Thanks a lot for changes. They'll go out with the next release of the plugin.

PR #1237, opened 2025-06-23
ISSUE-1236: build project with Java 24

zalio-pl:


bmuschko:
This is not the fix needed to make the plugin compatible with Java 24. It requires an upgrade of the ASM dependency version which I just committed.

